PR #6, opened 2023-02-23
Testing to see whether we can autogenerate code reference docs

skrawcz:
Testing autogenerating python reference docs.


skrawcz:
Proven out and merged in #71

PR #62, opened 2023-02-26
Configurable Validators

elijahbenizzy:
This allows us to configure validation on the per-node level. Done at config-time, this sets us up to determine how to add a validator from passed in configuration.

## Checklist

### DIFF
--- a/data_quality.md
+++ b/data_quality.md
@@ -164,7 +164,7 @@ All configuration keys have two components, joined by a `.` The first component 
 The value will be a dictionary with two possible values:
 
 1. `importance` -- the importance level of the data quality check. Can be either "warn" or "fail"
-2. `enable` -- a boolean indicating whether the data quality check is enabled or not.
+2. `enabled` -- a boolean indicating whether the data quality check is enabled or not.
 
 The specific node name will take precedence, and `global` will apply after that. The information in the code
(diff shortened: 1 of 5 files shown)


elijahbenizzy:
Thoughts:
1. Do we want to enable a per-data-quality check? Currently this is per-node. I like it this way as the API is simpler, but it could get annoying if just one check is breaking.
1. What if there's a node named 'global'? (it shouldn't be a function, and can't be a kwarg...). So I'm ok with there not being one)
1. Need to add more tests

Re: (1) I think we can add an extra field later for this (or add a subfield for check name) but otherwise I'm pretty happy with this. The tricky part is that checks don't have identifiable names...

data_quality.md
```diff
@@ -137,7 +137,7 @@ it executes on every column that's extracted.
 ## Handling the results
```
skrawcz, line 138: this file needs to be merged into the main docs...

skrawcz: (same with basics.md)

elijahbenizzy: Yep was waiting for the docs rewrite -- was planning to chop this commit then copy/paste.

elijahbenizzy:
Ok, I'm going to think through this API this weekend -- feel more confident about the others so i might hold off on this for a bit

PR #83, opened 2023-02-28
Pyspark udf support

skrawcz:
The introduces a new experimental graph adapter - PySparkUDFGraphAdapter.

The point of this adapter is to allow someone to build and manage a DAG of map functions that would result in columns
on a pyspark dataframe.

```python
# specifies a way to do pandas_udfs - for vectorized computation
def spend_per_signup(spend: pd.Series, signups: pd.Series) -> column[pd.Series, float]:
    """The cost per signup in relation to spend."""
    return spend / signups

# plain spark UDF that is row based
def spend_per_signup(spend: float, signups: float) -> float:
    """The cost per signup in relation to spend."""
    return spend / signups

```

# Limitations:
1. Only works for "map" operations.
2. Does not implement the full spectrum of pandas_udfs available. Not all are map based.
3. There potentially are performance impacts to using `with_column` for a large DAG. We should look to use expressions, and then do one final select statement in the `result_builder`.

## Changes
- adds graph adatper to h_spark
- adds example under examples/spark
- adds reference documentation on graph adatpers
- renames type_utils to htypes and adds a "column" annotation

## How I tested this
 - created tests
 - ran the example locally

## Notes
 - this is our first attempt

## Checklist


examples/spark/pyspark_udfs/pandas_udfs.py
```diff
@@ -0,0 +1,32 @@
+"""
+Pandas UDFs.
+Has to only contain map operations!
+"""
+import pandas as pd
+
+from hamilton.function_modifiers import tag
+
+
+@tag(return_type="float")
```
elijahbenizzy, line 10: let's have a special tag (namespaced0 for this. That way, if we want to make a decorator we can do so.

Actually, a simple decorator wrapping `tag` or `NodeDecorator` might be nice here... We have the ability to have "internal" tags that no one else can use...

elijahbenizzy (commented):
More comments/general structure thoughts but this is in draft mode so it works.

Some broader thoughts:
1. What happens when you have missing data/missing indices? Things could get difficult really quick...
2. How does one use this if there's a join at some point? Might be nice to have a recipe/example...
3. Could we use subdag or something similar to represent a series of map operations? Then do a join outside of that?

hamilton/experimental/h_spark.py
```diff
@@ -310,29 +301,95 @@ def _lambda_udf(
...
+        pandas_annotation = {
+            name: param.annotation == pd.Series
+            for name, param in new_sig_parameters.items()
+            if param.default
+            == inspect.Parameter.empty  # bound parameters will have a default value.
+        }
+        return pandas_annotation
+
+    @staticmethod
+    def bind_parameters_to_callable(
```
elijahbenizzy, line 322: Wonder if we can employ [signature.bind](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind) here...

hamilton/experimental/h_spark.py
```diff
@@ -149,3 +149,283 @@ def build_result(
...
+        for output_name, output_value in outputs.items():
+            if output_name not in output_schema:
+                output_schema.append(output_name)
+            if output_name in df.columns:
+                continue
+            else:
+                df = df.withColumn(output_name, lit(output_value))
+        # original schema + new columns should be the order.
+        # if someone requests a column that is in the original schema we won't duplicate it.
+        result = df.select(*[column(col_name) for col_name in output_schema])
```
ramannanda9, line 427: nit here, but you can build select expr here rather than doing df.withColumn. This will affect performance in wide dataframes.

skrawcz: code snippet please? :)

ramannanda9: Sure, 

```python
def some_fun(df):
    outputs={'col1': 23.0, 'col2': 45.0, 'col3': 67.0}
    df_columns=set(df.columns)
    select_exprs=[]
    for output_name, output_value in outputs.items():
        if output_name in df_columns:
            select_exprs.append(column(output_name))
        else:
            select_exprs.append(lit(output_value).alias(output_name))
    return df.select(*select_exprs)
```

elijahbenizzy: Ahh I remember last time you showed us this. IIRC the issue isn't the actual computation, its the fact that the plan gets big and it takes longer to optimize? https://spark.apache.org/docs/latest/api/scala/org/apache/spark/sql/Dataset.html#withColumn(colName:String,col:org.apache.spark.sql.Column):org.apache.spark.sql.DataFrame

elijahbenizzy: Unfortunately python doesn't have `withColumns` :/ Just java/scala.

ramannanda9: IIRC, it leads to a df copy, and that slows stuff down.

ramannanda9: from the docs... 
2.0.0

Note
`this method introduces a projection internally. Therefore, calling it multiple times, for instance, via loops in order to add multiple columns can generate big plans which can cause performance issues and even StackOverflowException. To avoid this, use select with the multiple columns at once.`

elijahbenizzy: Yeah, the problem is the plan right? Not the actual computation that gets run? E.G. executing the plan gets ugly (its kind of an n^2 set of parameters...)

skrawcz: Hold up. The `with_column` here is only called for a minority case. So that's likely never going to be a bottleneck.

What this does bring up though, is whether the internal function to call the UDFs will be. Since it does `with_column` for each UDF... 🤔

hamilton/experimental/h_spark.py
skrawcz: I don't think so -- that's only for the signature object, not the actual function?

hamilton/experimental/h_spark.py
skrawcz: Yeah I'm not going to refactor this now -- but yeah it would be a bit of surgery to rewrite things to return udfs, and then only in build_result do the final select expression. I am currently using the dataframe object to tell me what columns should exist in it -- so would need to replace that with an object, etc...

elijahbenizzy: This will get slow if there's  1000 columns -- but I think its not the end of the world right now, and not worth blocking.

examples/spark/pyspark_udfs/pandas_udfs.py
```diff
@@ -0,0 +1,52 @@
+"""Pandas UDFs.
+
+Has to only contain map operations! Aggregations, filters, etc. are not supported.
```
elijahbenizzy, line 3: What will happen if its not just map operations?

elijahbenizzy (approved):
LGTM -- happy to talk over today but I'm pretty happy with this.

hamilton/experimental/h_spark.py
```diff
@@ -149,3 +149,289 @@ def build_result(
             return df.to_pandas()
         else:
             return df
+
+
+def numpy_to_spark_type(numpy_type: Type) -> types.DataType:
```
elijahbenizzy, line 154: No utility functinos in spark to do this right?

examples/spark/pyspark_udfs/pandas_udfs.py
skrawcz: pandas udfs only have certain shapes -- so it'll break in pyspark.

hamilton/experimental/h_spark.py
skrawcz: I looked -- didn't find anything obvious.

PR #87, opened 2023-03-02
Adds ability to use variables with the driver.execute function #86

### LINKED ISSUE #86: Driver should take in variables to execute as well as functions/node names
We require something like this:

```python
output_columns = [
   item.name for item in dr.list_available_variables()
]
```
But we should be able to do something like this:

```python
output_columns = dr.list_available_variables()
```

skrawcz:
This implements #86  and enables one to pass driver.Variable objects as expected outputs to the driver.execute(). This reduces the boiler plate for someone to massage getting variables for use with .execute().

Adds to unit test.

## Changes
 - adds to driver code for creating the final variables

## How I tested this
 - unit test

## Notes
- Closes #86 

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -226,7 +226,7 @@ class Driver(object):
 
     def execute(
         self,
-        final_vars: List[Union[str, Callable]],
+        final_vars: List[Union[str, Callable, Variable]],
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
🚢 🇮🇹

PR #94, opened 2023-03-04
Feature engineering in multiple contexts examples

### LINKED ISSUE #93: Extract* doesn't work with async function
When using the async driver, you can write async functions. However the decorators don't take async functions into account.

# Current behavior
The following doesn't work:
```python
@extract_columns(*FEATURES_IN_STORE)
async def raw_data(client_id: str, feature_client: FeatureStoreHttpClient) -> pd.DataFrame:
     # pass
```
Instead you need to do:
```python
async def raw_data(client_id: str, feature_client: FeatureStoreHttpClient) -> pd.DataFrame:
    # logic

@extract_columns(*FEATURES_IN_STORE)
def expose_raw_data(raw_data: pd.DataFrame) -> pd.DataFrame:
    return raw_data
```

## Stack Traces
```python
  File "/Users/stefankrawczyk/dagworks/hamilton/hamilton/experimental/h_async.py", line 71, in new_fn
    return fn(**fn_kwargs)
  File "/Users/stefankrawczyk/dagworks/hamilton/hamilton/function_modifiers/expanders.py", line 478, in extractor_fn
    if column_to_extract not in df:
TypeError: argument of type 'coroutine' is not iterable
```

Use the async driver.

Latest.

# Expected behavior
Decorators should handle async functions.

skrawcz:
Adds feature engineering in multiple examples.

This adds two scenarios:

1. one where there is no feature store, and you want to recompute features in two places.
2. one where this is a feature store, however the feature store only stores "raw" data -- so you want to use that 
as the place to get data from to compute features for input to the model. 

In the process, this fixes #93, as this issue was found while building out this example.

## Changes
1. adds examples
2. makes extract* work with async functions.

## How I tested this
1. Runs locally.

## Notes
This isn't an exhaustive set of examples on this topic and how you could use Hamilton. For example this omits talking about:

1. Streaming settings. Though arguably reusing Hamilton functions would be simpler in that context.
2. How to ask Hamilton what features are needed as input to know what to request from the feature store. With tags, and querying the DAG at the start of the app, you could dynamically ask Hamilton what's required and then only go to the
feature store for that data. But I thought that might be a little too complex, so I leave it on the TODO list.

## Checklist


examples/feature_engineering_multiple_contexts/scenario_1/README.md
```diff
@@ -0,0 +1,75 @@
...
+# How it works
+
+1. You run `etl.py`, and in addition to storing the features/training a model, you store the aggregate global values
+for `age_mean` and `age_std_dev`. `etl.py` uses `offline_loader.py`, `features.py`, and `named_model_feature_sets.py`.
+2. You then run `fastapi_server.py`, which is the online webservice with the trained model (not shown here). `fastapi_server.py`
+uses `features.py`, and `named_model_feature_sets.py`.
+Note, in real-life you would need to figure out a process to inject the aggregate global values for `age_mean` and `age_std_dev`
+into the feature computation process. E.g. If you're getting started, these could be hardcoded values, or stored to a file that
+is loaded much like the model, or queried from a database, etc. Though you'll want
+to ensure these values match whatever values the model was trained with. If you need help here, join our slack and we're happy to help you figure this out!
```
elijahbenizzy, line 21: LInk to slack

examples/feature_engineering_multiple_contexts/scenario_1/fastapi_server.py
```diff
@@ -0,0 +1,131 @@
...
+from hamilton.experimental import h_async
+
+app = fastapi.FastAPI()
+
+# know the model schema somehow.
+model_input_features = named_model_feature_sets.model_x_features
+
+
+def load_invariant_feature_values() -> dict:
+    """This function would load the invariant feature values from a database or file.
```
elijahbenizzy, line 28: Make it explicit that you've faked it here for the sake of argument

examples/feature_engineering_multiple_contexts/scenario_1/fastapi_server.py
```diff
@@ -0,0 +1,131 @@
...
+    return {
+        "age_mean": 33.0,
+        "age_std_dev": 12.0,
+    }
+
+
+def fake_model(df: pd.DataFrame) -> pd.Series:
+    """Function to simulate a model"""
+    # do some transformation.
+    return df.sum()  # this is nonsensical but provides a single number.
```
elijahbenizzy, line 40: I think this is distracting cause its so simple here -- a lot of hamilton code is simple and `df.sum()` is going to end up in copy/pastes...

examples/feature_engineering_multiple_contexts/scenario_2/online_loader.py
```diff
@@ -0,0 +1,142 @@
...
+            "day_of_the_week": 3,
+            "seasons": 2,
+            "transportation_expense": 1,
+            "distance_from_residence_to_work": 1,
+            "service_time": 289,
+            "age": 36,
+            "work_load_average_per_day": 13,
+            "hit_target": 33,
+            "disciplinary_failure": 239.554,
+            "education": 97,
```
elijahbenizzy, line 66: What does `education=97` do?

examples/feature_engineering_multiple_contexts/scenario_2/online_loader.py
```diff
@@ -0,0 +1,142 @@
...
+    "body_mass_index",
+]
+
+
+@tag(source="feature_store")
+@extract_columns(*FEATURES_IN_STORE)
+async def raw_data(client_id: str, feature_client: FeatureStoreHttpClient) -> pd.DataFrame:
+    """Extracts the raw data, renames the columns to be valid python variable names, and assigns an index.
+
+    Note: we could get cleverer here and only request the features we need. We can use the Hamilton Driver to ask
```
elijahbenizzy, line 121: Side-note: this feels like a new use-case that's akin to the `select FOO` issue. 
1. Load up data from a db
2. extract columns
3. just use a subset

Should be able to be:
1. Load just the data you need
2. extract all the columns 
3. Use a subset

Is there a recipe here? can't think of an easy implementation to make this natural, but worth thinking through...

elijahbenizzy: Maybe if the function doing the `select` knew the entire path, it could be smart about it... We could automatically pass that in as an input and allow the data loader piece to handle it?

hamilton/function_modifiers/expanders.py
```diff
@@ -454,14 +454,28 @@ def expand_node(
...
-                    if col not in df_generated:
-                        registry.fill_with_scalar(df_generated, col, self.fill_with)
-                        assert col in df_generated
-            return df_generated
+        # if fn is an async function
+        if inspect.iscoroutinefunction(fn):
+
+            async def df_generator(*args, **kwargs) -> Any:
+                df_generated = await fn(*args, **kwargs)
+                if self.fill_with is not None:
```
elijahbenizzy, line 462: This should use the same code as the next one

hamilton/function_modifiers/expanders.py
```diff
@@ -471,16 +485,29 @@ def df_generator(*args, **kwargs) -> Any:
...
-                    raise base.InvalidDecoratorException(
-                        f"No such column: {column_to_extract} produced by {node_.name}. "
-                        f"It only produced {str(df.columns)}"
-                    )
-                return registry.get_column(df, column_to_extract)
+            if inspect.iscoroutinefunction(fn):
+
+                async def extractor_fn(column_to_extract: str = column, **kwargs) -> Any:
+                    df = kwargs[node_.name]
+                    if column_to_extract not in df:
```
elijahbenizzy, line 492: Ditto

hamilton/function_modifiers/expanders.py
```diff
@@ -573,29 +600,56 @@ def expand_node(
...
-                for field in self.fields:
-                    if field not in dict_generated:
-                        dict_generated[field] = self.fill_with
-            return dict_generated
+        # if fn is async
+        if inspect.iscoroutinefunction(fn):
+
+            async def dict_generator(*args, **kwargs):
+                dict_generated = await fn(*args, **kwargs)
+                if self.fill_with is not None:
```
elijahbenizzy, line 608: Ditto

hamilton/function_modifiers/expanders.py
elijahbenizzy: We should have a utility function that it calls to

hamilton/function_modifiers/expanders.py
```diff
@@ -573,29 +600,56 @@ def expand_node(
...
-                if field_to_extract not in dt:
-                    raise base.InvalidDecoratorException(
-                        f"No such field: {field_to_extract} produced by {node_.name}. "
-                        f"It only produced {list(dt.keys())}"
-                    )
-                return kwargs[node_.name][field_to_extract]
+            # if fn is async
+            if inspect.iscoroutinefunction(fn):
+
+                async def extractor_fn(field_to_extract: str = field, **kwargs) -> field_type:
```
elijahbenizzy, line 632: Why does this need to be `async`? IMO the extractor function *shouldn't* be `async`, and I think we can mix/match. There's no await, so its misleading to make it async...

elijahbenizzy (commented):
The first question everyone is going to be asking is "is it slow to be dataframes when its a single item", and the answer is "Not even close to as slow as that API call we're making". 

So, let's add:
1. A section on why we're using pd.Series/pd.DataFrame and how that helps us
2. Alternatives (see idea at #95)

examples/feature_engineering_multiple_contexts/scenario_2/online_loader.py
skrawcz: categorical value.

examples/feature_engineering_multiple_contexts/scenario_2/online_loader.py
skrawcz: yep, it's the same deal. Not in scope for this scenario.

hamilton/function_modifiers/expanders.py
skrawcz: see commit message.

Issue #97, opened 2023-03-04
tags for externally provided nodes


gravesee:
**Is your feature request related to a problem? Please describe.**
In many applications we expect the user to provide inputs to change the behavior of the DAG, e.g. `config.when...`. The Hamilton driver will display errors when these external dependencies aren't provided, but, AFIAK, there is no way to document these for the user or query them eventually for their metadata to, for example, build a config file skeleton.

**Describe the solution you'd like**
The ability to add tags to externally provided inputs. That is, dependencies of node functions that aren't defined anywhere but in the function signatures. Something that let's me signify in code that these are expected inputs of node functions and here is some information about them.

```python
@tag({...})
@external
def input_data_path():
   pass
```

**Describe alternatives you've considered**

I thought about something like this, but there is still a disconnect with the tagged node and the input the user needs to provide.

```python
@tag({...})
def input(raw_input: str) -> str:
   return raw_input
```

**Additional context**
My team wants to expose Hamilton by pre-creating a variety of configurable DAGs. The DAGs would depend on some inputs from the user. We would like to query the nodes-without-module-tags (which all seem to be external) to generate helpful documentation and context about what is expected from the user.

elijahbenizzy:
OK, this is interesting. Let me make sure I get your use-case:

- you want to be able to know, programmatically, which nodes are already *in* the DAG and which nodes should be supplied externally
- currently you're using the `module` tag (which is supobtimal)
- you might *also* want to do this for `@config.when`/configuration-driven DAG shaping
- And you want to be able to process this and give to your users

Right?

My thinking is that there shouldn't be a need to do this in code, unless you want to say "this input should *always* come from the outside world". Instead, you should be able to query the driver for nodes/configuration items that are needed. This would basically be adopting this [list_available_variables](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/driver.py#L378) functionality and adding an additional item to say the "source" of this (which is information we already track, and would just need to wire through). 
E.G.

```python
dr = driver.Driver(config={...}, modules=...)
external_inputs = [var for var in dr.list_available_variables() if var.is_external_node]
```

So, thoughts on that type of API?

_the following is less likely to be relevant but I'm just thinking through some edge-cases_

Note that there's a caveat -- the external inputs that `config` needs are of a different class, as they're used to shape the DAG. E.G. you can choose between two different functions, which could have two different sets of inputs:

```python
def load_file_from_s3(s3_url: str) -> ...
[...]

PR #99, opened 2023-03-05
Gives user visibility into whether nodes are external inputs or not

elijahbenizzy:
This allows you to programmatically determine the inputs to your DAG. One component of #97.

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -69,7 +69,8 @@ class Variable:
 
     name: str
     type: typing.Type
-    tags: Dict[str, str] = field(default_factory=frozenset)
+    tags: Dict[str, str] = field(default_factory=dict)
+    is_external_input: bool = field(default=False)
 
 
 class Driver(object):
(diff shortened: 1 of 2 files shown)


hamilton/driver.py
```diff
@@ -69,7 +69,8 @@ class Variable:
 
     name: str
     type: typing.Type
-    tags: Dict[str, str] = field(default_factory=frozenset)
+    tags: Dict[str, str] = field(default_factory=dict)
+    is_external_input: bool = field(default=False)
```
skrawcz, line 73: we use `UD` -- so should it be `is_user_defined` instead?

elijahbenizzy: If anything we should change that -- `user_defined` to me makes less sense than `external`. Also `UD` is confusing without a legend...

skrawcz: but maybe `is_external_input` is fine 🤔

skrawcz: Yep sure. I'm just thinking about language. @gravesee do you have any preference here?

skrawcz (approved):

PR #107, opened 2023-03-09
Adds group feature for injecting kwargs

elijahbenizzy:
We were doing this in the replacement function. Looks like everythin ghas to go in the replacement function, which makes things slightly trickier, but at least we can centralize the logic in that function.

## Checklist

### DIFF
--- a/hamilton/function_modifiers/expanders.py
+++ b/hamilton/function_modifiers/expanders.py
@@ -75,6 +75,7 @@ class parameterize(base.NodeExpander):
     You also have the capability to "group" parameters, which will combine them into a list.
 
     .. code-block:: python
+
         @parameterize(
             a_plus_b_plus_c={
                 'to_concat' : group(source('a'), value('b'), source('c'))
@@ -86,6 +87,13 @@ class parameterize(base.NodeExpander):
     """
 
     RESERVED_KWARG = "output_name"
+    # This is a kwarg that replaces it with the name of the function
+    # Double underscore means it will not be provided as user-base kwargs
+    # as hamilton is not OK with these output names
+    # We need this as we need to know the name of the function
+    # for the `@inject` usage but its not provided at
+    # construction time, so we provide a placeholder
+    PLACEHOLDER_PARAM_NAME = "__<function_name>"
 
     def __init__(
         self,
@@ -100,12 +108,12 @@ class parameterize(base.NodeExpander):
             - a tuple of assignments (consisting of literals/upstream specifications), and docstring.
             - just assignments, in which case it parametrizes the existing docstring.
         """
-        self.parametrization = {
+        self.parameterization = {
             key: (value[0] if isinstance(value, tuple) else value)
             for key, value in parametrization.items()
         }
         bad_values = []
-        for assigned_output, mapping in self.parametrization.items():
+        for assigned_output, mapping in self.parameterization.items():
             for parameter, val in mapping.items():
                 if not isinstance(val, ParametrizedDependency):
                     bad_values.append(val)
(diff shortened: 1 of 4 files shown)


hamilton/function_modifiers/expanders.py
```diff
@@ -298,9 +308,16 @@ def __missing__(self, key):
...
             return self.specified_docstrings[output_name]
+        doc = fn.__doc__
         if doc is None:
             return None
-        parametrization = self.parametrization[output_name]
+        parameterizations = self.parameterization.copy()
+        if self.PLACEHOLDER_PARAM_NAME in parameterizations:
+            import pdb
+
+            pdb.set_trace()
```
skrawcz, line 318: classic

skrawcz:
Part of me wants the grouped dependency to be dictionary (or list of tuples) and not a straight list. Why? So that someone could do this and get the correct behavior:

```python
@inject(group(source("foo"), source("bar"), source("baz"))
def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
      """One would likely expect the series to be named and thus the output to be "foo", "bar", "baz"... but it's not guaranteed.""""
     return pd.concat(series_list)

@inject(group(source("foo"), source("bar"), source("baz"))
def my_df(series_list: List[Tuple[str, pd.Series]]) -> pd.DataFrame:
      """This would be guaranteed to work""""
     return pd.DataFrame(dict(series_list))
```

skrawcz:
Otherwise @elijahbenizzy should someone be able to do this:

```python
@inject(group(source("foo"), source("foo"), source("foo")))
def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
    """Should we have three? or just one?"""
    return sum(series_list)

constant = pd.Series([1,2,3,4])
@inject(group(source("foo"), source("foo"), source("foo"), value(constant)))
def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
    """Are values allowed?"""
    return sum(series_list)
```

elijahbenizzy:
> Otherwise @elijahbenizzy should someone be able to do this:
> 
> ```python
> @inject(group(source("foo"), source("foo"), source("foo")))
> def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
>     """Should we have three? or just one?"""
>     return sum(series_list)

Ughh, yeah, currently this is broken but not for great reasons. Need to fix + add tests. Was thinking of releasing then fixing -- as this is a weird case. IMO the strangest case that will currently break is:

> @inject(group(source("foo"), source("boo"), source("baz")))
> def my_df(series_list: List[pd.Series], foo: pd.Series) -> pd.DataFrame:
>     """Are values allowed?"""
>     return sum(series_list)

This is busted due to the popping of kwargs when we parse

> constant = pd.Series([1,2,3,4])
> @inject(group(source("foo"), source("boo"), source("baz"), value(constant)))
> def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
>     """Are values allowed?"""
>     return sum(series_list)
> ```

This is fine by me -- it'll just work I think?

skrawcz:
> > Otherwise @elijahbenizzy should someone be able to do this:
> > ```python
> > @inject(group(source("foo"), source("foo"), source("foo")))
> > def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
> >     """Should we have three? or just one?"""
> >     return sum(series_list)
> > ```
> 
> Ughh, yeah, currently this is broken but not for great reasons. Need to fix + add tests. Was thinking of releasing then fixing -- as this is a weird case. 

I would argue against allowing duplicates, since you can achieve that another way more clearly.

> IMO the strangest case that will currently break is:
> 
> > @Inject(group(source("foo"), source("boo"), source("baz")))
> > def my_df(series_list: List[pd.Series], foo: pd.Series) -> pd.DataFrame:
> > """Are values allowed?"""
> > return sum(series_list)
> 
> This is busted due to the popping of kwargs when we parse
> 
> > constant = pd.Series([1,2,3,4])
> > @Inject(group(source("foo"), source("boo"), source("baz"), value(constant)))
> > def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
> > """Are values allowed?"""
> > return sum(series_list)
> 
> This is fine by me -- it'll just work I think?

I don't think we want that to work TBH. What's the "name" of the constant?

elijahbenizzy:
> Part of me wants the grouped dependency to be a dictionary (or list of tuples) and not a straight list. Why? So that someone could do this and get the correct behavior:
> 
> ```python
> @inject(group(source("foo"), source("bar"), source("baz")))
> def my_df(series_list: List[pd.Series]) -> pd.DataFrame:
>     """One would likely expect the series to be named and thus the output to be "foo", "bar", "baz"... but it's not guaranteed."""
>     return pd.concat(series_list)
> 
> @inject(group(source("foo"), source("bar"), source("baz")))
> def my_df(series_list: List[Tuple[str, pd.Series]]) -> pd.DataFrame:
>     """This would be guaranteed to work"""
>     return pd.DataFrame(dict(series_list))
> 
> @inject(group(source("foo"), source("bar"), source("baz")))
> def my_df(series_dict: Dict[str, pd.Series]) -> pd.DataFrame:
>     """This would be guaranteed to work"""
>     return pd.DataFrame(series_dict)
> ```
> 
> Or maybe we can do all the above and know what to do based on the parameter type annotation?

Yeah, so I think we should allow it to be a dict depending on the annotation -- it should be pretty easy. That said, we could just as well do a `grouped_dict` as well as `grouped_list`, so you could have to annotate it in the right way...

elijahbenizzy:
Merging this to consolidate PRs

PR #108, opened 2023-03-09
Configuration-driven DAG shaping + grouping

elijahbenizzy:

## Checklist

### DIFF
--- a/hamilton/function_modifiers/delayed.py
+++ b/hamilton/function_modifiers/delayed.py
@@ -1,0 +1,150 @@
+import enum
+import inspect
+from typing import Any, Callable, Dict, List, Optional, Tuple
+
+from hamilton import settings
+from hamilton.function_modifiers.base import (
+    DynamicResolver,
+    InvalidDecoratorException,
+    NodeTransformLifecycle,
+)
+
+
+class ResolveAt(enum.Enum):
+    CONFIG_AVAILABLE = "config_available"
+
+
+VALID_PARAM_KINDS = [inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY]
+
+
+def extract_and_validate_params(fn: Callable) -> Tuple[List[str], List[str]]:
+    """Gets the parameters from a function, while validating that
+    the function has *only* named arguments.
+
+    :param fn: Function to extract parameters from
+    :return: List of parameter names
+    :raises InvalidDecoratorException: If the function has any non kwargs-friendly arguments
+    """
+    invalid_params = []
+    required_params = []
+    optional_params = []
+    sig = inspect.signature(fn)
+    for key, value in inspect.signature(fn).parameters.items():
+        if value.kind not in VALID_PARAM_KINDS:
+            invalid_params.append(key)
+        else:
+            if value.default is not value.empty:
+                optional_params.append(key)
+            else:
+                required_params.append(key)
+    if invalid_params:
+        raise InvalidDecoratorException(
+            f"Configuration-parsing functions can only except keyword-friendly arguments. "
+            f"Instead got signature: {sig}"
+        )
+    return required_params, optional_params
+
+
+class resolve(DynamicResolver):
+    """Decorator class to delay evaluation of decorators until after the configuration is available.
+    Note: this is a power-user feature, and you have to enable power-user mode! To do so, you have
+    to add the configuration hamilton.enable_power_user_mode=True to the config you pass into the
+    driver.
+
+    If not, this will break when it tries to instantiate a DAG.
+
+    This is particularly useful when you don't know how you want your functions to resolve until
+    configuration time. Say, for example, we want to add two series, and we need to pass the set of
+    series to add as a configuration parameter, as we'll be changing it regularly. Without this,
+    you would have to have them as part of the the same dataframe. E.G.
+
+    .. code-block:: python
+
+        @parameterize(
+            series_sum_1={"s1": "series_1", "s2": "series_2"},
+            series_sum_2={"s1": "series_3", "s2": "series_4"},
+        )
+        def summation(df: pd.DataFrame, s1: str, s2: str) -> pd.Series:
+            return sum(df[item] for item in series_to_sum)
+
+    Note that there are a lot of benefits to this code, but it is a workaround for the fact that
+    we cannot configure the dependencies. With the `@resolve` decorator, we can actually dynamically
+    set the shape of the DAG based on config:
+
+    .. code-block:: python
+
+        from hamilton.function_modifiers import resolve, ResolveAt
+
+        @resolve(
+            when=ResolveAt.CONFIG_AVAILABLE
+            decorate_with=lambda first_series_sum, second_series_sum: parameterize(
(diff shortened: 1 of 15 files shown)


tests/test_end_to_end.py
```diff
@@ -87,3 +94,27 @@ def test_smoke_screen_module(future_import_annotations, monkeypatch):
...
+
+
+def test_end_to_end_with_dynamic_config():
+    config = {
+        "columns_to_sum_map": {
+            "ab": ["a", "b"],
+            "cd": ["c", "d"],
+            "ed": ["e", "d"],
+            "ae": ["a", "e"],
+            # You can create some insane self-referential bullshit
```
skrawcz, line 106: language

tests/test_end_to_end.py
```diff
@@ -1,15 +1,22 @@
...
-
-# import tests.resources.smoke_screen_module
+import tests.resources.dynamic_config
 from hamilton.data_quality.base import DataValidationError, ValidationResult
 
 
+@pytest.fixture(autouse=True)
+def power_user_mode():
+    with mock.patch.dict(os.environ, {"HAMILTON_POWER_USER_MODE": "true"}):
+        yield
```
skrawcz, line 17: should we make it a constructor argument to the driver?

That way when someone is reading the code, they see that driver is taking in the power mode variable? 

My concern here is that someone could turn this on or off and a user wouldn't necessarily know what's changed.

tests/resources/dynamic_config.py
```diff
@@ -0,0 +1,33 @@
...
+
+@dynamic(
+    lambda columns_to_sum_map: parameterize(
+        **{
+            key: {"col_1": source(value[0]), "col_2": source(value[1])}
+            for key, value in columns_to_sum_map.items()
+        }
+    ),
+    at=ResolveAt.COMPILE,
+)
```
skrawcz, line 28: if someone has the flag off -- will Hamilton parse this without breaking?

My assumption is that, with the flag off -- Hamilton will just make this a no-op and drop it from the DAG? or?

elijahbenizzy: Currently it'll break when the driver is instantiated, but it works at import time

hamilton/function_modifiers/delayed.py
```diff
@@ -0,0 +1,150 @@
...
+                required_params.append(key)
+    if invalid_params:
+        raise InvalidDecoratorException(
+            f"Configuration-parsing functions can only except keyword-friendly arguments. "
+            f"Instead got signature: {sig}"
+        )
+    return required_params, optional_params
+
+
+class dynamic(DynamicResolver):
```
skrawcz, line 51: I don't think dynamic is the write word here.

Perhaps having it under `@config.inject` would make the most sense?

tests/resources/dynamic_config.py
skrawcz: also, what happens when there is no `columns_to_sum_map` in the config passed in? Should it fail? or should the node not resolve? My thinking is that it should not resolve to anything.

hamilton/function_modifiers/delayed.py
skrawcz: `@config.resolve` ?

tests/resources/dynamic_config.py
```diff
@@ -0,0 +1,33 @@
...
+def generic_summation(col_1: pd.Series, col_2: pd.Series) -> pd.Series:
+    return col_1 + col_2
```
skrawcz, line 30:
```python
@config.resolve(
    lambda columns_to_sum_map: parameterize(
        **{
            key: {"col_1": source(value[0]), "col_2": source(value[1])}
            for key, value in columns_to_sum_map.items()
        }
    ),
    at=ResolveAt.COMPILE,
)
def generic_summation(col_1: pd.Series, col_2: pd.Series) -> pd.Series:
    return col_1 + col_2
```
I think that reads better...

tests/resources/dynamic_config.py
skrawcz: I think it should be fine to have it, with it off if the flag is off. If anyone depends on it -- they'll get an error saying that it's not satisfied. If they don't need it to run, then it wont error.

hamilton/function_modifiers/delayed.py
elijahbenizzy: Hmm... Some ideas:
```python
config.inject()
config.resolve()
config.when_available()
resolve_at(time=COMPILE)
resolve_at(time=CONFIG_AVAILABLE)
resolve_when(CONFIG_AVAILABLE)
delay_resolution(until=CONFIG_AVAILABLE)
_(when=CONFIG_AVAILABLE)
```

So far I like putting it under `config` -- maybe `resolve` as you said? Not a fan of `inject`, especially since we're already using it for something else.

tests/resources/dynamic_config.py
elijahbenizzy: Yeah that's not bad

hamilton/function_modifiers/delayed.py
skrawcz: In my head the behavior would be something like:
```python
# signature
@config.resolve(func)
def my_func(...)
```
If flag is off:
 - resolve() -> {}

if flag is on and parameters not in config:
 - resolve() -> {}
 
if flag is on and parameters are in config:
 - resolve() -> {my_func}

elijahbenizzy: > *

I think it should hard break if it's not going to work. If you need specifics in the settings to get it to work, then I'd rather fail fast...

tests/test_end_to_end.py
elijahbenizzy: On second thought, the error message is pretty clear, and doing it on the driver-level is weird --  it is an inherent property of the module, rather than the driver. We can always add it to the driver to turn it on, but I think it's pretty easy to see what's happening, and will end up in the driver script anyway.

tests/resources/dynamic_config.py
elijahbenizzy: I think that it should fail -- otherwise its strange DAG behavior. And these are likely to be central to it -- if one inherits it, it should work, otherwise it'll be confusing.

tests/resources/dynamic_config.py
elijahbenizzy: OK, so I don't actually like putting this under `config` now that I think about it -- its pretty confusing. Specifically, we're not actually `injecting` anything, we're using it to delay evaluation. Furthermore, it doesn't actually have anything to do with the other `config` decorators, which people already have a tough time understanding. Its also pretty meta, so I want it kept on its own.

I'm thinking `@delay_resolution(until=CONFIG_AVAILABLE, lambda ...)`. That's very accurate what it does, lives separately, and it nice and readable.

elijahbenizzy:
```python
@resolve(
    when=CONFIG_AVAILABLE,
    decorate_with=lambda: conf_var_1 -> ...):
```

hamilton/function_modifiers/delayed.py
skrawcz: yes let's keep it pretty constrained now -- and then relax if needed.

skrawcz (approved):

PR #116, opened 2023-03-19
Fixes to allow nesting of subdags

elijahbenizzy:
decorators

Every decorator should have access to this, just in case, so we don't have to worry about passing it through. This changes the resolve_config function in function_modifiers.base to pass it through globally. Note that this also *removes* the default setting of this in the @resolve(...) decorator as it is no longer necessary.

Furthermore, this cleans up some issues in which the optional configs provided by a decorator was a list instead of a dictionary.

## Checklist

### DIFF
--- a/hamilton/function_modifiers/recursive.py
+++ b/hamilton/function_modifiers/recursive.py
@@ -256,16 +256,13 @@ class subdag(base.NodeCreator):
         :param nodes:
         :return:
         """
-        already_namespaced_nodes = []
+        # already_namespaced_nodes = []
         new_nodes = []
         new_name_map = {}
         # First pass we validate + collect names so we can alter dependencies
         for node_ in nodes:
             new_name = assign_namespace(node_.name, namespace)
             new_name_map[node_.name] = new_name
-            current_node_namespaces = node_.namespace
-            if current_node_namespaces:
-                already_namespaced_nodes.append(node_)
         for dep, value in self.inputs.items():
             # We create nodes for both namespace assignment and source assignment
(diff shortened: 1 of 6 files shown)


tests/function_modifiers/test_recursive.py
```diff
@@ -269,3 +269,46 @@ def subdag_processor(foo: int, bar: int, baz: int) -> Tuple[int, int, int]:
...
+    def bar(input_1: int) -> int:
+        return input_1 + 1
+
+    # @config.when(plus_one=True)
+    def foo(input_2: int) -> int:
+        return input_2 + 1
+
+    # @config.when(plus_one=None)
+    # def foo__plain(input_2: int) -> int:
+    #     return input_2
```
skrawcz, line 284: why the comments?

hamilton/function_modifiers/recursive.py
```diff
@@ -274,11 +271,11 @@ def _add_namespace(self, nodes: List[node.Node], namespace: str) -> List[node.No
...
-        if already_namespaced_nodes:
-            raise ValueError(
-                f"The following nodes are already namespaced: {already_namespaced_nodes}. "
-                f"We currently do not allow for multiple namespaces (E.G. layered subDAGs)."
-            )
+        # if already_namespaced_nodes:
+        #     raise ValueError(
+        #         f"The following nodes are already namespaced: {already_namespaced_nodes}. "
+        #         f"We currently do not allow for multiple namespaces (E.G. layered subDAGs)."
+        #     )
```
skrawcz, line 278: remove?

skrawcz (approved):

PR #127, opened 2023-04-01
Adds images to examples

skrawcz:
To make it simpler to grok what an example is doing. We should be leveraging this fact to always make it clear that you can provide this documentation if you use Hamilton!

Also fixes some minor bugs I found with one of the examples.

## Changes
 - adds DAG images
 - fixes bug in data loader example
 - fixes typo in one doc string

## How I tested this
 - ran the examples

## Checklist


elijahbenizzy (approved):
Looked over, everything looks good!

PR #128, opened 2023-04-01
Data adapters WIP

elijahbenizzy:

## Checklist


hamilton/function_modifiers/adapters.py
```diff
@@ -0,0 +1,543 @@
...
+        if cls is None:
+            raise InvalidDecoratorException(
+                f"No loader class found for type: {type_} specified by "
+                f"parameter: {inject_parameter} in function: {fn.__qualname__}"
+            )
+        loader_factory = AdapterFactory(cls, **self.kwargs)
+        loader_factory.validate()
+
+
+class load_from__meta__(type):
```
skrawcz, line 258: class doc string to explain to our future selves?

hamilton/function_modifiers/adapters.py
```diff
@@ -0,0 +1,543 @@
...
+
+        def create_decorator(
+            __loaders=tuple(loaders), inject_=None, **kwargs: ParametrizedDependency
+        ):
+            return LoadFromDecorator(__loaders, inject_=inject_, **kwargs)
+
+        return create_decorator
+
+
+class save_to__meta__(type):
```
skrawcz, line 362: ditto

hamilton/function_modifiers/adapters.py
```diff
@@ -0,0 +1,543 @@
...
+    parameter then says how to refer to the output of this node in the DAG.
+
+    If you called this with the driver:
+
+    .. code-block:: python
+
+        dr = driver.Driver(my_module)
+        output = dr.execute(['final_output'], {'raw_data_path': '/path/my_data.json'})
+
+    You would *just* the final result, and nothing would be saved.
```
skrawcz, line 501:
```suggestion
    You would *just* get the final result, and nothing would be saved.
```

hamilton/function_modifiers/adapters.py
```diff
@@ -0,0 +1,543 @@
...
+    As the load_from decorator does, this decorator can be referred to in a
+    dynamic way. For instance, @save_to.json will save the output of the function
+    to a json file. Note that this means that the output of the function must
+    be a dictionary (or subclass thereof), otherwise the decorator will fail.
+
+    Looking at the json example:
+
+    .. code-block:: python
+
+        @save_to.json(path=source("raw_data_path"), artifact_="data_save_output")
```
skrawcz, line 484: yeah I don't like `artifact_` here.

What is bothering me is:

1. it's the name to request it as output from the driver -- and nothing to do with the actual artifact name itself, e.g. the file name. So `artifact_` is a little misleading?
2. `name_` could also be just as ambiguous? 
3. `output_name_` could be better?
4. `hamilton_name_` ?

hamilton/io/utils.py
```diff
@@ -0,0 +1,19 @@
+import os
+from datetime import datetime
+from typing import Any, Dict
+
+
+def get_file_loading_metadata(path: str) -> Dict[str, Any]:
+    """Gives metadata from loading a file.
```
skrawcz, line 7: probably need to rename to `get_file_metadata`, since it's used in both save and load?

hamilton/io/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    - the file size
+    - the file path
+    - the last modified time
+    - the current time
+    """
+    return {
+        "file_size": os.path.getsize(path),
+        "file_path": path,
+        "file_last_modified": os.path.getmtime(path),
+        "file_loaded_at": datetime.now().utcnow().timestamp(),
```
skrawcz, line 18: probably not needed?

skrawcz (commented):
Just some doc stuff I think. Looks good otherwise -- though I don't see where `hamilton/io/__init__.py` is imported?

hamilton/io/utils.py
skrawcz: or at least the field should be changed.

hamilton/io/utils.py
```diff
@@ -0,0 +1,19 @@
...
+    - the file size
+    - the file path
+    - the last modified time
+    - the current time
+    """
+    return {
+        "size": os.path.getsize(path),
+        "path": path,
+        "last_modified": os.path.getmtime(path),
+        "loaded_at": datetime.now().utcnow().timestamp(),
```
skrawcz, line 18: `loaded_at` doesn't make sense? Shouldn't it be `touched_at` or `inspected_at` ?

elijahbenizzy: I spent like 3 years trying to figure out what the `L` in `ETL` stood for and now we've decided it doesn't mean what it always did 🤯 😄  But yeah, agreed.

hamilton/function_modifiers/adapters.py
elijahbenizzy: Agreed, waiting on response in the hamilton community. `access_name_` maybe? `refer_to_`?

elijahbenizzy: Chose `node_name_` for now as its pretty clear what's happening... Can change before release.

skrawcz (approved):

PR #140, opened 2023-04-13
Minor bug/usability fixes for `1.22.1`

elijahbenizzy:

## Checklist


hamilton/function_modifiers/recursive.py
```diff
@@ -173,26 +200,10 @@ def __init__(
         self.inputs = inputs if inputs is not None else {}
         self.config = config if config is not None else {}
         self.external_inputs = external_inputs if external_inputs is not None else []
-        self._validate_config_inputs(self.config, self.inputs)
+        _validate_config_inputs(self.config, self.inputs)
```
skrawcz, line 203: Probably want a try catch here to then contextualize that it's a `subdag` decorator for node `FOO`?

skrawcz (approved):
yeah so looks alright -- we might want to think about how to ensure when an error occurs it provides the right context of the decorator and node name.

hamilton/function_modifiers/recursive.py
elijahbenizzy: Yep good idea

PR #159, opened 2023-05-01
Ensures that we can refer to another node from save_to/load_from

elijahbenizzy:
This fixes a bug in which input types were passed in as "Any", despite the fact that we knew the type from the dataclass. It would result in a mismatch, but now it works (as we use the type we know).

## Checklist


tests/function_modifiers/test_adapters.py
```diff
@@ -367,7 +367,7 @@ def test_loader_fails_for_missing_attribute():
 
 
 def test_pandas_extensions_end_to_end(tmp_path_factory):
-    output_path = tmp_path_factory.mktemp("test_pandas_extensions_end_to_end") / "output.csv"
```
elijahbenizzy, line 370: Posix paths are fine, but I didn't want to change everything to have a union now, so casting is just as easy.

skrawcz (approved):

PR #163, opened 2023-05-04
Adds node name to hamilton dask DAG

elijahbenizzy:

## Checklist

### DIFF
--- a/hamilton/experimental/h_dask.py
+++ b/hamilton/experimental/h_dask.py
@@ -6,6 +6,7 @@ import dask.dataframe
 import numpy as np
 import pandas as pd
 from dask import compute
+from dask.base import tokenize
 from dask.delayed import Delayed, delayed
 from dask.distributed import Client as DaskClient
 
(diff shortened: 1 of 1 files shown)


hamilton/experimental/h_dask.py
```diff
@@ -113,7 +113,7 @@ def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) ->
         :param kwargs: the arguments that should be passed to it.
         :return: returns a dask delayed object.
         """
-        return delayed(node.callable)(**kwargs)
+        return delayed(node.callable, name=node.name)(**kwargs)
```
skrawcz, line 116: I don't think we want this -- it needs to be the fully qualified "name" that uniquely identifies this function.
> Because this name is used as the key in task graphs, you should ensure that it uniquely identifies obj.

From https://docs.dask.org/en/stable/delayed-api.html

elijahbenizzy: Yeah, so I need to think this through. It *does* uniquely identify it though -- there's only one node per name, and a 1:1 to task. That said, `dask_key_name` looks like it might be better. I'm going to try that and update things.

skrawcz: okay if it's unique, then that's what we want I think.

hamilton/experimental/h_dask.py
```diff
@@ -113,7 +113,12 @@ def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) ->
         :param kwargs: the arguments that should be passed to it.
         :return: returns a dask delayed object.
         """
-        return delayed(node.callable)(**kwargs)
+        # No idea why this works, but it also works with `_` at the end
+        import uuid
```
skrawcz, line 117: move this to top level import?

hamilton/experimental/h_dask.py
```diff
@@ -113,7 +113,12 @@ def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) ->
         :param kwargs: the arguments that should be passed to it.
         :return: returns a dask delayed object.
         """
-        return delayed(node.callable, name=node.name)(**kwargs)
+        # No idea why this works, but it also works with `_` at the end
+        import uuid
+
+        return delayed(node.callable, name=node.name)(
+            **kwargs, dask_key_name=str(node.name) + str(uuid.uuid4())[0:6]
```
skrawcz, line 120: need to explain this in the commit as to why UUID?

skrawcz (approved):
Fixed this up a bit.

PR #195, opened 2023-06-20
Caching Graph Adapter

elshize:
Implementation of a caching adapter.

## Changes

Adds an experimental caching adapter that caches nodes tagged with "cache".

## How I tested this

Unit tests.

## Checklist

### DIFF
--- a/examples/caching_nodes/README.md
+++ b/examples/caching_nodes/README.md
@@ -1,3 +1,15 @@
 # Caching Graph Adapter
 
 You can use `CachingGraphAdapter` to cache certain nodes.
+
+This is great for:
+
+1. Iterating during development, where you don't want to recompute certain expensive function calls.
+2. Providing some lightweight means to control recomputation in production, by controlling whether a "cached file" exists or not.
+
+For iterating during development, the general process would be:
+
+1. Write your functions.
+2. Mark them with `tag(cache="SERIALIZATION_FORMAT")`
+3. Use the CachingGraphAdapter and pass that to the Driver to turn on caching for these functions.
+4. When no longer required, you can just skip (3) and any caching behavior will be skipped.
--- a/examples/caching_nodes/README.md
+++ b/examples/caching_nodes/README.md
@@ -12,4 +12,7 @@ For iterating during development, the general process would be:
 1. Write your functions.
 2. Mark them with `tag(cache="SERIALIZATION_FORMAT")`
 3. Use the CachingGraphAdapter and pass that to the Driver to turn on caching for these functions.
+    a. If at any point in your development you need to re-run a cached node, you can pass
+       its name to the adapter in the `force_compute` argument. Then, this node and its downstream
+       nodes will be computed instead of loaded from cache.
 4. When no longer required, you can just skip (3) and any caching behavior will be skipped.
--- a/examples/caching_nodes/README.md
+++ b/examples/caching_nodes/README.md
@@ -1,0 +1,3 @@
+# Caching Graph Adapter
+
+You can use `CachingGraphAdapter` to cache certain nodes.
--- a/examples/caching_nodes/run.py
+++ b/examples/caching_nodes/run.py
@@ -1,17 +1,22 @@
 import logging
+import pathlib
 import sys
 
-from hamilton import driver
-
-import data_loaders
 import business_logic
+import data_loaders
+
+from hamilton import base, driver
+from hamilton.experimental import h_cache
 
 logging.basicConfig(stream=sys.stdout)
 
 # This is empty, we get the data from the data_loaders module
 initial_columns = {}
 
-dr = driver.Driver(initial_columns, business_logic, data_loaders)
+cache_path = "tmp"
+pathlib.Path(cache_path).mkdir(exist_ok=True)
+adapter = h_cache.CachingGraphAdapter(cache_path, base.PandasDataFrameResult())
+dr = driver.Driver(initial_columns, business_logic, data_loaders, adapter=adapter)
 output_columns = [
     data_loaders.spend,
     data_loaders.signups,
--- a/examples/caching_nodes/run.py
+++ b/examples/caching_nodes/run.py
@@ -1,0 +1,24 @@
+import logging
+import sys
+
+from hamilton import driver
+
+import data_loaders
+import business_logic
+
+logging.basicConfig(stream=sys.stdout)
+
+# This is empty, we get the data from the data_loaders module
+initial_columns = {}
+
+dr = driver.Driver(initial_columns, business_logic, data_loaders)
+output_columns = [
+    data_loaders.spend,
+    data_loaders.signups,
+    business_logic.avg_3wk_spend,
+    business_logic.spend_per_signup,
+    business_logic.spend_zero_mean_unit_variance,
+]
+
+df = dr.execute(output_columns)
+print(df.to_string())
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -189,8 +189,8 @@ class CachingAdapter(SimplePythonGraphAdapter):
         cache_path: str,
         *args,
         force_compute: Optional[Set[str]] = None,
-        writers: Optional[Dict[str, Callable[[Any, str], None]]] = None,
-        readers: Optional[Dict[str, Callable[[str], Any]]] = None,
+        writers: Optional[Dict[str, Callable[[Any, str, str], None]]] = None,
+        readers: Optional[Dict[str, Callable[[Any, str], Any]]] = None,
         **kwargs,
     ):
         """Constructs the adapter.
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -318,7 +318,10 @@ class CachingAdapter(SimplePythonGraphAdapter):
                 result = node.callable(**kwargs)
                 logger.debug(
                     "Writing cache for %s to %s with type %s to %s",
-                    node.name, filepath, type(result), cache_format,
+                    node.name,
+                    filepath,
+                    type(result),
+                    cache_format,
                 )
                 self._write_cache(cache_format, result, filepath, node.name)
                 self.computed_nodes.add(node.name)
@@ -326,7 +329,10 @@ class CachingAdapter(SimplePythonGraphAdapter):
             empty_expected_type = self._get_empty_expected_type(node.type)
             logger.debug(
                 "Reading cache for %s from %s with type %s to %s",
-                node.name, filepath, type(empty_expected_type), cache_format,
+                node.name,
+                filepath,
+                type(empty_expected_type),
+                cache_format,
             )
             return self._read_cache(cache_format, empty_expected_type, filepath)
 
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import os
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Dict, Optional
 
 import pandas as pd
 from hamilton.base import SimplePythonGraphAdapter
@@ -118,7 +118,9 @@ class CachingAdapter(SimplePythonGraphAdapter):
 
     def execute_node(self, node: Node, kwargs: dict[str, Any]) -> Any:
         cache_format = node.tags.get("cache")
-        implicitly_forced = any(dep.name in self.computed_nodes for dep in node.dependencies)
+        implicitly_forced = any(
+            dep.name in self.computed_nodes for dep in node.dependencies
+        )
         if cache_format is not None:
             filepath = f"{self.cache_path}/{node.name}.{cache_format}"
             explicitly_forced = node.name in self.force_compute
@@ -134,3 +136,8 @@ class CachingAdapter(SimplePythonGraphAdapter):
             # Otherwise, dependants would always be recomputed if they have a non-cached dependency.
             self.computed_nodes.add(node.name)
         return node.callable(**kwargs)
+
+    def build_result(self, **outputs: Dict[str, Any]) -> Any:
+        """Clears the computed nodes information and delegates to the super class."""
+        self.computed_nodes = set()
+        return super().build_result(**outputs)
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -111,6 +111,14 @@ class CachingAdapter(SimplePythonGraphAdapter):
         return self.readers[fmt](filepath)
 
     def execute_node(self, node: Node, kwargs: dict[str, Any]) -> Any:
+        """Executes nodes conditionally according to caching rules.
+
+        This node is executed if at least one of these is true:
+        * no cache is present,
+        * it is explicitly forced by passing it to the adapter in ``force_compute``,
+        * at least one of its upstream nodes was computed, either due to lack of cache
+          or being explicitly forced.
+        """
         cache_format = node.tags.get("cache")
         implicitly_forced = any(
             dep.name in self.computed_nodes for dep in node.dependencies
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -116,8 +116,8 @@ class CachingAdapter(SimplePythonGraphAdapter):
         This node is executed if at least one of these is true:
         * no cache is present,
         * it is explicitly forced by passing it to the adapter in ``force_compute``,
-        * at least one of its upstream nodes was computed, either due to lack of cache
-          or being explicitly forced.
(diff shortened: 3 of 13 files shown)


elshize:
@skrawcz hey, I didn't really have the time to fill out the PR info above, and the code is essentially copy-pasted from what I implemented in my use case, but I figured you might want to give this a look sooner than later, and maybe provide any comments.

As I said, there might be some hacky stuff there, especially the way I'm testing, but it's a good start I think.

skrawcz:
Thanks @elshize yep this is great. I'll queue it up for review for later this week/early next week.

hamilton/experimental/h_cache.py
```diff
@@ -0,0 +1,136 @@
...
+                self._write_cache(result, cache_format, filepath)
+                self.computed_nodes.add(node.name)
+                return result
+            return self._read_cache(cache_format, filepath)
+
+        if implicitly_forced:
+            # For purposes of caching, we only mark it as computed if any cached input was computed.
+            # Otherwise, dependants would always be recomputed if they have a non-cached dependency.
+            self.computed_nodes.add(node.name)
+        return node.callable(**kwargs)
```
skrawcz, line 136: Just to make sure I'm understanding this, this will:

1. Skip computation of a node, if it's cached, independent of input values.
2. It will compute a node with a cache annotation if:
    - the cache file does not exist
    - it was explicitly told to compute it via explicitly_forced
    - if there was an earlier cached node that was upstream that was recomputed via implicitly_forced

This means that a user operating this would:
 - stick @cache on say data loading steps -- usually the beginning of the DAG
 - stick @cache on anything else
 - then as they iterate, either force re-computation, or blow away the directory to run and cache new values

Right?

hamilton/experimental/h_cache.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _write_cache(self, data: Any, fmt: str, filepath: str) -> None:
+        self._check_format(fmt)
+        self.writers[fmt](data, filepath)
+
+    def _read_cache(self, fmt: str, filepath: str) -> None:
+        self._check_format(fmt)
+        return self.readers[fmt](filepath)
+
+    def execute_node(self, node: Node, kwargs: dict[str, Any]) -> Any:
```
skrawcz, line 119: this warrants a doc string to explain the functionality.

hamilton/experimental/h_cache.py
```diff
@@ -0,0 +1,136 @@
...
+        Parameters
+        ----------
+        cache_path : str
+            Path to the directory where cached files are stored.
+        force_compute : optional, set
+            Set of nodes that should be forced to compute even if cache exists.
+        writers : optional, dict
+            A dictionary of writers for custom formats.
+        readers : optional, dict
+            A dictionary of readers for custom formats.
```
skrawcz, line 78: we follow the restructured style of doc strings.

hamilton/experimental/h_cache.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Any node with tag "cache" will be cached (or loaded from cache) in the format defined by the
+    tag's value. There are a handful of formats supported, and other formats' readers and writers
+    can be provided to the constructor.
+
+    Values are loaded from cache if the node's file exists, unless one of these is true:
+     * node is explicitly forced to be computed with a constructor argument,
+     * any of its (potentially transitive) dependencies that are configured to be cached
+       was nevertheless computed (either forced or missing cached file).
+    """
```
skrawcz, line 56: It'd be great to include some example code here. Since this string will go into being reference documentation in the docs.

tests/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+
+    caplog.clear()
+
+    # need a new adapter for caching to take proper effect
+    adapter = CachingAdapter(
+        cache_path,
+        base.DictResult(),
+        readers={"str": read_str},
+        writers={"str": write_str},
+    )
```
skrawcz, line 128: I was thinking that we could just reset the current object after `.execute()` was run.  Was this an issue for you? or not?

If so, then options are:
 - provide explicit .reset() 
 - override the result builder in the adapter to reset the internal values, and then call the parent result builder?

If not, then we can leave it as is. Not hard to change the current behavior in the future.

tests/test_caching.py
```diff
@@ -0,0 +1,160 @@
...
+        }
+    )
+
+
+def _read_write_df_test(df, filepath, read_fn, write_fn):
+    write_fn(df, filepath)
+    pd.testing.assert_frame_equal(read_fn(filepath), df)
+
+
+def test_feather(df, tmp_path):
```
skrawcz, line 35: is tmp_path a pytest thing? I don't see a fixture.

tests/test_caching.py
```diff
@@ -0,0 +1,160 @@
+import logging
+
+import pandas as pd
+import pytest
+
+from hamilton import base
+from hamilton.driver import Driver
+from hamilton.experimental.h_cache import (
```
skrawcz, line 8: this is a nit pick, but I think module imports are nicer --  then when you read you know where the function definition lives.  But happy to be convinced otherwise.
```suggestion
from hamilton.experimental import h_cache
```

skrawcz (commented):
Yeah this looks good. I like how simple it is. It can always be made fancier, but I think this probably serves as a good 80/20.

Otherwise to get this out, I think we just need some more documentation on usage, e.g. outlining how one might use it for development, as well as production. Listing any caveats.

hamilton/experimental/h_cache.py
elshize: Yeah, this sounds right. One case that is missing (I'll need to double-check, but at the first glance it seems true) is that a node will be computed if an upstream node was computed due to missing file, not only if it was forced.

I'm not entirely sure if this is desirable, though. It's definitely erring on the side of caution (it _is_ computed, so there's no guarantee that the downstream shouldn't change, e.g., because it's reading an updated dataset), but not quite sure if this would be an intuitive behavior for everyone.

Any thoughts on that? In the usecase I was writing it for, the current way is the expected one, but doesn't mean it would be always.

We could always give an option to do one way or another but this also further complicates the interface and this could be difficult to express the intention (it took me a paragraph here to explain).

hamilton/experimental/h_cache.py
elshize: Good point, will do.

tests/test_caching.py
elshize: Yeah, I actually agree, often I do it myself but as you can see, I struggle with consistency :)

tests/test_caching.py
elshize: Yes, it's a built-in pytest fixture: https://docs.pytest.org/en/7.1.x/how-to/tmp_path.html#the-tmp-path-fixture

tests/test_caching.py
elshize: What do you mean by reset the current object?

elshize:
Thanks, I'll try to get the changes ready soon.

tests/test_caching.py
skrawcz: TIL!

hamilton/experimental/h_cache.py
skrawcz: If you wanted to double check how it renders, you can add to this file https://github.com/DAGWorks-Inc/hamilton/blob/main/docs/reference/api-reference/graph-adapters.rst and then build the docs locally following https://github.com/DAGWorks-Inc/hamilton/blob/main/README-DOCS.md.

tests/test_caching.py
skrawcz: If I'm understanding this, correctly. You create a second CachingAdapter object, because it does not clear the internal dictionaries between executions. Right? We could add something to clear the state in the adapter after a run.

hamilton/experimental/h_cache.py
skrawcz: Ah yes. I would just mention that as part of the caveats; I wouldn't build for that case.
To that end, we could add `debug` logs that people can turn on to see what was or was not loaded from cache.

tests/test_caching.py
elshize: Oh, I thought you were suggesting this would work out of the box or something.

Yeah, we can certainly add a `reset` function, but let me look into whether I can do it with result builder first.

elshize:
Made some changes, still some stuff left, including more docs (with examples), and debug logs.

One thing I noticed when I ran tests on my other desktop is that I'm getting errors due to not having either feather or parquet support, despite installing all the deps.

I see that you do use both in other parts of the code, like in `pandas_extensions.py` so I'm assuming this is ok? And it's not in requirements because it's an optional dep?

skrawcz:
> Made some changes, still some stuff left, including more docs (with examples), and debug logs.

Thanks looking good thus far. Made a suggestion for clarity.

> 
> One thing I noticed when I ran tests on my other desktop is that I'm getting errors due to not having either feather or parquet support, despite installing all the deps.
> 
> I see that you do use both in other parts of the code, like in `pandas_extensions.py` so I'm assuming this is ok? And it's not in requirements because it's an optional dep?

Ah good point!  Hmmm. We could do some "try/catch-ing" to determine what exists on import, and only include the things that are? 

But...  we probably want to make this adapter generic/extensible in a way. e.g. make it work for polars/pluggable for polars/other object types. 

So as to support someone doing this:
```python
@tag(cache="json")
def my_df(...) -> pd.DataFrame:
    return df

@tag(cache="json")
def my_dict(...) -> dict:
    return {"a":"b"}

@tag(cache="json")
def my_polars_df(...) -> pl.DataFrame:
    return df
```
So we could build a registry of (format, type) -> serde function, but maybe that's outside the scope of this PR. Right now the design is simple. Format implies datatype and serialization -- did you think about that aspect at all?  

So if anything maybe we just want to add `isinstance` checks to the writer functions to help surface a good error if someone screws this up. Thoughts?

elshize:
> But... we probably want to make this adapter generic/extensible in a way. e.g. make it pluggable for polars/other object types.

Yeah, I see what you mean. I haven't really thought of it because I didn't need it when implementing it, but this is definitely going to come up.

To not make it too complicated at the moment, I would approach it this way: We can default serializers for some number of output formats, and it would check what is passed to it and do the right thing -- for whatever number of types we want to support out of the box. If someone needs to support an additonal type, then they can pass an extended writer:

```python
def extended_writer(data, path):
    if isinstance(data, SomeType):
        write_some_type(data)
    else:
        hamilton_writer(data, path)
```

It's probably not the most elegant way to approach it, but it's very simple, most people would be covered by the default types supported, and those who are not would have a way of extending it.

Then, we could think whether a more complex approach is needed and what it would be, and if there is one, we implement it later. Thoughts?

skrawcz:
okay @elshize I came up with https://github.com/DAGWorks-Inc/hamilton/pull/211 to extend things. Thoughts?

elshize:
I checked it out and commented on the other PR. This is better than my solution I spoke about above :)

So is there anything else missing there other than some examples in the module doc? I should be able to have some time over the weekend to write some down; how would you want me to add it? Should I use the other branch, branch off and create a new PR, or edit here, and you can rebase?

skrawcz:
@elshize sorry, I missed this (the weekend is now over...). Yeah let's just get the module doc in order I think. 

In terms of adding it, do whatever is easiest for you. The other branch is based off of this one. You could pull in all my commits to this branch, and then add the new one(s)? 

We might want to squash some commits to keep history tidy but let's worry about that last.

Thanks!

elshize:
> You could pull in all my commits to this branch, and then add the new one(s)?

Yeah, I could do that.

> We might want to squash some commits to keep history tidy but let's worry about that last.

Do you imagine there would be more than one commit for this? I was thinking of squashing all of it together, seems like a small enough and isolated change.

skrawcz:
> > You could pull in all my commits to this branch, and then add the new one(s)?
> 
> Yeah, I could do that.
👍 

> 
> > We might want to squash some commits to keep history tidy but let's worry about that last.
> 
> Do you imagine there would be more than one commit for this? I was thinking of squashing all of it together, seems like a small enough and isolated change.

I hadn't looked hard to be honest. But a single squashed commit would work too.

elshize:
@skrawcz Btw, there are more modules I had to install in order to fully run tests:
```
pandera pyspark ray dask pyarrow distributed
```

elshize:
Also, I noticed that there's `data` parameter in `read` functions -- this probably shouldn't be there, should it? It's not used, and I'm not sure how it would be.

skrawcz:
> @skrawcz Btw, there are more modules I had to install in order to fully run tests:
> 
> ```
> pandera pyspark ray dask pyarrow distributed
> ```
> 
> And then still some spark and ray tests failed for me locally. Not sure why, could be because versions of the ones above don't match what is required.

Ah - what did you assume you needed, so I can fix up any docs. My assumption is (1) I should probably change https://github.com/DAGWorks-Inc/hamilton/blob/main/.ci/test.sh to instead refer to `requirements-test.txt` from within the graph adapter directories. (2) make a note in the general `requirements-test.txt` that it only applies to unit test, not the graph adapter integration test.

Note: Since we use circleci, you can also download the circleci CLI and run a job locally and it'll replicate the behavior of that test with docker -- but that is a little slow to iterate with.

Mind posting which tests? This is annoying if they're flakey.

elshize:
here are logs: [pytest.log](https://github.com/DAGWorks-Inc/hamilton/files/12101048/pytest.log) notice that besides graph adapter tests, there is also one telemetry test that failed: 
image


I didn't really _need_ it for the Pr, I just wanted to run all tests :) so really not a problem for me. I just didn't know exactly what you expected to be in requirements, so wanted to mention it.

skrawcz:
> here are logs: [pytest.log](https://github.com/DAGWorks-Inc/hamilton/files/12101048/pytest.log) notice that besides graph adapter tests, there is also one telemetry test that failed: image
> 
> I didn't really _need_ it for the Pr, I just wanted to run all tests :) so really not a problem for me. I just didn't know exactly what you expected to be in requirements, so wanted to mention it.

Ah yep. The telemetry test fails locally; something I haven't dug into -- but it's not flakey on circleci proper.

elshize:
@skrawcz ok, I think this is ready for you to have another look at. One more thing that we maybe want to do is to include info about it somewhere in the `docs/`. Let me know if and where you would want that.

I haven't squashed commits yet, but I think we should just squash to one single commit, it's a simple enough change, I don't see a need for any more than that, but let me know if you feel different.

skrawcz:
@elshize for docs we'd add a new file to https://github.com/DAGWorks-Inc/hamilton/tree/main/docs/reference/graph-adapters (and add it to index.rst) and then perhaps a doc in https://github.com/DAGWorks-Inc/hamilton/tree/main/docs/how-tos explaining how you'd make use of it? We can always put the latter in a separate PR, but the first one would then at least add the adapter to the reference section.

examples/caching_nodes/run.py
```diff
@@ -0,0 +1,24 @@
...
+
+import data_loaders
+import business_logic
+
+logging.basicConfig(stream=sys.stdout)
+
+# This is empty, we get the data from the data_loaders module
+initial_columns = {}
+
+dr = driver.Driver(initial_columns, business_logic, data_loaders)
```
skrawcz, line 14: This should have the adapter passed in here right?

examples/caching_nodes/README.md
```diff
@@ -0,0 +1,3 @@
+# Caching Graph Adapter
+
+You can use `CachingGraphAdapter` to cache certain nodes.
```
skrawcz, line 3: Not sure if you finished the README, but something like this would work:

```suggestion
You can use `CachingGraphAdapter` to cache certain nodes.

This is great for:

1. Iterating during development, where you don't want to recompute certain expensive function calls.
2. Providing some lightweight means to control recomputation in production, by controlling whether a "cached file" exists or not.

For iterating during development, the general process would be:

1. Write your functions.
2. Mark them with `tag(cache="SERIALIZATION_FORMAT")`
3. Use the CachingGraphAdapter and pass that to the Driver to turn on caching for these functions.
4. When no longer required, you can just skip (3) and any caching behavior will be skipped.
```

elshize: No, sorry, I was doing it at work and got distracted by other stuff, but wanted to push my changes :)

Thanks for the suggestion. I should be able to check this out and your other comments tomorrow.

examples/caching_nodes/run.py
elshize: Correct, fixed now.

elshize:
@skrawcz ok, this is ready for you to review.

skrawcz (approved):
Looks good. Just needed to adjust the requirements.txt for the example. Thanks for all the work!

elshize:
:fire:

PR #200, opened 2023-06-23
example using Hamilton for NLP / LLM [WIP]

zilto:

## Checklist

### DIFF
--- a/examples/llm/data_collection.py
+++ b/examples/llm/data_collection.py
@@ -1,0 +1,74 @@
+import feedparser
+import unstructured
+import weaviate
+from unstructured.partition.html import partition_html
+from unstructured.staging.weaviate import create_unstructured_weaviate_class, stage_for_weaviate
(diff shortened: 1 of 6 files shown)


examples/llm/data_collection.py
```diff
@@ -0,0 +1,74 @@
...
+    entry_urls = []
+    for url in rss_feed_urls:
+        rss_feed = feedparser.parse(url)
+        urls = [e["link"] for e in rss_feed.entries if e.get("link")]
+        entry_urls.extend(urls)
+
+    return entry_urls
+
+
+def rss_entry_url(rss_entry_urls: list[str]) -> str:
```
zilto, line 27: Here, I cheat by only looking at one entry in the list

skrawcz: because there could be multiple urls? So you just take the first one?

PR #208, opened 2023-07-02
Adds originating_functions to copy_to

elijahbenizzy:

## Checklist


skrawcz (commented):
unit tests?

tests/test_node.py
```diff
@@ -20,3 +20,16 @@ def fn():
...
     with pytest.raises(ValueError):
         Node.from_fn(fn)
+
+
+def test_node_copy_with_retains_originating_functions():
+    def fn() -> int:
+        return 1
+
+    node = Node.from_fn(fn)
+    node_copy = node.copy_with(name="rename_fn", originating_functions=(fn,))
```
skrawcz, line 30: You shouldn't need `originating_functions=(fn,)` here right?

skrawcz (changes requested):
Um, I don't think this fixes it.

You need to actually populate it the first time.  E.g. [here](https://github.com/DAGWorks-Inc/hamilton/pull/208/commits/e726fe778aac16053186255a2cf7b3163d05e3dd#diff-6b4a15b6f5263150ab8674431da707b5ea1c4a855e07a42ffb79ff5aa56cea86R209-R215)

skrawcz (approved):
Ah okay I understand now. This is on purpose.

tests/test_node.py
elijahbenizzy: Nope its currently assigned in `resolve_nodes`

skrawcz (approved):

PR #211, opened 2023-07-03
Example to extend #195

skrawcz:
This generalizes the capability to serialize and deserialize for caching.

It uses singledispatch to do the magic, and a cute trick on reading to also make singledispatch work.

This builds off of #195 @elshize .

## Changes
 - adds to h_cache

## How I tested this
 - via unit tests

## Notes
 - there is some duplication of functionality here with respect to the data savers and loaders. We should figure out how to combine them.

## Checklist

### DIFF
--- a/hamilton/experimental/h_cache.py
+++ b/hamilton/experimental/h_cache.py
@@ -1,7 +1,7 @@
 import json
 import logging
 import os
-from typing import Any, Callable, Optional
+from typing import Any, Callable, Dict, Optional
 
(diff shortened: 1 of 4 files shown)


elshize:
@skrawcz Sorry for the late reply, have been quite busy and only now finding the time to look at this.

I really like this. I didn't know about `singledispatch` annotation, I think it's a great solution.

skrawcz:
This was included into #195.

Issue #19, opened 2023-02-26
Do we want to support iterators for data loading?


HamiltonRepoMigrationBot:
<a href="https://github.com/skrawcz"><img src="https://avatars.githubusercontent.com/u/2328071?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [skrawcz](https://github.com/skrawcz)**
_Monday Feb 07, 2022 at 17:25 GMT_
_Originally opened as https://github.com/stitchfix/hamilton/issues/68_

----

# What?
If we want to chunk over data, a natural way to do that is via an iterator.

## Example: Enable "input" functions to be iterators
```python
def my_loading_funct(...) -> Iterator[pd.Series]:
     ...
     yield some_chunk 
```
This is fraught with some edge cases. But could be a more natural way to chunk over large data sets? This perhaps requires a new driver -- as we'd want some `next()` type semantic logic on the output of execute...

_Originally posted by @skrawcz in https://github.com/stitchfix/hamilton/issues/43#issuecomment-1030961335_

## Things to think through whether this is something useful to have:

1. Where would we allow this? Only on "input" nodes?
2. How would we exercise them in a deterministic fashion? i.e. does execute() care? and we iterate over them until they're exhausted? Or does `execute()` only do one iteration?
3. How do we coordinate multiple inputs that are iterators? What if they're of different lengths?
4. How would we ensure people don't create a mess that's hard to debug?
5. Would this work for the distributed graph adapters?

HamiltonRepoMigrationBot:
<a href="https://github.com/elijahbenizzy"><img src="https://avatars.githubusercontent.com/u/3886307?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [elijahbenizzy](https://github.com/elijahbenizzy)**
_Saturday Oct 29, 2022 at 17:31 GMT_

----

Yeah I'd need to see more of a workflow/motivating example for this TBH. E.G. a graphadapter that chunks up data and iterates in a streaming sense could be high-value as well...

elijahbenizzy:
OK, inspired by @jdonaldson's hamilton-like framework, I'm curious what could happen if we use chunking. Going to walk through a few different UIs I've been mulling over that would fit into the way we currently do things.

# some use-cases:

1. Run the whole DAG on a stream of data -- E.G. run it within a kafka stream
2. Mini-batch training (everything streaming until the end, when we combine it all together to form a model)
[...]

PR #216, opened 2023-07-07
Shows generator proposal for LLMs (#19)

elijahbenizzy:
This PR shows the before and after code of the API we're proposing to support:
* def data_set() -> Parallel[ TYPE]:
* def data_set() -> Sequential[ TYPE]:  (not shown in this PR)
* def results(parsed_data: Collect[TYPE]) -> ...

This effectively removes "for loops" from "batch" functions, in favor of having the framework handle this.

## Problem this addresses
If you have a list of things to parse/hit an API with. You end up needing a for-loop to do it. This means that the internals of the function can become complex and hard to test. 

## Changes
 - changes to LLM KR example

## How I tested this
- N/A

## Notes
-  This PR is for demonstration purposes.

### DIFF
--- a/examples/LLM_Workflows/knowledge_retrieval/arxiv_articles.py
+++ b/examples/LLM_Workflows/knowledge_retrieval/arxiv_articles.py
@@ -1,21 +1,19 @@
-import concurrent
 import os.path
-from typing import List, Tuple
+from typing import Dict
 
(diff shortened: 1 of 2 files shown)


skrawcz:
This has been implemented I believe. So no need for this PR.

Issue #227, opened 2023-07-23
Data quality pandera example broken since pandas v2


lmmx:
I'm presuming that the example is meant to show a successful validation, but instead it fails due to numerous issues in the data.

**Edit** I downgraded pandas to v1.4 on the hunch that it was breaking changes in pandas itself and sure enough it fixed the errors. This is a v2 breaking change (with `pip install "pandas<2"` it still works, `pip install pandas==2.0` gives these errors)

Also, the final part of the example

```py
if __name__ == "__main__":
    # this is here as a quick way to check that things are working.
    d = raw_data__base("Absenteeism_at_work.csv")
    print(d.to_string())
```

fails and is not fixed by downgrading pandas. It looks like it's supposed to load the data first but it's just passing the filename in when the function expects a dataframe.

```
Traceback (most recent call last):
line 60, in raw_data__base
    df.columns = _sanitize_columns(df.columns)
AttributeError: 'str' object has no attribute 'columns'
```

# Current behavior

Running `python run.py` in the pandera data quality example, after installing the requirements:

```py
WARNING:hamilton.telemetry:Note: Hamilton collects completely anonymous data about usage. This will help us improve Hamilton over time. See https://github.com/dagworks-inc/hamilton#usage-analytics--data-privacy for details.
WARNING:hamilton.data_quality.base:[day_of_week_encoded:pandera_schema_validator] validator failed. Message was: Schema None: A total of 5 schema errors were found.

Error Counts
------------
- SchemaErrorReason.SCHEMA_COMPONENT_CHECK: 5

Schema Error Summary
--------------------
                                                failure_cases  n_failure_cases
schema_context column            check                                        
Column         day_of_the_week_2 dtype('uint8')        [bool]                1
               day_of_the_week_3 dtype('uint8')        [bool]                1
               day_of_the_week_4 dtype('uint8')        [bool]                1
               day_of_the_week_5 dtype('uint8')        [bool]                1
               day_of_the_week_6 dtype('uint8')        [bool]                1

Usage Tip
---------

Directly inspect all errors by catching the exception:

'''
try:
    schema.validate(dataframe, lazy=True)
except SchemaErrors as err:
    err.failure_cases  # dataframe of schema errors
    err.data  # invalid dataframe
'''
[...]

PR #228, opened 2023-07-23
Avoid get_dummies dtype change in pandas v2 (keep backwards compatible)

lmmx:
- #227

## Changes

- The `dtype` argument has been added to all uses of `get_dummies`, as the default value changed in pandas v2

## How I tested this

- `python run.py` no longer produces errors

## Checklist

### DIFF
--- a/examples/data_quality/pandera/feature_logic.py
+++ b/examples/data_quality/pandera/feature_logic.py
@@ -95,7 +95,7 @@ def seasons_encoded__base(seasons: pd.Series) -> pd.DataFrame:
     3 - third season
     4 - fourth season
     """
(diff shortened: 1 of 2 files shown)


skrawcz (commented):
Awesome thanks for catching and fixing. One quick question -- is this backwards compatible with pandas <2.X? If not, we could just bump the pandas minimum for the example.

skrawcz:
Actually just looked at the docs. The change in `get_dummies` was going from a default of unit8 to bool. So this will be backwards compatible with prior pandas versions.

skrawcz (approved):

skrawcz:
FYI - https://github.com/DAGWorks-Inc/hamilton/pull/229

PR #243, opened 2023-07-30
Starts a contrib module to capture LLM DAGs

skrawcz:
This one specifically brings in the text summarization example code and makes it more specific, that given some text, you can summarize it.

The next extension would be to connect to a vector store to retrieve docs/look up which ones to summarize...

## Usage example
This would mirror what one does with Langchain:
```python
from langchain ... import NAME_OF_CHAIN 

llm = ChatOpenAI(temperature=0, model="gpt-3.5-turbo-0613")

chain = NAME_OF_CHAIN(llm)

result = chain.run(question=question, context=context)
```
This is what it would look like to an end user to use:
```python
from hamilton.contrib.... import NAME_OF_FLOW

dr = driver.Driver({"temperature":0, "model":"gpt-3.5-turbo-0613"}, NAME_OF_FLOW)

result = dr.execute(["facts"], inputs={"question": question, "context": context})
```

## Changes
 - adds contrib module
 - adds text summarization flow

## How I tested this 
 - locally

## Notes / Things to think about
 - Rather than having a new repo, I thought putting it into Hamilton would be fine for now. 
 - naming -- can we standardize so these flows can connect? or is that too hard?

## Checklist

### DIFF
--- a/hamilton/contrib/llm_dags/text_summarization.py
+++ b/hamilton/contrib/llm_dags/text_summarization.py
@@ -1,0 +1,179 @@
+import concurrent
+from typing import Generator, List
+
+import openai
+import tiktoken
+from PyPDF2 import PdfReader
+from tenacity import retry, stop_after_attempt, wait_random_exponential
+from tqdm import tqdm
+
+from hamilton.function_modifiers import config
+
+
+def summarize_chunk_of_text_prompt(content_type: str = "an academic paper") -> str:
+    """Base prompt for summarizing chunks of text."""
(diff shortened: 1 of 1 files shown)


skrawcz:
TODOs:
 1. make this run in a fastapi.
 2. make this run on spark.

skrawcz:
also probably want to figure out a way to evolve the package/changes/complexity of implementation? 

Randomly this also made me think about making our examples pip installable too...

skrawcz:
This has been superseded -- see contrib/ folder.

PR #251, opened 2023-08-03
Adds DaskDataFrameResult

skrawcz:
Adds a DaskDataFrameResult to help make it
clearer what's going on if someone wants to bring in dask to scale to larger datasets.

This also tidies up the dask hello world by introducing more run_* files. Since there are now effectively three ways you can use dask:
 - compute out non dask stuff via delayed
 - use dask datatypes
 - use both together

So I try to make it clear what is going on when you're using dask.delayed versus changing the inputs to be
dask dataframes/series.

## Changes
 - h_dask -- augments the DaskGraphAdapter, and adds DaskDataFrameResult.
 - dask hello world examples
 - updates tests
 - adds to docs

## How I tested this
 - locally

## Checklist


skrawcz:
## This information would be useful to debug
 - what are the datatypes of your dataframe?
 - do you know what is causing the memory to increase?
 - can you run the pandas version with the `logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)` set? It should tell you at what function things die during computation: e.g. search for `Computing XXXX`.
 
## What I think you should try:
1. Upgrade to pandas 2+
2. Turn on the pyarrow backend to read data
    - with pandas try https://pandas.pydata.org/docs/user_guide/pyarrow.html#i-o-reading
3. are you using strings? (assume pandas 2+)
   - with dask https://www.coiled.io/blog/pyarrow-strings-in-dask-dataframes 
4. Install this branch and don't use delayed and use the dask dataframe result builder (see run.py)
   -  `pip install  git+https://github.com/dagworks-inc/hamilton.git@dask_additions`
   -  try what's shown in `run.py` -- add `dask.config.set({"dataframe.convert-string": True})`

PR #264, opened 2023-08-10
Materialization improvements

elijahbenizzy:

## Checklist


examples/materialization/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+This will be a high-level overview. For more details,
+see [documentation](https://hamilton.dagworks.io/en/latest/reference/drivers/Driver/#hamilton.driver.Driver.materializehttps://hamilton.dagworks.io/en/latest/reference/drivers/Driver/#hamilton.driver.Driver.materialize).
+
+`driver.materialize()` does the following:
+1. Processes a list of materializers to create a new DAG
+2. Alters the output to include the materializer nodes
+3. Processes a list of "additional variables" (for debugging) to return intermediary data
+4. Executes the DAG, including the materializers
+5. Returns a tuple of (`materialization metadata`, `additional variables`)
```
skrawcz, line 37: why "additional variables"? Because that's what we use in execute? Should we rename it to "additional_outputs"?

examples/materialization/run.py
```diff
@@ -0,0 +1,87 @@
...
+        ),
+        # materialize the model to a pickle file
+        to.pickle(dependencies=["fit_clf"], id="clf_to_pickle", path="./data/clf.pkl"),
+        # materialize the predictions we made to a csv file
+        to.csv(
+            dependencies=["predicted_output_with_labels"],
+            id="predicted_output_with_labels_to_csv",
+            path="./data/predicted_output_with_labels.csv",
+        ),
+    ]
```
skrawcz, line 74: do we want `path` to be a source, and have it provided as inputs? Or at least one of these should show that pattern so that people can see what can still be dynamic.

skrawcz (approved):
Only questions is do we want to call it `additional_vars`  vs `additional_outputs` (more intuitive)... but otherwise just comments that aren't blockers.

examples/materialization/run.py
elijahbenizzy: Yeah so i can add that as an example or at least note it

examples/materialization/README.md
elijahbenizzy: Yep to mirror "final_vars", but I like "outputs" better than vars

skrawcz: I would just make it outputs personally. But 🤷 .

elijahbenizzy: On second thought we already released this and I'm a stickler for semantic versioning...

skrawcz: your call.

PR #281, opened 2023-08-19
Adds CSV materializers for polars + fix

skrawcz:
A minor fix to ensure that polars dataframe builder can be used within materializers.

Otherwise adds csv reader and writer for polars.

## Changes
 - csvreader for polars
 - csvwriter for polars
 - fix for polars dataframe builder

## How I tested this
 - locally

## Notes
 - should really now add an example ... but will create an issue for it...

## Checklist

- [ ] ~Changes are limited to a single goal (no scope creep)~ slight scope creep.

### DIFF
--- a/hamilton/plugins/polars_extensions.py
+++ b/hamilton/plugins/polars_extensions.py
@@ -1,11 +1,19 @@
-from typing import Any
+import dataclasses
+from io import BytesIO, TextIOWrapper
+from pathlib import Path
+from typing import Any, BinaryIO, Collection, Dict, Mapping, Sequence, TextIO, Tuple, Type, Union
 
 try:
     import polars as pl
 except ImportError:
     raise NotImplementedError("Polars is not installed.")
 
+from polars import PolarsDataType
+from polars.type_aliases import CsvEncoding, CsvQuoteStyle
+
 from hamilton import registry
+from hamilton.io import utils
+from hamilton.io.data_adapters import DataLoader, DataSaver
 
 DATAFRAME_TYPE = pl.DataFrame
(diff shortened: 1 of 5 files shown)


elijahbenizzy:
Add to https://github.com/DAGWorks-Inc/hamilton/blob/c5c06d07285120d351724e0f8ed0fce2e8c0f8e0/docs/data_adapters_extension.py#L24

hamilton/plugins/polars_extensions.py
```diff
@@ -29,3 +33,68 @@ def register_types():
 
 
 register_types()
+
+
+class DataFrameIO(DataLoader, DataSaver, abc.ABC):
+    """Base class for data loaders that saves/loads pandas dataframes.
```
elijahbenizzy, line 39: polars not pandas...

hamilton/plugins/polars_extensions.py
```diff
@@ -29,3 +37,186 @@ def register_types():
...
+    null_value: str = None
+    quote_style: CsvQuoteStyle = None
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return [DATAFRAME_TYPE]
+
+    def _get_saving_kwargs(self):
+        kwargs = {}
+        if self.separator is not None:
```
elijahbenizzy, line 176: You could do this with a for-loop but I guess copilot does this cleanly without one :)

elijahbenizzy: Would be nice to have a function wrapper annotation, but it requires types on their part

elijahbenizzy (approved):

PR #286, opened 2023-08-22
Moves scaling out of experimental

elijahbenizzy:

## Checklist


hamilton/registry.py
```diff
@@ -76,15 +76,7 @@ def load_extension(plugin_module: str):
...
     """
-    mod = importlib.import_module(f"hamilton.plugins.{plugin_module}_extensions")
-    assert hasattr(mod, "register_types"), "Error extension missing function register_types()"
-    assert hasattr(
-        mod, f"get_column_{plugin_module}"
-    ), f"Error extension missing get_column_{plugin_module}"
-    assert hasattr(
-        mod, f"fill_with_scalar_{plugin_module}"
-    ), f"Error extension missing fill_with_scalar_{plugin_module}"
-    logger.info(f"Detected {plugin_module} and successfully loaded Hamilton extensions.")
```
skrawcz, line 87: can we not make this conditional?
e.g. have the module declare some variable, and if that's not present we do these asserts?

skrawcz:
Need a README somewhere too (maybe in plugins) to explain how to add/adjust and name files.

skrawcz (commented):
Plugins will get large. We also don't want very large files either.

So the evolution should be into packages, right?
h_spark.py -->

 - __init__.py 
 -  ... modules organized however we want


And then similarly for *_extensions.py...?

hamilton/registry.py
elijahbenizzy: Sure, that works

elijahbenizzy:
> Plugins will get large. We also don't want very large files either.
> 
> So the evolution should be into packages, right? h_spark.py -->
> 
> * **init**.py
> * ... modules organized however we want
> 
> And then similarly for *_extensions.py...?

Yeah, I think that'll be the best way for it to evolve. Should be able to do it and keep compatible references later on, no need to break into packages yet.

skrawcz (approved):
I think things work.

PR #299, opened 2023-08-25
Fixes bug in which Collect[] couldn't take inputs not from the subdag

elijahbenizzy:
This was due to an untested codepath that was getting the wrong key from upstream.

## Checklist


tests/execution/test_executors.py
```diff
@@ -230,24 +229,20 @@ def test_get_execotor_for_tasks_default_execution_manager(purpose, check):
...
+def test_end_to_end_parallelizable_with_overrides():
     dr = (
         driver.Builder()
-        .with_modules(tmp_module)
+        .with_modules(inputs_in_collect)
         .enable_dynamic_execution(allow_experimental_mode=True)
         .with_remote_executor(SynchronousLocalTaskExecutor())
     ).build()
-    assert dr.execute(["collect"])["collect"] == 46
+    assert dr.execute(["collect"], overrides={"par": [1, 2, 3]})["collect"] == 7
```
skrawcz, line 248: this doesn't test overriding collect though?

skrawcz: can you override identity?

elijahbenizzy: Nope, you cannot override `identity` -- its in the middle of the block. Can test overriding `collect` as well.

elijahbenizzy: Yes, but I don't have the time right now to dig in on that -- this is a patch and why its still in experimental. Will create an issue.

elijahbenizzy: https://github.com/DAGWorks-Inc/hamilton/issues/301

skrawcz (approved):

PR #300, opened 2023-08-25
Create a dataflow to mimic create_citation_fuzzy_match_chain

### LINKED ISSUE #242: [good first issue - advanced][Example] Create a dataflow to mimic create_citation_fuzzy_match_chain
We need more examples to show how Hamilton can work in the LLM context.

Task - implement [create_citation_fuzzy_match_chain](https://api.python.langchain.com/en/latest/_modules/langchain/chains/openai_functions/citation_fuzzy_match.html#create_citation_fuzzy_match_chain) (or something similar) but with Hamilton.  I.e. a dataflow that expresses citing information in the response to a question.  To see the code in action see [this example](https://python.langchain.com/docs/use_cases/question_answering/how_to/qa_citations).

Put the result in the examples/LLM_Workflows folder.

rajveer43:
close #242 

## Checklist

### DIFF
--- a/examples/LLM_Workflows/off_the_shelf/fuzzy_chain.py
+++ b/examples/LLM_Workflows/off_the_shelf/fuzzy_chain.py
@@ -1,0 +1,50 @@
+from hamilton import (
+    Chain,
+    DataFrame,
+    Function,
+    Input,
+    Output,
+    Source,
+    Text,
+)
+
+
+def create_citation_fuzzy_match_chain(llm: BaseLanguageModel) -> Chain:
+  
+
+    output_parser = PydanticOutputFunctionsParser(pydantic_schema=QuestionAnswer)
+    schema = QuestionAnswer.schema()
+    function = Function(
+        name=schema["title"],
+        description=schema["description"],
+        parameters=schema,
+    )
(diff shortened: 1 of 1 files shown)


skrawcz:
@rajveer43 yep thanks for the PR.

So the task is a little more complex than this. What we want to do, is reimplement this chain, but with Hamilton constructs.

E.g. take a look at the code defined in https://github.com/DAGWorks-Inc/hamilton/blob/main/examples/LLM_Workflows/pdf_summarizer/backend/summarization.py. It shows the Hamilton like way to define a "chain" but as a dataflow in Hamilton using declarative functions.

Does that make sense? There really shouldn't be any `imports` that invoke to Langchain.

skrawcz:
> Gotcha, Just a quick PR to understand how it actually works, next commit will be a enhanced version of this example, which would be much better then this. and file locations was a bit confusing so created a new file.

Sounds good. We could leave it in for the time being to compare.


> can you give a quick understanding whether I should implement that? in a sepearate folder or in `pdf_summerizer`?

Yep, why don't you create a directory called `off_the_shelf` under LLM_Workflows, and stick this file there. We'll then potentially move this into a contrib module, or some other place so people can easily access it.

PR #309, opened 2023-08-29
[Feedback requested] Add world of warcraft example for pandas and spark

roelbertens:
I was missing a clear example on how to use Hamilton to create a feature catalog using spark.
For reference purpose I also added the same example for pandas.

I have an open question, which is how to extract the individual columns in the spark example, like I've done in the pandas one.

Another question is if I used the framework as intended.

Happy to hear your feedback!

## Checklist


roelbertens:
@elijahbenizzy thanks for the quick reply!

I tried to work out your example but here are my thoughts so far:
- `zone_features__spark_2` is not working yet because I don't pass the aggregation level as column but as string (not sure yet how to fix), and maybe I made more mistakes.
- It creates a lot of additional code and also the DAG seems to get quite crowded
- do you know if the execution speed of these two approaches would be the same? Will it find out to do all aggregations after a single groupby (e.g. for durotar_count, darkshore_count and total_count)?

elijahbenizzy:
Hey! So:
1. Fixed it for you -- the problem is that the pandas UDFs are map-only -- I think you were doing aggregations inside it which messed it up somehow.
2. Yep, so the crowdedness is an artifact of the extra functions that the parser picks up -- currently best practice is to bury the map-specific transforms in another module -- in the code below I kept them in a function as to have it all be one file. We could also think of a semantic way to hide it.
3. The same -- it was designed to be the same efficiency as repeated `withColumn` calls. Its a little complex how it works, but the TL;DR is that it takes these functions and topologically sorts them within each `with_columns` group. You write them as individuals, but the maps get repeatedly applied to the same one. Note that we're not dealing with aggregations in this -- I think we can probably come up with a good way of grouping it but for now I'd prefer to do those at once.

The design was to separate the mapping capabilities so:

1. Your maps can be expressed as a DAG
2. Your aggregations, joins, filters, etc... can all be expressed as a linear chain

Which I've seen get useful with lots of features.

Here's the DAG I have:
image

And here's the code:

```python
import pandas as pd
import pyspark.sql as ps
from pyspark.sql import functions as sf

from hamilton import htypes
from hamilton.plugins.h_spark import with_columns


def spark_session() -> ps.SparkSession:
    return ps.SparkSession.builder.master("local[1]").getOrCreate()


def world_of_warcraft(spark_session: ps.SparkSession) -> ps.DataFrame:
    return spark_session.read.parquet("data/wow.parquet")


# We just put these under functions so that they don't show up in the DAG
# You could put them in another module, but I also think we need to find a good way
# to hide them from the DAG.
def _flag_functions():
    def durotar_flag(zone: pd.Series) -> htypes.column[pd.Series, int]:
        return (zone == ' Durotar').astype(int)

    def darkshore_flag(zone: pd.Series) -> htypes.column[pd.Series, int]:
        return (zone == ' Darkshore').astype(int)

    return durotar_flag, darkshore_flag


@with_columns(
    *_flag_functions(),
    columns_to_pass=['zone']
)
def with_flags(world_of_warcraft: ps.DataFrame, aggregation_level: str) -> ps.DataFrame:
    return world_of_warcraft


def zone_counts(with_flags: ps.DataFrame, aggregation_level: str) -> ps.DataFrame:
    return (
        with_flags
            .groupby(aggregation_level)
            .agg(
            sf.count("*").alias("total_count"),
            sf.sum("darkshore_flag").alias("darkshore_count"),
            sf.sum("durotar_flag").alias("durotar_count"),
        )
    )


def _likelihood_functions():
    def durotar_likelyhood(durotar_count: pd.Series, total_count: pd.Series) -> htypes.column[pd.Series, float]:
        return durotar_count / total_count

    def darkshore_likelihood(darkshore_count: pd.Series, total_count: pd.Series) -> htypes.column[pd.Series, float]:
        return darkshore_count / total_count

    return durotar_likelyhood, darkshore_likelihood


@with_columns(
    *_likelihood_functions(),
    columns_to_pass=['durotar_count', 'darkshore_count', 'total_count']
)
def zone_likelihoods(zone_counts: ps.DataFrame) -> ps.DataFrame:
    return zone_counts
```

I'd be happy to hop on a call at some point and walk you through the design -- there are likely some pieces we can improve, and your feedback is valuable already!

roelbertens:
Thanks for the elaborate feedback! I've incorporated the changes and now there is 1 pandas example and 2 spark examples. One spark example groups the operations and therefore also keeps the DAG simple (v1), the other one is the final example you provided (v2).

They both have their advantages. 
- V1 keeps it simpler, less code, thus easier to extend.
- V2 is more clear on the exact columns that can be created (although I don't agree with the arrows in the DAG, e.g. why is `with_flags.durotar_flag` pointing to `with_flags.darkshore_flag` and not to `with_flags`?)

What do you think?

elijahbenizzy:
> Thanks for the elaborate feedback! I've incorporated the changes and now there is 1 pandas example and 2 spark examples. One spark example groups the operations and therefore also keeps the DAG simple (v1), the other one is the final example you provided (v2).
> 
> They both have their advantages.
> 
> * V1 keeps it simpler, less code, thus easier to extend.
> * V2 is more clear on the exact columns that can be created (although I don't agree with the arrows in the DAG, e.g. why is `with_flags.durotar_flag` pointing to `with_flags.darkshore_flag` and not to `with_flags`?)
> 
> What do you think?

Yeah, so I think both are quite valuable. I'd argue that, for this case, `v1` is the cleanest. That said, `v2` yields column-level lineage, which can be valuable, especially when you have quite. afew transformations, and your map functions form a feature catalog of sorts. E.G. you can imagine a structure like this:

1. A few data sources
2. They all get joined, filtered
3. A few aggregations
4. A ton (5+, but more like 100+) of map functions (some of which are just identity, and others are more complicated)

In that case the more complex DAG could be what you want -- you can unit test everything in (4). The really cool thing is that, if they're pandas UDFs, you can take those and rerun in an online context, reusing the majority of the code.

Your point about extra lines is super astute -- this is a TODO. What's happening is that the operator effectively "compiles" the transformations, running them in a linear order (e.g. `withColumn` repeated, calling out the UDFs that you define). Thus you can define UDFs individually and have it all run efficiently. In order to keep lineage, we preserve extra dependencies -- not just the linear ones, but also the DAG ones in the features. Opened up this to track! https://github.com/DAGWorks-Inc/hamilton/issues/316

elijahbenizzy:
OK, so this looks great. Two things I think we should add to get this merged:
1. An explanation of the trade-offs in the README between the approaches
2. We could consider breaking out these functions into separate modules (they're small, but its a much easier way to organize them and doesn't require the weird functions returning other functions): https://github.com/DAGWorks-Inc/hamilton/pull/309/files#diff-1da777fd995ac4d08eeca1731b53786533d649db6a2dea29183681933993e6a5R17.

LMK if this makes sense -- happy to help out if you need!

examples/wow/zone_features__spark_v2.py
```diff
@@ -0,0 +1,60 @@
...
+        durotar_count: pd.Series, total_count: pd.Series
+    ) -> htypes.column[pd.Series, float]:
+        return durotar_count / total_count
+
+    def darkshore_likelihood(
+        darkshore_count: pd.Series, total_count: pd.Series
+    ) -> htypes.column[pd.Series, float]:
+        return darkshore_count / total_count
+
+    return durotar_likelihood, darkshore_likelihood
```
skrawcz, line 53: ditto.

examples/wow/zone_features__spark_v2.py
```diff
@@ -0,0 +1,60 @@
...
+def _flag_functions():
+    """Hidden from the DAG by using this wrapper function"""
+
+    def durotar_flag(zone: pd.Series) -> htypes.column[pd.Series, int]:
+        return (zone == " Durotar").astype(int)
+
+    def darkshore_flag(zone: pd.Series) -> htypes.column[pd.Series, int]:
+        return (zone == " Darkshore").astype(int)
+
+    return durotar_flag, darkshore_flag
```
skrawcz, line 26: this is cute, but my preference here would be to put this into another module. That way it can be shared and reused more easily.

examples/wow/zone_features__pandas.py
```diff
@@ -0,0 +1,27 @@
...
+
+from hamilton.function_modifiers import extract_columns
+
+
+def with_flags(avatarId: pd.Series, zone: pd.Series) -> pd.DataFrame:
+    return (
+        pd.concat([avatarId, zone], axis=1)
+        .assign(darkshore_flag=lambda x: (x["zone"] == " Darkshore").astype(int))
+        .assign(durotar_flag=lambda x: (x["zone"] == " Durotar").astype(int))
+    )
```
skrawcz, line 11: The more Hamiltonian way to do this would be:

```suggestion
def darkshore_flag(zone: pd.Series) -> pd.Series:
    return pd.Series((zone == "Darkshore").astype(int), index=zone.index)

def durotar_flag(zone: pd.Series) -> pd.Series:
    return pd.Series((zone == "Durotar").astype(int), index=zone.index)

def with_flags(avatarId: pd.Series, darkshore_flag: pd.Series, durotar_flag: pd.Series) -> pd.DataFrame:
    return pd.concat([avatarId, darkshore_flag, durotar_flag], axis=1)
```

examples/wow/zone_features__pandas.py
```diff
@@ -0,0 +1,27 @@
...
+        darkshore_count=("darkshore_flag", "sum"),
+        durotar_count=("durotar_flag", "sum"),
+    )
+
+
+@extract_columns("darkshore_likelihood", "durotar_likelihood")
+def zone_likelihoods(zone_counts: pd.DataFrame) -> pd.DataFrame:
+    return zone_counts.assign(
+        darkshore_likelihood=lambda x: x["darkshore_count"] / x["total_count"]
+    ).assign(durotar_likelihood=lambda x: x["durotar_count"] / x["total_count"])
```
skrawcz, line 27: Since darkshore_count and total_count are already exposed you could simplify this to:

```suggestion
def darkshore_likelihood(darkshore_count: pd.Series, total_count:pd.Series) -> pd.Series:
      return darkshore_count /  total_count
      
def durotar_likelihood(durotar_count: pd.Series, total_count: pd.Series) -> pd.Series:
      return durotar_count / total_count

```

skrawcz (commented):
If you do this for the pandas example:

```python
dr = driver.Driver({}, zone_features) 
dr.display_all_functions(None)
```
You'll get this:
Screen Shot 2023-09-04 at 10 58 10 AM

The extra "nodes" won't show up -- so instead at execute time you'll want to do this:

```python
df = dr.execute(output_columns, inputs={**input_df.to_dict(orient='series'), "aggregation_level": aggregation_level})
```
to make it work. Generally in the "driver" you want to pass in configuration to shape the DAG, or inputs that invariant to execution.

examples/wow/zone_features__spark_v2.py
roelbertens: I made some changes, is that what you had in mind?

examples/wow/zone_features__spark_v2.py
```diff
@@ -0,0 +1,40 @@
...
+def spark_session() -> ps.SparkSession:
+    return ps.SparkSession.builder.master("local[1]").getOrCreate()
+
+
+def world_of_warcraft(spark_session: ps.SparkSession) -> ps.DataFrame:
+    return spark_session.read.parquet("data/wow.parquet")
+
+
+@with_columns(darkshore_flag, durotar_flag, columns_to_pass=["zone"])
+def with_flags(world_of_warcraft: ps.DataFrame, aggregation_level: str) -> ps.DataFrame:
```
skrawcz, line 22: aggregation_level doesn't seem to be needed here?

roelbertens: Correct, that's a mistake

skrawcz:
Looking good. I'll defer to @elijahbenizzy since you two chatted in case there was anything else being added/changed.

elijahbenizzy:
This looks great! I just moved the files to a better location.

PR #313, opened 2023-08-30
Handles overrides in a cleaner way

### LINKED ISSUE #307: Input presence checks don't account for overrides
```python
from hamilton import base, driver, ad_hoc_utils
def foo(baz: int) -> int:
    return baz + 1

def bar(foo: int) -> int:
    return foo + 1
    
mod = ad_hoc_utils.create_temporary_module(foo, bar)
dr = driver.Driver({}, mod, adapter=base.DefaultAdapter())
dr.execute(['bar'], inputs={'baz' : 10})['bar'] # 12 = 10 + 1 + 1

mod = ad_hoc_utils.create_temporary_module(foo, bar)
dr = driver.Driver({}, mod, adapter=base.DefaultAdapter())
dr.execute(['bar'], overrides={'foo' : 10}, inputs={'baz' : 10})['bar'] # 12 = 10 + 1
```

This breaks, it should get `11`

elijahbenizzy:
We just weren't passing it to the node-determining function, so raw_execute wasn't calling the validator with the right inputs.

Fixes #307

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -571,6 +571,7 @@ class Driver:
         render_kwargs: dict,
         inputs: Dict[str, Any] = None,
         graphviz_kwargs: dict = None,
+        overrides: Dict[str, Any] = None,
     ):
         """Helper function to visualize execution, using a passed-in function graph.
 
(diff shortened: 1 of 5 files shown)


hamilton/driver.py
```diff
@@ -581,21 +582,30 @@ def _visualize_execution_helper(
...
-        # _final_vars = self._create_final_vars(final_vars)
-        nodes, user_nodes = fn_graph.get_upstream_nodes(final_vars, inputs)
+        nodes, user_nodes = fn_graph.get_upstream_nodes(final_vars, inputs, overrides)
         Driver.validate_inputs(fn_graph, adapter, user_nodes, inputs, nodes)
         node_modifiers = {fv: {graph.VisualizationNodeModifiers.IS_OUTPUT} for fv in final_vars}
         for user_node in user_nodes:
             if user_node.name not in node_modifiers:
                 node_modifiers[user_node.name] = set()
             node_modifiers[user_node.name].add(graph.VisualizationNodeModifiers.IS_USER_INPUT)
+        all_nodes = nodes | user_nodes
```
skrawcz, line 592: will this work for all versions of python?

skrawcz: otherwise why not use the `.union` ?

skrawcz (approved):

elijahbenizzy:
See as an example 
image

hamilton/driver.py
elijahbenizzy: Consistent with how we were doing it in another place, and yeah, this was a python 2 feature

PR #315, opened 2023-08-31
WIP for injected kwargs

elijahbenizzy:

## Checklist


hamilton/function_modifiers/adapters.py
```diff
@@ -18,6 +19,14 @@
 from hamilton.registry import LOADER_REGISTRY, SAVER_REGISTRY
 
 
+class InjectedKwargs(enum.Enum):
+    """These are special kwargs for data adapters that are declared and automatically injected by Hamilton.
+    These can"""
+
+    node_name = "__node_name"
+    node_tags = "__node_tags"
```
skrawcz, line 27: I feel like we need a prefix, since this isn't "this" node's name, but the one it depends on / wraps.

Also in the case of multiple outputs -> combiner -> saver. We'd likely want this to be an ordered list for names, and tags to be a dictionary of name -> tags?

elijahbenizzy: Yeah this PR is very much in flux -- I've been thinking it through today and need to solve those edge cases.

zilto:
will reopen if feature is requested.

PR #317, opened 2023-09-02
Implements example to use for write-once-run-everywhere blog post

elijahbenizzy:
Note that this is similar to the more complex feature engineering post, just meant to be bite-sized and demonstrate a point.

## Checklist


skrawcz (commented):
why are you deleting the previous example?

elijahbenizzy:
> why are you deleting the previous example?

I should be -- meant to move it into the subdirectory.

hamilton/version.py
```diff
@@ -1 +1 @@
-VERSION = (1, 27, 3)
```
skrawcz, line 1: this isn't needed?

skrawcz: should go away with a rebase?

skrawcz (approved):
notes -- these don't need to be changed?
 - version
 - star counting notebook

skrawcz (approved):

PR #335, opened 2023-09-11
Adds Polars Jupyter Notebook example for e.g., data scientists

bryangalindo:
This adds a Jupyter Notebook example for e.g., data scientists to read instead of the Python code. The notebook emulates the `examples/polars/my_script.py` module.

## Changes
Adds Jupyter Notebook to the `examples/polars` directory.

## How I tested this
Successfully ran all unit tests locally using the following command in the root directory: `pytest tests/`

## Notes
This is a work in progress. I intend to add a table of contents and headings to make it easier to read.

## Checklist


examples/polars/notebook.ipynb
```diff
@@ -0,0 +1,226 @@
...
+     "name": "#%%\n"
+    }
+   }
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "outputs": [],
+   "source": [
+    "# We'll place the spend calculations into a new module\n",
```
skrawcz, line 99: @bryangalindo looks good. One edit please. Can you change this cell?

What you have works, however we want to minimize the use of adhoc_utils, since we don't want to proliferate that too much. It doesn't make it easy to go to production easily. 

Instead we either want to import `my_functions` directly (so delete this cell),  or use the `%%writefile my_functions.py` syntax at the top of the cell. In any case, the next cell would do `import my_functions` and pass that into the driver.  Thoughts? What would be more clear to a new person? Maybe the `%%writefile` option?

skrawcz (changes requested):
Looking good, just one requested change.

examples/polars/notebook.ipynb
bryangalindo: @skrawcz thanks for the feedback! A couple of thoughts come to mind:
- Who are the primary consumers of the Jupyter Notebooks (e.g., data scientists)? 
- Would the primary consumers know about the `%%writefile` option? 
- Are there other Jupyter Notebooks with the same consumers that use the `%%writefile` option? [**1**]

Even if they wouldn't be familiar with it, I like that with the `%%writefile` option, we can still expose the e.g., spend functions so it's clear what we're sending into the driver (vs `import my_functions` -> `driver(..., my_functions, ...)`. 

So, I'm leaning more towards `%%writefile`. What are your thoughts?

[**1**] I can probably look this up myself, so no need to dig into it if you don't know off the top of your head.

skrawcz: This should work
```python
%%writefile my_functions.py
# code goes here
```
Next cell
```python
# this should work because in you have loaded the auto reload extension in a prior cell.
%aimport my_functions
```
Next cell
```python
# driver code.
```

examples/polars/notebook.ipynb
```diff
@@ -193,34 +408,28 @@
     "# To visualize do `pip install \"sf-hamilton[visualization]\"` if you want these to work\n",
     "dr.visualize_execution(output_columns, './polars', {\"format\": \"png\"})\n",
     "dr.display_all_functions('./my_full_dag.dot')"
```
skrawcz, line 410: oh in a notebook this can be simplified to:
```suggestion
    "dr.visualize_execution(output_columns)\n",
    "dr.display_all_functions()"
```
though I think only one will display in jupyter at a time.

examples/polars/notebook.ipynb
```diff
@@ -0,0 +1,435 @@
...
+   ],
+   "source": [
+    "# Execute the driver.\n",
+    "\n",
+    "df = dr.execute(output_columns)\n",
+    "print(df)\n",
+    "\n",
+    "# To visualize do `pip install \"sf-hamilton[visualization]\"` if you want these to work\n",
+    "dr.visualize_execution(output_columns, './polars', {\"format\": \"png\"})\n",
+    "dr.display_all_functions('./my_full_dag.dot')"
```
skrawcz, line 410: This can be simplified to this:
```suggestion
    "dr.visualize_execution(output_columns)\n",
```
Also only need one. Jupyter only displays the last one.

skrawcz (approved):
Looks good.

examples/polars/notebook.ipynb
bryangalindo: thanks for the heads up

PR #346, opened 2023-09-13
Adds Pandas JSON data loaders

bryangalindo:
This change introduces an extension of the Pandas `read_json` method found ([here](https://pandas.pydata.org/docs/reference/api/pandas.read_json.html#pandas.read_json)) and the Pandas `to_json` method found ([here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html#pandas.DataFrame.to_json)).

## Changes
- `hamilton/plugins/pandas_extensions.py`
- `tests/plugins/test_pandas_extensions.py`

## How I tested this
- Successfully ran unit tests for py3.11 locally using the following command in the root directory: `pytest tests/`
- Successfully ran unit tests for py3.7-3.10 using docker run command found [here](https://github.com/DAGWorks-Inc/hamilton/blob/main/developer_setup.md#running-tests-in-docker).

## Notes
N/A

## Checklist

### DIFF
--- a/hamilton/plugins/pandas_extensions.py
+++ b/hamilton/plugins/pandas_extensions.py
@@ -1,36 +1,26 @@
 import abc
 import dataclasses
 import sys
+from collections.abc import Hashable
 from io import BufferedReader, BytesIO
 from pathlib import Path
-from typing import Any, Callable, Collection, Dict, List, Literal, Optional, Tuple, Type, Union
-
-from hamilton.io import utils
-from hamilton.io.data_adapters import DataLoader, DataSaver
+from typing import Any, Callable, Collection, Dict, List, Optional, Tuple, Type, Union
 
 try:
     import pandas as pd
 except ImportError:
     raise NotImplementedError("Pandas is not installed.")
 
-from pandas._libs import lib
-from pandas._typing import (
-    CompressionOptions,
-    DtypeArg,
-    DtypeBackend,
-    FilePath,
-    JSONEngine,
-    JSONSerializable,
-    ReadBuffer,
-    StorageOptions,
-    TimeUnit,
-    WriteBuffer,
-)
+from pandas._typing import Dtype
 
 from hamilton import registry
+from hamilton.io import utils
+from hamilton.io.data_adapters import DataLoader, DataSaver
 
 DATAFRAME_TYPE = pd.DataFrame
 COLUMN_TYPE = pd.Series
+
+JSONSerializable = Optional[Union[str, float, bool, List, Dict]]
 
 
 @registry.get_column.register(pd.DataFrame)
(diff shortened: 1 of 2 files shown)


hamilton/plugins/pandas_extensions.py
```diff
@@ -13,6 +13,20 @@
...
+    DtypeArg,
+    DtypeBackend,
+    FilePath,
+    JSONEngine,
+    JSONSerializable,
+    ReadBuffer,
+    StorageOptions,
+    TimeUnit,
+    WriteBuffer,
+)
```
skrawcz, line 28: so we might have to drop some of them, or do something specific based on pandas version. E.g. `TimeUnit` doesn't exist in pandas 2.0.2.

If it's a "literal" we should just make it a string type annotation to get around this.

hamilton/plugins/pandas_extensions.py
```diff
@@ -229,6 +243,157 @@ def name(cls) -> str:
...
+    """Class specifically to handle saving JSON files/buffers with Pandas.
+
+    Disclaimer: We're exposing all the *current* params from the Pandas DataFrame.to_json method.
+    Some of these params may get deprecated or new params may be introduced. In the event that
+    the params/kwargs below become outdated, please raise an issue or submit a pull request.
+
+    Should map to https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html
+    """
+
+    filepath_or_buffer: Optional[Union[FilePath, WriteBuffer[bytes], WriteBuffer[str]]] = None
```
skrawcz, line 341: also I think this shouldn't be optional or have a default of None, right?

skrawcz: otherwise in python 3.9 -- I needed to change `FilePath` to be just `str` because typing broke inspecting this class...

bryangalindo: > also I think this shouldn't be optional or have a default of None, right?

Good catch! I removed the default value but left the `Optional` type annotation for the reasons below. I can also remove the `Optional` type annotation if you prefer. ``

I was emulating the Pandas `to_json` type annotations found [here](https://github.com/pandas-dev/pandas/blob/ba1cccd19da778f0c3a7d6a885685da16a072870/pandas/io/json/_json.py#L136).

`path_or_buf: FilePath | WriteBuffer[str] | WriteBuffer[bytes] | None,`
`"""Optional[X] is equivalent to Union[X, None]."""`

skrawcz: > > also I think this shouldn't be optional or have a default of None, right?
> 
> Good catch! I removed the default value but left the `Optional` type annotation [**1**] to emulate the Pandas `to_json` type annotation [**2**]. Though, I can remove the `Optional` type annotation if you prefer.
> 
> * [**1**] `"""Optional[X] is equivalent to Union[X, None]."""`
> * [**2**] `path_or_buf: FilePath | WriteBuffer[str] | WriteBuffer[bytes] | None,` [source](https://github.com/pandas-dev/pandas/blob/ba1cccd19da778f0c3a7d6a885685da16a072870/pandas/io/json/_json.py#L136).

Ah TIL. So yeah that's because it can return just the JSON string -- https://github.com/pandas-dev/pandas/blob/ba1cccd19da778f0c3a7d6a885685da16a072870/pandas/io/json/_json.py#L210C1-L217C17. Which I don't think makes sense here for the materializer. So let's not make it optional at all, and always required `path_or_buf`.

bryangalindo: Done 😄

tests/plugins/test_pandas_extensions.py
```diff
@@ -23,3 +30,25 @@ def test_pandas_pickle(tmp_path):
 
     # correct number of files returned
     assert len(list(tmp_path.iterdir())) == 1, "Unexpected number of files in tmp_path directory."
+
+
+def test_pandas_json_reader(tmp_path: pathlib.Path) -> None:
+    file_path = "tests/resources/data/test_load_from_data.json"
```
skrawcz, line 36: did you check this file in?

skrawcz (approved):

PR #353, opened 2023-09-16
Adds Pandas DF SQL reader and writer

bryangalindo:
This change introduces an extension of the Pandas `read_sql` method found ([here](https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html)) and the Pandas `to_sql` method found ([here](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html)).

## Changes
- examples/pandas/materialization/my_script.py
- examples/pandas/materialization/notebook.ipynb
- examples/pandas/materialization/requirements.txt
- hamilton/io/utils.py
- hamilton/plugins/pandas_extensions.py
- requirements-test.txt
- tests/io/test_utils.py
- tests/plugins/test_pandas_extensions.py
- tests/resources/data/test.db

## How I tested this
- Confirmed unit tests for py3.11 passed; locally ran the following command in the root directory: `pytest tests/`
- Confirmed circleci py37-py3.11 tests all passed.

## Notes
N/A

## Checklist


bryangalindo:
checking circleci failures for py37/38

hamilton/plugins/pandas_extensions.py
```diff
@@ -11,6 +11,8 @@
 except ImportError:
     raise NotImplementedError("Pandas is not installed.")
 
+from sqlite3 import Connection
```
skrawcz, line 14: we could probably do a try catch here on import exception and provide something the works in case sqllite isn't present.. I think something like this should work:
```suggestion
try:
   from sqlite3 import Connection
except ImportError:
   Connection = object
```

hamilton/plugins/pandas_extensions.py
```diff
@@ -383,6 +385,127 @@ def name(cls) -> str:
...
+    if_exists: str = "fail"
+    index: bool = True
+    index_label: Optional[Union[Hashable, Sequence[Hashable]]] = None
+    method: Optional[Union[str, Callable]] = None
+    schema: Optional[str] = None
+
+    def __post_init__(self):
+        # might be safer than passing in mutable obj to init
+        if self.engine_kwargs is None:
+            self.engine_kwargs = {}
```
skrawcz, line 472: yeah I don't think we need this. I think _get_saving_kwargs should be able to handle it.

hamilton/io/utils.py
```diff
@@ -17,3 +17,7 @@ def get_file_metadata(path: str) -> Dict[str, Any]:
         "last_modified": os.path.getmtime(path),
         "timestamp": datetime.now().utcnow().timestamp(),
     }
+
+
+def get_sql_metadata() -> Dict[str, Any]:
+    pass
```
skrawcz, line 23: yep we need to think about what to put here?

Maybe we take the original DF and do "len" on it, and pull a few other things, and just the table details? thoughts?

bryangalindo: yep, makes sense to me. added in commit [07cc882](https://github.com/DAGWorks-Inc/hamilton/pull/353/commits/07cc88274db672f4591240f1066a6dd3426f6e58).

bryangalindo:
checking py37 [failures](https://circleci.com/gh/DAGWorks-Inc/hamilton/16467)...(╯°□°)╯︵ ┻━┻

hamilton/plugins/pandas_extensions.py
bryangalindo: TIL: I think py3.7-3.11 already includes SQLite by default, so I'll probably remove the try/except block around the SQLite3 import.

hamilton/io/utils.py
```diff
@@ -1,6 +1,13 @@
 import os
 from datetime import datetime
-from typing import Any, Dict
+from typing import Any, Dict, Union
+
+try:
+    from sqlite3 import Connection
+except ImportError:
+    Connection = object
```
skrawcz, line 8: this isn't used here?

hamilton/plugins/pandas_extensions.py
```diff
@@ -383,6 +389,122 @@ def name(cls) -> str:
...
 
+@dataclasses.dataclass
+class PandasSqlReader(DataLoader):
+    """Class specifically to handle loading SQL data using Pandas.
+
+    Disclaimer: We're exposing all the *current* params from the Pandas read_sql method.
+    Some of these params may get deprecated or new params may be introduced. In the event that
+    the params/kwargs below become outdated, please raise an issue or submit a pull request.
+
+    Should map to https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html
```
skrawcz, line 400:
```suggestion
    Should map to https://pandas.pydata.org/docs/reference/api/pandas.read_sql.html
    Requires optional pandas dependencies. See https://pandas.pydata.org/docs/getting_started/install.html#sql-databases.
```

hamilton/plugins/pandas_extensions.py
```diff
@@ -383,6 +389,122 @@ def name(cls) -> str:
...
+
+@dataclasses.dataclass
+class PandasSqlWriter(DataSaver):
+    """Class specifically to handle saving DataFrames to SQL databases using Pandas.
+
+    Disclaimer: We're exposing all the *current* params from the Pandas DataFrame.to_sql method.
+    Some of these params may get deprecated or new params may be introduced. In the event that
+    the params/kwargs below become outdated, please raise an issue or submit a pull request.
+
+    Should map to https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html
```
skrawcz, line 457:
```suggestion
    Should map to https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html
    Requires optional pandas dependencies. See https://pandas.pydata.org/docs/getting_started/install.html#sql-databases.
```

hamilton/io/utils.py
```diff
@@ -17,3 +24,28 @@ def get_file_metadata(path: str) -> Dict[str, Any]:
...
+    - the current time
+    """
+    query = query_or_table if "SELECT" in query_or_table else None
+    table_name = query_or_table if "SELECT" not in query_or_table else None
+    rows = (
+        results
+        if isinstance(results, int)
+        else len(results)
+        if isinstance(results, pd.DataFrame)
+        else None
```
skrawcz, line 44: can still do a `len` on a dataframe though?

hamilton/io/utils.py
```diff
@@ -17,3 +24,28 @@ def get_file_metadata(path: str) -> Dict[str, Any]:
...
+    - the table name (e.g., "bar")
+    - the current time
+    """
+    query = query_or_table if "SELECT" in query_or_table else None
+    table_name = query_or_table if "SELECT" not in query_or_table else None
+    rows = (
+        results
+        if isinstance(results, int)
+        else len(results)
+        if isinstance(results, pd.DataFrame)
```
skrawcz, line 43: also this needs to be `elif`

hamilton/io/utils.py
```diff
@@ -17,3 +24,28 @@ def get_file_metadata(path: str) -> Dict[str, Any]:
         "last_modified": os.path.getmtime(path),
         "timestamp": datetime.now().utcnow().timestamp(),
     }
+
+
+def get_sql_metadata(query_or_table: str, results: Union[int, pd.DataFrame]) -> Dict[str, Any]:
```
skrawcz, line 29: can you add unit tests on this please? This will help document the expected behavior.

skrawcz (commented):
Left a few comments & requests. If you could modify the example and add that commit here too that would be great please :)

hamilton/io/utils.py
bryangalindo: the second if block calls len on the dataframe! the ternary expression looks confusing, so I switched to a regular if/elif block. it should read more clearly.

hamilton/io/utils.py
bryangalindo: i have unit tests on the get sql metadata function! unless you meant the get file metadata function?

skrawcz: oh nvm! I see it. Must of missed it?

skrawcz (approved):

PR #303, opened 2023-08-26
WIP - Response to Issue #292: Pandas Pickle data reader and writer

benhhack:
Response to Issue #292 which required the implementation of classes for reading and writing Pickle files with Pandas

## Changes
Implemented the PandasPickleReader and PandasPickleWriter classes in pandas_extensions.py

## How I tested this
I looked for the polars tests and couldn't find them and am unsure how to test these classes, please provide some assistance.

## Notes
Implementation closely followed the Pandas Pickle documentation and the example Polars classes provided
## Checklist

- [x] PR has an informative and human-readable title (this will be pulled into the release notes)
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [ ] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future TODOs are captured in comments
- [ ] Project documentation has been updated if adding/changing functionality.

PR #357, opened 2023-09-18
WIP - PandasExcelReader and Writer initial implementations and tests

benhhack:

## Changes
- `pandas_extensions.py`: wrote the pandas excel reader and writer files, based off the [pandas IO docs](https://pandas.pydata.org/docs/reference/io.html#excel)
- `test_pandas_extensions.py`: tests for both of the functions
## How I tested this
Created a mock df which was written into an Excel file and then read back.
## Notes
- The Pandas Excel docs specify a bunch of other classes in addition to the `read_excel` and `to_excel` that don't deal only with data frames. Should I include these classes in the wrappers?
- converting to Excel and then back to Pandas has some issues with data type conversion. I have had to add a `dtype_dict` to specify the inputs needed but I'm not sure how feasible this will be in practice. Please advise. 
- The extensions and test files are getting a little messy with all the new contributions. Would it be too far out of scope to do a little cleanup? Also on the refactoring note, there are now two plugin packages under tests.
- The tests for pandas plugins are fairly dissimilar. Should I write the tests the way I have done here where I tested the reader and writer in one function? Or should I use the style that was employed for the JSON and SQL readers with a dummy file in the resources package?

Lot's of questions and notes, sorry about that. Hopefully, we can try solve them sequentially.
## Checklist

- [ x] Changes are limited to a single goal (no scope creep)
- [ x] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [ x] Any _change_ in functionality is tested
- [ x] New functions are documented (with a description, list of inputs, and expected output)
- [ x] Placeholder code is flagged / future TODOs are captured in comments


skrawcz:
> The Pandas Excel docs specify a bunch of other classes in addition to the read_excel and to_excel that don't deal only with data frames. Should I include these classes in the wrappers?

read_excel and to_excel are fine for now.

> converting to Excel and then back to Pandas has some issues with data type conversion. I have had to add a dtype_dict to specify the inputs needed but I'm not sure how feasible this will be in practice. Please advise.

If people need to think about this usually, then in the tests/examples putting in something that works wouldn't be unreasonable either. I assume it's only required for reading from `excel` ? 


> The extensions and test files are getting a little messy with all the new contributions. Would it be too far out of scope to do a little cleanup? Also on the refactoring note, there are now two plugin packages under tests.

What sort of clean up did you have in mind? Alphabetically grouping them? or?

> The tests for pandas plugins are fairly dissimilar. Should I write the tests the way I have done here where I tested the reader and writer in one function? Or should I use the style that was employed for the JSON and SQL readers with a dummy file in the resources package?

Yeah I was mulling over this myself. I'm fine with both. But I think we should standardize. I am inclined to say, not adding a resource file is probably simpler. So writing and then reading in one test. That way it'll exercise that what we write can be read, and we won't have to worry about "versions" of those formats (since that's not something we control).

skrawcz:
You have some python3.7 incompatibility to fix, and need to add `openpyxl` to `requirements-test.txt` to fix the tests I think.

benhhack:
> read_excel and to_excel are fine for now.

Okay perfect, will leave them as is.

> If people need to think about this usually, then in the tests/examples putting in something that works wouldn't be unreasonable either. I assume it's only required for reading from `excel` ?

Yeah, it saves fine into an excel file. Regarding the examples, I'm trying to write the `my_script.py` but am getting a ModuleNotFound error for hamilton, same as with the Jupyter Notebook from #303 . I tried installing `sf-hamilton` but it hasn't worked. Do you have any advice?

> What sort of clean up did you have in mind? Alphabetically grouping them? or? Otherwise there's plugin_tests which is different from tests/plugins. The former is more for integration testing graph adapters, the latter is for unit tests. We want to keep them separate because they test different things. But I should document that to make that clearer.

Ah, I see the difference now.

> Yeah I was mulling over this myself. I'm fine with both. But I think we should standardize. I am inclined to say, not adding a resource file is probably simpler. So writing and then reading in one test. That way it'll exercise that what we write can be read, and we won't have to worry about "versions" of those formats (since that's not something we control).

Okay, I've written the test as you suggest and will leave it that way.

benhhack:
Hey @skrawcz. I've put in the logic to the `my_script.py` file but I am getting the following error:

```AttributeError: No data materializer named: excel. Available materializers are: dict_keys(['json', 'file', 'csv', 'feather', 'parquet', 'pickle', 'sql']). If you've gotten to this point, you either (1) spelled the loader name wrong, (2) are trying to use a loader that doesnot exist (yet). For a list of available materializers, see  https://hamilton.readthedocs.io/reference/io/available-data-adapters/#data-loaders 
```

I have registered the loader in `pandas_extentions.py` so I'm not sure how to resolve this error.

docs/main.md
```diff
@@ -5,31 +5,30 @@
     <a href="https://pepy.tech/project/sf-hamilton" target="_blank"><img src="https://pepy.tech/badge/sf-hamilton" alt="Total Downloads"/></a>
 </div>
 
-
-The general purpose micro-orchestration framework for creating dataflows from python functions! That is, your single tool to express things like data, ML, LLM pipelines/workflows, and even web request logic! [Please star it here!](https://github.com/dagworks-inc/hamilton).
```
skrawcz, line 9: mind reverting this part? or do you have thoughts?

README.md
```diff
@@ -413,8 +413,6 @@ If you do not wish to participate, one can opt-out with one of the following met
 - Thierry Jean (@zilto)
 - Michał Siedlaczek (@elshize)
 - Benjamin Hack (@benhhack)
-- Bryan Galindo (@bryangalindo)
-- Jordan Smith (@JoJo10Smith)
```
skrawcz, line 417: need to revert this please

skrawcz (changes requested):
@benhhack yeah I think you had a rebase issue? You've reverted a few things.

Note: you have also replaced the XML reader and writer with the Excel one. The excel one should be a new addition. 

So there's a litte bit of work here to fix up this PR. It might be simpler to create a new branch. That is, ensure your `main` is up to date, and create a new branch and port your changes across. Then open a new PR using that branch.

benhhack:
> @benhhack yeah I think you had a rebase issue? You've reverted a few things.
> 
> Note: you have also replaced the XML reader and writer with the Excel one. The excel one should be a new addition.
> 
> So there's a litte bit of work here to fix up this PR. It might be simpler to create a new branch. That is, ensure your `main` is up to date, and create a new branch and port your changes across. Then open a new PR using that branch.
> 
> If you need help, happy to point you towards some `git` resources to help.

Yup it seems I've made a mess of this. Am going to cut my losses and refork because I was working from my main branch. Expect a new PR soon.

PR #374, opened 2023-09-22
Garbage collection/memory optimization

### LINKED ISSUE #373: Garbage Collection for Parallelism
Parallelism/task-based dynamic execution appears to hold onto all of the intermediate information. This is suboptimal for i/o driven workflows that handle large data but only return small pieces.

Garbage collect all intermediate data. Should be as simple as pruning this to just the results we need? https://github.com/DAGWorks-Inc/hamilton/blob/5ab636068c89e759acb97a28554bc04a63b39569/hamilton/execution/executors.py#L80

...

**Additional context**

elijahbenizzy:
This deletes unused references in parallelism/dynamism. That said, the python garbage collector occasionally has to be told what to do, so the user should be aware.

Fixes #373

## Checklist

### DIFF
--- a/hamilton/execution/executors.py
+++ b/hamilton/execution/executors.py
@@ -79,7 +79,10 @@ def _modify_callable(node_source: node.NodeType, callabl: Callable):
 
 def base_execute_task(task: TaskImplementation) -> Dict[str, Any]:
     """This is a utility function to execute a base task. In an ideal world this would be recursive,
-    but for now we just call out to good old DFS.
+    (as in we can use the same task execution/management system as we would otherwise)
+    but for now we just call out to good old DFS. Note that this only returns the result that
+    a task is required to output, and does not return anything else. It also returns
+    any overrides.
 
     We should probably have a simple way of doing this for single-node tasks, as they're
     going to be common.
@@ -97,12 +100,21 @@ def base_execute_task(task: TaskImplementation) -> Dict[str, Any]:
         if not getattr(node_, "callable_modified", False):
             node_._callable = _modify_callable(node_.node_role, node_.callable)
         setattr(node_, "callable_modified", True)
-    return execute_subdag(
+    out = execute_subdag(
         nodes=task.nodes,
         inputs=task.dynamic_inputs,
         adapter=task.adapters[0],  # TODO -- wire through multiple graph adapters
         overrides={**task.dynamic_inputs, **task.overrides},
     )
+    # This selection is for GC
+    # We also need to get the override values
+    # This way if its overridden we can ensure it gets passed to the right one
+    final_retval = {
+        key: value
+        for key, value in out.items()
+        if key in task.outputs_to_compute or key in task.overrides
+    }
+    return final_retval
 
 
 class SynchronousLocalTaskExecutor(TaskExecutor):
--- a/hamilton/execution/graph_functions.py
+++ b/hamilton/execution/graph_functions.py
@@ -114,6 +114,7 @@ def execute_subdag(
         overrides = {}
     if computed is None:
         computed = {}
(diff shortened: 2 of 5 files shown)


hamilton/execution/executors.py
```diff
@@ -97,12 +97,21 @@ def base_execute_task(task: TaskImplementation) -> Dict[str, Any]:
...
     )
+    # This selection is for GC
+    # We also need to get the override values
+    # This way if its overridden we can ensure it gets passed to the right one
+    final_retval = {
+        key: value
+        for key, value in out.items()
+        if key in task.outputs_to_compute or key in task.overrides
+    }
+    return final_retval
```
skrawcz, line 114: Yeah you just need to update the doc string in this function to say you're only returning the subset of nodes that are in the output or are part of overrides.

elijahbenizzy:
This works nicely. See this script:


```python
import numpy as np
import pandas as pd

from hamilton import driver
from hamilton.ad_hoc_utils import create_temporary_module
from hamilton.function_modifiers import parameterize, source

NUM_ITERS = 100

import pandas as pd
import numpy as np


def foo_0(memory_size: int = 100_000_000) -> pd.DataFrame:
    """
    Generates a large DataFrame with memory size close to the specified memory_size_gb.

    Parameters:
    memory_size_gb (float): Desired memory size of the DataFrame in GB. Default is 1 GB.

    Returns:
    pd.DataFrame: Generated DataFrame with approximate memory usage of memory_size_gb.
    """
    # Number of rows in the DataFrame
    num_rows = 10 ** 6

    # Calculate the number of columns required to make a DataFrame close to memory_size_gb
    # Assuming float64 type which takes 8 bytes
    bytes_per_row = 8 * num_rows
    target_bytes = memory_size
    num_cols = target_bytes // bytes_per_row

    # Create a DataFrame with random data
    data = {f'col_{i}': np.random.random(num_rows) for i in range(int(num_cols))}
    df = pd.DataFrame(data)

    # Print DataFrame info, including memory usage
    print(df.info(memory_usage='deep'))
    return df


@parameterize(
    **{f"foo_{i}" : {"foo_i_minus_one" : source(f"foo_{i-1}")}  for i in range(1,NUM_ITERS)}
)
def foo_i(foo_i_minus_one: pd.DataFrame) -> pd.DataFrame:
    print(f"foo_{i}")
    return foo_i_minus_one * 1.01

if __name__ == '__main__':
    mod = create_temporary_module(foo_i, foo_0)
    dr = driver.Builder().with_modules(mod).build()
    output = dr.execute(
        [f"foo_{NUM_ITERS-1}"],
        inputs=dict(memory_size=1_000_000_000)
    )
```

Using [mprof](https://pypi.org/project/memory-profiler/), we can easily run it, and get the following plot. My computer dies without that code:

image

elijahbenizzy:
If you comment out the new code, you'll notice that things act strangely. The caveat here is that mprof doesn't record swap space. So, you see a large alocation, then things start to act quite junky (my computer's music skips 😆). Note this is 100mb increments instead of 1gb. 1gb will just give up.

image

Now, with the code added, it caps out at 500mb before GC -- I think this is storing a few pointers to the node + general python memory usage.  Its keeping a few extra ones around -- Hard to tell why, as there are likely more than one reference being kept around due to the way the `del` function works with dicts/when GC is happening:

image

hamilton/version.py
```diff
@@ -1 +1 @@
-VERSION = (1, 30, 1)
```
skrawcz, line 1: I think we have enough pushed that 1.32.0 is warrented.

skrawcz (approved):

skrawcz (approved):

PR #376, opened 2023-09-22
🚧 Adds Pandas DF GBQ reader and writer

bryangalindo:
🚧  WIP 🚧 

## Checklist


bryangalindo:
Need to add google auth lib to requirements-test.txt to fix test failures

hamilton/plugins/pandas_extensions.py
```diff
@@ -4,24 +4,28 @@
...
     from typing import Literal
 except ImportError:
     from typing_extensions import Literal
 
 try:
-    from collections.abc import Iterable, Sequence
+    import pandas_gbq
 except ImportError:
-    from collections import Sequence, Iterable
+    raise NotImplementedError("pandas_gbq is not installed.")
```
skrawcz, line 23: let's place this check in the dataclass itself. Ideally in the "constructor" portion, i.e. when someone goes to create the materializer we should check the import. E.g. as a post init thing  https://docs.python.org/3/library/dataclasses.html#post-init-processing.

requirements-docs.txt
```diff
@@ -7,8 +7,9 @@ gitpython # Required for parsing git info for generation of data-adapter docs
 grpcio-status
 mock==1.0.1 # read the docs pins
 myst-parser==0.18.1  # latest version of myst at this time
-pillow
 # required for all the plugins
+pandas-gbq
+pillow
```
skrawcz, line 12: why are these needed?

hamilton/plugins/pandas_extensions.py
skrawcz: We can then do the "import" within the `load_data` and `save_data` methods to remove the need to do it here.

hamilton/plugins/pandas_extensions.py
```diff
@@ -837,22 +841,151 @@ def name(cls) -> str:
         return "html"
 
 
+@dataclasses.dataclass
+class PandasGbqReader(DataLoader):
```
skrawcz, line 845: Also let's rename this to `PandasGBQReader` since in GBQ is an acronym.

hamilton/plugins/pandas_extensions.py
```diff
@@ -837,22 +841,151 @@ def name(cls) -> str:
...
+        metadata = utils.get_df_metadata(df)
+        return df, metadata
+
+    @classmethod
+    def name(cls) -> str:
+        return "gbq"
+
+
+@dataclasses.dataclass
+class PandasGbqWriter(DataSaver):
```
skrawcz, line 913: Also let's rename this to `PandasGBQWriter` since in GBQ is an acronym.

requirements-docs.txt
skrawcz: if we move the import in `pandas_extensions.py` I think we will no need `pandas-gbq` here.

tests/plugins/test_pandas_extensions.py
```diff
@@ -123,3 +126,23 @@ def test_pandas_html_writer(tmp_path: pathlib.Path) -> None:
...
+    mock_conn.run_query.return_value = df
+
+    with mock.patch("pandas_gbq.gbq.GbqConnector", return_value=mock_conn):
+        writer = PandasGbqWriter("project-id.dataset.table", project_id="test")
+        kwargs1 = writer._get_saving_kwargs()
+        writer.save_data(df)
+
+        reader = PandasGbqReader("SELECT foo FROM bar", max_results=1)
+        kwargs2 = reader._get_loading_kwargs()
+        df2, metadata = reader.load_data(pd.DataFrame)
```
skrawcz, line 142: yeah this works.
What you'd usually do afterwards is check that the mock was called appropriately. E.g. assert that the right calls were made and they were passed in the right objects.

requirements-docs.txt
bryangalindo: i'm not sure, it looks like it was previously in the .txt file, but one of the pre-commit hooks alphabetized it? 
i can move the `pandas-gbq` library import to `pandas_extensions.py`.

skrawcz: 🤦  lol. oh now I see it. Thanks

tests/plugins/test_pandas_extensions.py
bryangalindo: I added an assert_called_once call in the unit test. Thanks for the feedback,  TIL 💡. Let me know if you think it makes sense to call any other assertion methods!

PR #379, opened 2023-09-24
Adding code to remove the table named test in the script and notebook…

JoJo10Smith:
This PR is in response to #380 and adds additional code to `my_script.py` and `notebook.ipynb` files to make future development easier and stop causing `ValueErrors`.

## Changes
I am proposing changes to both the `my_script.py` and the `notebook.ipynb` file to drop the test table so that is the file is run again the user does not  inadvertently cause a value error as shown below:

image

The error is caused by the to.sql in the list of materializers since it would start a connection with the database named `df.db` (as intended), create the `test` table (as intended) and the rest of the program would run. When you run the code again it reconnects to the `df.db` and tries to create the test table again. 

When the connection to the `df.db` is closed it does not remove any tables that were created so you need to remove any tables you plan on creating. 

## How I tested this
This code was tested by running the script multiple times and there was no value error caused since every time the script would start it would remove the table names test before attempting to create it.

## Notes
This change is minimal and will have no impact on those using Hamilton who just need to see an example. This will however stop value errors from occuring when running the examples multiple times which I would assume mostly occurs during development when you are testing a feature. 

## Checklist


skrawcz (approved):

skrawcz:
thanks for fixing @JoJo10Smith !

JoJo10Smith:
No problem @skrawcz, I'll close the issue.

Thanks
Jordan

Issue #384, opened 2023-09-26
pandas feather (need to expose all params)


skrawcz:
Adjust the existing feather implementation to match https://pandas.pydata.org/docs/reference/io.html#feather as a data loader & saver.

1. Modify the implementation in pandas_extensions.py, one class for the writer, on class for the reader, e.g. much like how the [pandas pickle reader and writer](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/plugins/pandas_extensions.py) are structured.
2. There should be requisite tests to exercise the functionality. We should move to using a single test if possible -- one that writes and then reads.
3. Add to the pandas materializer example showing writing of a pandas dataframe.
4. If there's an issue with type hints, let me know and we can chat through it.

JoJo10Smith:
Taking this ticket!

skrawcz:
merged!

PR #385, opened 2023-09-26
Pandas Feather Reader and Writer Class

JoJo10Smith:
This PR is in response to #384 where I will be adding additional parameters to the Feather functionality. I have gone ahead and set the Read and Write classes to match the other materializes (PandasFeatherReader and PandasFeatherWriter). 

Pandas `to_feather`: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_feather.html
Additional Parameters: https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html#pyarrow.feather.write_feather
Pandas `read_feather`: https://pandas.pydata.org/docs/reference/api/pandas.read_feather.html

## Changes
`pandas_extensions.py:` Added the functions to read and write Feather files
`my_script.py`: Added the example of Feather materialization and optional to add Hamilton to PATH with directory
`notebook.ipynb`: Added the example of Feather materialization and optional to add Hamilton to PATH with directory
`test_pandas_extensions.py`: Added test cases for the read_feather and to_feather functions that are called PandasFeatherReader and PandasFeatherWriter respectively
`test_load_from_data.feather`: A simple Feather file with example data to test functionality
`requirements-test.txt`: Added the lz4 compression module which is required for compression 

## How I tested this

Code passed all the CIrcleCI tests and added examples of how to implement the new functionality. The new DAG looks as follows: 

image

## Notes
You will now be required to import the lz4 module for compression. 
`lz4`: https://pypi.org/project/lz4/
I added this to the `requirements-test.txt` and NOT the `requirements.txt
`
## Checklist


skrawcz:
```python
hamilton/plugins/pandas_extensions.py:27: in <module>
    from pandas._typing import FilePath, NpDtype, ReadBuffer, StorageOptions, WriteBuffer
E   ImportError: cannot import name 'FilePath' from 'pandas._typing' (/home/circleci/venvs/hamilton-venv/lib/python3.7/site-packages/pandas/_typing.py)
```
3.7 pandas typing issue.

hamilton/plugins/pandas_extensions.py
```diff
@@ -24,7 +24,7 @@
 
 from sqlite3 import Connection
 
-from pandas._typing import NpDtype
+from pandas._typing import FilePath, NpDtype, ReadBuffer, StorageOptions, WriteBuffer
```
skrawcz, line 27: For python 3.7 compatibility it might be simpler to drop the ones that aren't in pandas + python 3.7?

JoJo10Smith: I have removed it. Thought this was a better implementation so I tried it but failed the Python 3.7 check.

JoJo10Smith:
Core classes and tests have been added for PandasFeatherReader and the PandasFeatherWriter. Just to confirm @skrawcz I think it would be best to keep the `FeatherDataLoader` as there might already be some implementations of this class in user's code. I would suggest having both the `FeatherDataLoader` as well as the new classes and then deprecating the `FeatherDataLoader` at a later stage.

TODO: add materializer examples.

JoJo10Smith:
@skrawcz Depending on your response this PR could be merged. I opted to not remove the current `FeatherDataLoader `class as described above. If you would like me to remove it then I can make that change but believe it is best to keep it and deprecate later. I could add a note in the docstring to use the `PandasFeatherReader `and `PandasFeatherWriter` instead.

skrawcz:
> Core classes and tests have been added for PandasFeatherReader and the PandasFeatherWriter. Just to confirm @skrawcz I think it would be best to keep the `FeatherDataLoader` as there might already be some implementations of this class in user's code. I would suggest having both the `FeatherDataLoader` as well as the new classes and then deprecating the `FeatherDataLoader` at a later stage.

@JoJo10Smith yes that's generally best practice. But luckily that's not required here as `path` is the only parameter required as input to your implementation and the old one. The way things are designed is that no one "directly" depends on these classes. So you can go ahead and delete the old class, and everyone else's code should still just work!

requirements-test.txt
```diff
@@ -1,5 +1,6 @@
 graphviz
 lxml
+lz4
```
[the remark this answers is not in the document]
skrawcz: All good then. No need to add it. Thanks for the clarification.

skrawcz (approved):

PR #430, opened 2023-10-03
Fixes inconsistency in new builder API

skrawcz:
The default should be a dictionary adapter. In the case that you were not using the "V2" functionality there was a typo and it would not set the adapter to the dictionary one, instead passing in None, and therefore defaulting to the dataframe adapter.

## Changes
 - driver.py
 - test_hamilton_driver.py

## How I tested this
 - locally ran unit tests

## Notes
 - AFAICT nothing assumed this bug as a feature.

## Checklist


elijahbenizzy (approved):
LGTM

PR #431, opened 2023-10-03
Enables the materializers to depend on functions & driver variables

skrawcz:
.execute() enables that. The materializers should also be able to handle it.

The complication however is on who knows what and when to do it. So I made the decision to make the driver sanitize the materializers given the context that it has. New objects are returned so we don't modify the underlying materializer object -- that way if someone is reusing a materializer definition they can do so in other driver runs/etc.

## Changes
 - driver.py
 - materialization.py
 - common
 - adapters

## How I tested this
 - locally in a notebook
 - unit tests

## Notes
 - Ready for review.

## Checklist

- [x] Changes are limited to a single goal (no scope creep) [there is scope creep but it's in each commit]

### DIFF
--- a/hamilton/common/__init__.py
+++ b/hamilton/common/__init__.py
@@ -1,0 +1,58 @@
+# code in this module should no depend on much
+from typing import Any, Callable, List, Optional, Set, Tuple, Union
+
+
+def convert_output_value(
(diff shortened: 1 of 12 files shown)


elijahbenizzy (approved):
LGTM!

elijahbenizzy (approved):
LGTM!

PR #459, opened 2023-10-13
Added DataLoader and DataSaver for XGBoost models using JSON

zilto:
Added DataLoader and DataSaver for XGBoost models using JSON

## Changes
- added the DataLoader and DataSaver to a plugin `xgboost_extensions.py`
- currently supports Scikit-Learn API models (any subclass of `xgboost.XGBModel`), which have scikit-learn has a dependency
- JSON is preferrable to pickle because it makes the saved model independent from the xgboost package version
- The class `xgboost.Booster` supports the same `save_model()` and `load_model()` API, but has no dependencies with scikit-learn
- When trying to save an unfitted model, it raises an UnfittedError. I catch it in a try/except and return a dictionary with the error and message instead. It could be preferable to make the DAG fail to make the user aware of the issue more transparently

## How I tested this
- added a test file with cases for both `xgboost.Booster` and `xgboost.XGBModel` types

## Checklist

- [x ] PR has an informative and human-readable title (this will be pulled into the release notes)
- [x ] Changes are limited to a single goal (no scope creep)
- [x ] Any _change_ in functionality is tested

### DIFF
--- a/hamilton/plugins/xgboost_extensions.py
+++ b/hamilton/plugins/xgboost_extensions.py
@@ -17,8 +17,8 @@ from hamilton import registry
 from hamilton.io import utils
 from hamilton.io.data_adapters import DataLoader, DataSaver
 
-XGBOOST_MODEL_TYPES = [xgboost.XGBModel, xgboost.Booster]
-XGBOOST_MODEL_TYPES_ANNOTATION = Union[xgboost.XGBModel, xgboost.Booster]
+XGBOOST_MODEL_TYPES = [xgboost.XGBModel, xgboost.Booster, xgboost.XGBClassifier]
+XGBOOST_MODEL_TYPES_ANNOTATION = Union[xgboost.XGBModel, xgboost.Booster, xgboost.XGBClassifier]
 
 
 @dataclasses.dataclass
@@ -77,3 +77,5 @@ def register_data_loaders():
 
 
 register_data_loaders()
+
+COLUMN_FRIENDLY_DF_TYPE = False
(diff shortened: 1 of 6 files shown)


hamilton/plugins/xgboost_extensions.py
```diff
@@ -17,8 +17,8 @@
 from hamilton.io import utils
 from hamilton.io.data_adapters import DataLoader, DataSaver
 
-XGBOOST_MODEL_TYPES = [xgboost.XGBModel, xgboost.Booster]
-XGBOOST_MODEL_TYPES_ANNOTATION = Union[xgboost.XGBModel, xgboost.Booster]
+XGBOOST_MODEL_TYPES = [xgboost.XGBModel, xgboost.Booster, xgboost.XGBClassifier]
+XGBOOST_MODEL_TYPES_ANNOTATION = Union[xgboost.XGBModel, xgboost.Booster, xgboost.XGBClassifier]
```
zilto, line 21: XGBoost documentation is hard to navigate but here are all the classes:
https://xgboost.readthedocs.io/en/stable/python/index.html

- xgboost.Booster is the core object; it is never used directly. It should be supported for users that don't want any other dependencies
- xgboost.XGBModel is the basis of the sklearn API (what most people use)
- xgboost.[DaskScikitLearnBase ](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/dask/__init__.py#L1608)is the basis of the dask API. It subclasses xgboost.XGBModel
- _SparkXGBEstimator is the basis of the spark API, which is experimental. It subclasses pyspark.ml classes. It has dedicated reader and writers to interact with Booster objects [ref](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/spark/core.py#L1542)

skrawcz (approved):

PR #462, opened 2023-10-13
added data loader and saver for LightGBM

zilto:
Same logic as XGBoost PR

## Changes
- LightGBM has 3 API: Model (sklearn compliant), booster, and cross validation booster
- LightGBM uses raw text to store the booster
- A cross validation booster is a list of regular booster

## Checklist

### DIFF
--- a/hamilton/plugins/lightgbm_extensions.py
+++ b/hamilton/plugins/lightgbm_extensions.py
@@ -1,0 +1,95 @@
+import dataclasses
+from pathlib import Path
+from typing import Any, Collection, Dict, Literal, Optional, Tuple, Type, Union
+
+try:
+    import lightgbm
+except ImportError:
+    raise NotImplementedError("LightGBM is not installed.")
+
+try:
+    from sklearn.exceptions import NotFittedError
+except ImportError:
+    raise NotImplementedError("scikit-learn is not installed.")
+
+
+from hamilton import registry
+from hamilton.io import utils
+from hamilton.io.data_adapters import DataLoader, DataSaver
+
+LIGHTGBM_MODEL_TYPES = [lightgbm.LGBMModel, lightgbm.Booster, lightgbm.CVBooster]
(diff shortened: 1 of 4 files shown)


hamilton/plugins/lightgbm_extensions.py
```diff
@@ -0,0 +1,100 @@
...
+                )
+            except NotFittedError as e:
+                return utils.get_exception_details(e)
+        else:
+            data.save_model(
+                filename=self.path,
+                num_iteration=self.num_iteration,
+                start_iteration=self.start_iteration,
+                importance_type=self.importance_type,
+            )
```
skrawcz, line 59: not knowing the API, would it make more sense to check if `data` has an attribute `booster_`  versus the class? Also it seems counterintuitive that you're not checking for the `Booster` model.

Otherwise it seems like we should fail if it's not fitted, rather than raising an error? what's your rationale there?

zilto: Regarding failure, it's a question I raised in the [XGBoost PR](https://github.com/DAGWorks-Inc/hamilton/pull/459#issue-1942443037) (see initial message). Currently, the committed XGBoost extension works like this LightGBM one and propagates the package `NotFittedError` through an exception message (instead of returning metadata). It is difficult to anticipate all usage patterns, but the most likely scenario for triggering `NotFittedError` would be an incorrect dependency specified in the materializer definition (e.g., trying to materialize `base_model` instead of `trained_model`). 

IMO, the ideal approach is unclear. On one hand, if you are writing code to create and store models you would want it to fail. On the other hand, if you wrote a DAG that executes properly, adding a materializer shouldn't break your code. If materialization nodes are always executed at the end, the DAG should be able to complete before the materializer fails.

skrawcz: will that NotFittedError be raised on access to the `booster_` attribute? Otherwise looks good.

skrawcz (approved):

skrawcz:
@zilto you just need to add registering the lighgbm module. Otherwise good to go.

skrawcz:
and we should also add an example.

skrawcz (approved):

Issue #450, opened 2023-10-08
Polars: reada/write spreadsheet materializer


skrawcz:
Implement https://pola-rs.github.io/polars/py-polars/html/reference/io.html#spreadsheet as a data saver/loader. i.e. [Read](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.read_excel.html#polars.read_excel) & [write](https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.DataFrame.write_excel.html#polars.DataFrame.write_excel) to a spreadsheet.

1. It should go into polars_extensions.py, one class for the writer, on class for the reader, e.g. much like how the [polars csv reader and writer ](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/plugins/polars_extensions.py#L57-L234) are structured.
2. There should be requisite tests to exercise the functionality. We can do one test that writes & then reads it's own output.
3. Then an example should be added to the existing materialization example under [example/polars](https://github.com/DAGWorks-Inc/hamilton/tree/main/examples/polars/materialization).
4. If there's an issue with type hints, let me know and we can chat through it.

skrawcz:
Completed. Thanks @swapdewalkar  🚀 ! Let me know if you want to take something else on!

swapdewalkar:
@skrawcz Yes, let me know where I can help!

skrawcz:
> @skrawcz Yes, let me know where I can help!

Sure. Maybe something from https://github.com/DAGWorks-Inc/hamilton/issues/786? Or perhaps there's another issue you're interested in?

PR #471, opened 2023-10-18
#450: Polars spreadsheet IO along with example and test case.

swapdewalkar:
PR address https://github.com/DAGWorks-Inc/hamilton/issues/450

## Changes
Polars spreadsheet IO along with example and test case.
## How I tested this
- Unit test
- materialize example
- script
## Checklist


hamilton/plugins/polars_extensions.py
```diff
@@ -15,6 +15,7 @@
     Tuple,
     Type,
     Union,
+    Literal,
```
swapdewalkar, line 18: Any suggestion about this, as it is failing test cases.

skrawcz: yep we need to gate this:
```python
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```

hamilton/plugins/polars_extensions.py
```diff
@@ -499,6 +507,146 @@ def name(cls) -> str:
...
+            kwargs["hide_gridlines"] = self.hide_gridlines
+        if self.sheet_zoom is not None:
+            kwargs["sheet_zoom"] = self.sheet_zoom
+        if self.freeze_panes is not None:
+            kwargs["freeze_panes"] = self.freeze_panes
+        return kwargs
+
+    def save_data(self, data: DATAFRAME_TYPE) -> Dict[str, Any]:
+        data.write_excel(self.workbook, **self._get_saving_kwargs())
+        return utils.get_file_metadata(self.workbook)
```
skrawcz, line 643: this might not work if it's a workbook object. So probably want to check the type here/pull out some information if possible.

tests/plugins/test_polars_extensions.py
```diff
@@ -77,6 +79,24 @@ def test_polars_feather(tmp_path: pathlib.Path) -> None:
     assert metadata["path"] == file_path
 
 
+def test_polars_spreadsheet(df: pl.DataFrame, tmp_path: pathlib.Path) -> None:
+    file = tmp_path / "test.excel"
```
skrawcz, line 83: this will break.

tests/plugins/test_polars_extensions.py
```diff
@@ -77,6 +79,24 @@ def test_polars_feather(tmp_path: pathlib.Path) -> None:
     assert metadata["path"] == file_path
 
 
+def test_polars_spreadsheet(df: pl.DataFrame, tmp_path: pathlib.Path) -> None:
```
skrawcz, line 82: can you add a test that uses 
```python
from xlsxwriter import Workbook
with Workbook(tmp_path + "/test2.execl") as wb:  
      writer = PolarsSpreadsheetWriter(workbook=file)
      ...
```

hamilton/plugins/polars_extensions.py
```diff
@@ -23,12 +24,19 @@
...
 has_alias = False
 if hasattr(pl, "type_aliases"):
     has_alias = True
 
+
+from polars.type_aliases import ColumnFormatDict, OneOrMoreDataTypes, \
+    ConditionalFormatDict, ColumnTotalsDefinition, ColumnWidthsDefinition, RowTotalsDefinition, SelectorType, SchemaDict
+
+from polars.selectors import _selector_proxy_
+from polars.datatypes import DataType, DataTypeClass
```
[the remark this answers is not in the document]
skrawcz: actually I don't understand, why are they required if they're not referenced below?

skrawcz:
@swapdewalkar how are things going? Do you need any help?

swapdewalkar:
@skrawcz I will try to finish this week?

skrawcz:
@swapdewalkar how are things going? Do you need any help?

tests/plugins/test_polars_extensions.py
swapdewalkar: Why do this this will break?

hamilton/plugins/polars_extensions.py
```diff
@@ -499,6 +507,146 @@ def name(cls) -> str:
...
+
+
+@dataclasses.dataclass
+class PolarsSpreadsheetWriter(DataSaver):
+    """
+    Class specifically to handle saving Spreadsheet files with Polars.
+    Should map to https://pola-rs.github.io/polars/py-polars/html/reference/api/polars.DataFrame.write_excel.html
+    """
+
+    workbook: Union[Workbook, BytesIO, str, Path] = None
```
skrawcz, line 565: I think we shouldn't allow this to be None. 
I don't think the default that polars has makes sense here. So I think we want someone to always input something here to direct where to save it.

tests/plugins/test_polars_extensions.py
skrawcz: 🤔  I don't recall. 😅  But I think I might have been referring to how the code looked previously? Otherwise from reading things, you can pass a path and it'll save to it.

tests/plugins/test_polars_extensions.py
```diff
@@ -77,6 +79,24 @@ def test_polars_feather(tmp_path: pathlib.Path) -> None:
...
+    file = tmp_path / "test.excel"
+    writer = PolarsSpreadsheetWriter(workbook=file)
+    kwargs1 = writer._get_saving_kwargs()
+    writer.save_data(df)
```
skrawcz, line 86: you should save the response from `save_data()`  and validate that.

swapdewalkar:
https://github.com/DAGWorks-Inc/hamilton/pull/776

PR #474, opened 2023-10-18
Removing `NotFittedError` check from XGBoostJsonWriter

zilto:
Follow-up to this PR: https://github.com/DAGWorks-Inc/hamilton/pull/459

XGBoostJsonWriter no longer checks for NotFittedError. This is consistent with LightGBMFileWriter and removes direct dependency with scikit-learn.

## Checklist
All tests still pass


elijahbenizzy:
Seems reasonable so far -- looks like it needs a rebase though?

zilto:
I force pushed the commit for the XGBoost DataLoader/DataSaver code. Will open a separate branch for the XGBoost + Optuna contrib

PR #476, opened 2023-10-18
Added Numpy `.npy`saver/loader

zilto:

Added a NumpyNpyWriter and NumpyNpyReader using the Hamilton `file` format. This custom numpy format is efficient for multidimensional arrays which aren't directly supported by parquet for instance.

## How I tested this
Added tests to `tests/plugins/test_numpy_extensions.py` for write and read.

## Checklist

### DIFF
--- a/hamilton/plugins/numpy_extensions.py
+++ b/hamilton/plugins/numpy_extensions.py
@@ -39,7 +39,7 @@ class NumpyNpyWriter(DataSaver):
 
     @classmethod
     def name(cls) -> str:
-        return "file"
+        return "npy"
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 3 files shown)


hamilton/plugins/numpy_extensions.py
```diff
@@ -0,0 +1,87 @@
...
+            allow_pickle=self.allow_pickle,
+            fix_imports=self.fix_imports,
+            encoding=self.encoding,
+        )
+        metadata = utils.get_file_metadata(self.path)
+        return array, metadata
+
+    @classmethod
+    def name(cls) -> str:
+        return "file"
```
elijahbenizzy, line 74: Name (I think) should be `npy`, given that its a `.npy` file -- more specific to it. 

`@load_from.npy` versus `@load_from.file`.

skrawcz: ➕

skrawcz (commented):
need to add it to the registry too.

skrawcz (approved):

Issue #409, opened 2023-09-30
pandas csv (need to expose all params)


skrawcz:
Adjust the existing csv implementation to match https://pandas.pydata.org/docs/reference/io.html#flat-file (read_csv and to_csv) as a data loader & saver.

1. Modify the implementation in pandas_extensions.py, one class for the writer, on class for the reader, e.g. much like how the [pandas pickle reader and writer](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/plugins/pandas_extensions.py) are structured.
2. We will need to double check whether the new implementation is or is not backwards compatible. If it is not backwards compatible we should modify the new on so that it is, probably marking some fields as deprecated -- I can comment on the code once there is something to show you what I mean. Then we can remove the old implementation.
3. There should be requisite tests to exercise the functionality. We should move to using a single test if possible -- one that writes and then reads.
4. Add to the pandas materializer example showing writing of a pandas dataframe.
5. If there's an issue with type hints, let me know and we can chat through it.

JoJo10Smith:
I'll be working on this issue.

Thanks 
Jordan

skrawcz:
Forgot to close this one!

PR #480, opened 2023-10-21
Adding Pandas CSV Reader and Writer class

JoJo10Smith:
This PR is in response to issue #409 which requires the expansion of the current `CSVDataAdapter` so that all parameters described below can be used: 

- (Reader): https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
- (Writer): https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html

## Changes

- `pandas_extensions.py` --> I have added the PandasCSVWriter class and the PandasCSVReader class
- `test_pandas_extensions.py` --> I added tests for the functions

## Checklist

### DIFF
--- a/hamilton/plugins/pandas_extensions.py
+++ b/hamilton/plugins/pandas_extensions.py
@@ -1,4 +1,5 @@
 import abc
+import csv
 import dataclasses
 import sys
 from collections.abc import Hashable
@@ -18,9 +19,9 @@ except ImportError:
     from typing_extensions import Literal
 
 try:
-    from collections.abc import Iterable, Sequence
+    from collections.abc import Iterable, Mapping, Sequence
 except ImportError:
-    from collections import Sequence, Iterable
+    from collections import Sequence, Iterable, Mapping
 
 from sqlite3 import Connection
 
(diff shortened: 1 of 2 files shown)


JoJo10Smith:
@skrawcz just an additional update since I have made changes to a file test_adapters.py since it was no longer necessary. I've made all tests from that file into the `test_pandas_extensions.py` file.

The warning I was getting was: 
`WARNING:hamilton.function_modifiers.adapters:More than one applicable adapter detected for <class 'pandas.core.frame.DataFrame'>. Using the last one registered <class 'hamilton.plugins.pandas_extensions.PandasCsvWriter'>`

Since it was always going to default to the new classes I removed (commented out, in case you don't agree with this approach) the `CsvDataAdapter`. This also meant I could remove the tests from the `test_adapters.py` file since tests would be duplicated in the `test_pandas_extensions.py`

If you agree I can remove all the commented-out code to clean up the files.

hamilton/plugins/pandas_extensions.py
```diff
@@ -123,6 +125,266 @@ def load_data(self, type_: Type) -> Tuple[DATAFRAME_TYPE, Dict[str, Any]]:
     @classmethod
     def name(cls) -> str:
         return "csv"
+'''
```
skrawcz, line 128: I'd just delete this old implementation.

tests/function_modifiers/test_adapters.py
```diff
@@ -457,6 +458,7 @@ def df(data: pd.DataFrame) -> pd.DataFrame:
     )
     # This is just reading the same file we wrote out, so it should be the same
     pd.testing.assert_frame_equal(result["df"], result_just_read["df"])
+"""
```
skrawcz, line 461: Leave this test in. Since it's testing the `@save_to` and `@load_from` paths.

skrawcz (commented):
Looking good. Just need to delete the old one, and reinstate that test since I think it's useful. Otherwise thanks!

JoJo10Smith:
@skrawcz I've reinstated the previously commented out test and noticed something noteworthy. 

The implementation of the CSVDataAdapter had the default for the index kwarg be False when saving to csv files. According to the [pandas docs](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_csv.html) the default should be true. 

I've devastated from the default value to the implementation of the CSVDataAdapter class and set this value to False

hamilton/plugins/pandas_extensions.py
JoJo10Smith: Corrected in most recent PR

tests/function_modifiers/test_adapters.py
JoJo10Smith: Corrected in most recent PR and reinstated

skrawcz (approved):
LGTM - thanks!

PR #483, opened 2023-10-23
Decoupling io from materialization blog post

elijahbenizzy:
## Checklist

### DIFF
--- a/examples/decoupling_io/run.py
+++ b/examples/decoupling_io/run.py
@@ -1,0 +1,95 @@
+# Since this is not in plugins we need to import it *before* doing anything else
+import importlib
+from typing import List, Union
+
+import click
+
+importlib.import_module("adapters")
+
+from components import feature_data, model_evaluation, model_training
+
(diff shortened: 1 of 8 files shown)


examples/decoupling_io/run.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+modules = {
+    "features_processed": [feature_data],
+    "model_trained": [feature_data, model_training],
+    "metrics_captured": [feature_data, model_training, model_evaluation],
+}
+
+
+def get_materializers(which_stage: str):
```
skrawcz, line 27: return type annotation?

examples/decoupling_io/run.py
```diff
@@ -0,0 +1,85 @@
...
+    materializers = [
```
skrawcz, line 28: Probably worth a comment to say this set represents `metrics_captured`,  as the logic in this function works, but could be simpler.

skrawcz (approved):

PR #504, opened 2023-10-28
GeoPandas ShapeFile Reader and Writer

JoJo10Smith:
WIP: Adding the GeoPandas FIle Read and Write Class to Hamilton. 

- GeopandasFileReader: https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html
- GeopandasFileWriter:https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_file.html

## Changes
`geopandas_extensions.py` --> Adding the read and write class 
`requirements-test.txt` -> Adding additional requirements that are needed for the classes to work
`test_geopandas_extensions.py` --> Adding test for each of the files that are outputted 

## How I tested this
Added the classes and showed that they output data in a local notebook. I also wrote a single test to check that all the geodataframses that are outputted (dbf, shp, shx) which are all the memory files are the same as the original example file I used.

more info: https://en.wikipedia.org/wiki/Shapefile

## Notes
Still to do 

- [ ] Add _geopandas_ to `hamilton/function_modifiers/base.py` ??
- [ ] Add the ability to have additional arguments (kwargs) when the class is initialized
- [ ] Add a `my_functions.py` with some example functions from Geopandas
- [ ] Add a `my_script.py` which will run through all those example functions
- [ ] Create a notebook with a DAG output similar to the pandas extension 

## Checklist

### DIFF
--- a/hamilton/plugins/geopandas_extensions.py
+++ b/hamilton/plugins/geopandas_extensions.py
@@ -1,14 +1,50 @@
-from typing import Any
+import dataclasses
+import os
+from io import BufferedReader, BytesIO
+from pathlib import Path
+from typing import IO, Any, Collection, Dict, Optional, Tuple, Type, Union
+
+from pyproj import CRS
+from shapely import (
+    GeometryCollection,
+    LinearRing,
+    LineString,
+    MultiLineString,
+    MultiPoint,
+    MultiPolygon,
+    Point,
+    Polygon,
+)
 
 try:
     import geopandas as gpd
 except ImportError:
     raise NotImplementedError("geopandas is not installed.")
 
+try:
+    from typing import Literal
+except ImportError:
+    from typing_extensions import Literal
+
 from hamilton import registry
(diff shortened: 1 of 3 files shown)


JoJo10Smith:
Tests are still failing for python 3.7. The earliest version of geodatasets is 0.1.0 (where the example datasets are collected) requires python version 3.8. 
https://pypi.org/project/geodataset/0.1.0/

skrawcz:
> Tests are still failing for python 3.7. The earliest version of geodatasets is 0.1.0 (where the example datasets are collected) requires python version 3.8. https://pypi.org/project/geodataset/0.1.0/

yeah you can gate this with the `pytest` check I showed.

Otherwise some dataclass issue with the current tests failing? Can't see anything obvious on first glance.

skrawcz:
@JoJo10Smith BTW we removed support for 3.7 -- so no need to special case anything for that.

Issue #441, opened 2023-10-05
EOL python3.7, support 3.12


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
3.7 EOL is here, we should not support it.

**Describe the solution you'd like**
Out with 3.7, in with 3.12
- [x] kill python3.7
- [x] support python3.12

elijahbenizzy:
This is complete! Have updated all hacks we added in for 3.7 backwards compatibility. https://github.com/DAGWorks-Inc/hamilton/pull/513

Issue #284, opened 2023-08-21
[good first issue - beginner] Pandas Readers & Writers


skrawcz:
**Is your feature request related to a problem? Please describe.**
We need to add more Pandas Readers & Writers (Savers & Loaders in our internal parlance).

**Describe the solution you'd like**
We need to have readers & if appropriate, writers, covering:

- [x] #409
- [x] #292 (assigned to @benhhack)
- [x] #407 (assigned to @bengineerdavis)
- [ ] pandas fwf
- [x] #342 ()
- [x] #341 (assigned to @bryangalindo)
- [x] #369 (assigned to @JoJo10Smith)
- [x] #352 (assigned to @JoJo10Smith )
- [ ] pandas hdf
- [x] #384 (assigned to @JoJo10Smith )
- [x] #406
- [x] pandas orc (assigned to @JoJo10Smith )
- [ ] pandas sas
- [x] pandas spss
- [x] #355 (assigned to @bryangalindo )
- [x] #375 
- [x] #377 (assigned to @JoJo10Smith )
- [ ] [Latex writer](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_latex.html)

We should cover I/O as listed [here](https://pandas.pydata.org/docs/reference/io.html).

**Additional context**
We need to start building wrappers around the common ways people will want to save/load data. That way they'll have off the shelf ways to get onto Hamilton easily.

# If you're interested in contributing

If you are interested in contributing, picking up one of the above should be straightforward.

1. Ask for one, and we'll assign it.
2. We'll create an issue for you.
3. We'll then work with you on that issue.

In terms of effort, for an example of a desired class, see [this code](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/plugins/polars_extensions.py#L43-L147). It basically involves:

1. Reading the subsequent documentation.
2. Creating the right class.
3. Creating some tests for it.
4. Creating an example to put into our examples repository.

benhhack:
Hey there, I would love to give this a go with one of the pandas i/o methods. I'm new to contributing on GitHub so I appreciate the cooperative work.

elijahbenizzy:
> Hey there, I would love to give this a go with one of the pandas i/o methods. I'm new to contributing on GitHub so I appreciate the cooperative work.

@benhhack 

[...]

PR #509, opened 2023-10-31
Adding an Pandas ORC Reader and Writer class

JoJo10Smith:
This PR add the Pandas Reader and Writer Classes for the ORC file format

- Reader: https://pandas.pydata.org/docs/reference/api/pandas.read_orc.html
- Writer: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_orc.html

## Changes
`pandas_extensions.py` --> I have added the PandasORCWriter class and the PandasORCReader class
`test_pandas_extensions.py` --> I added tests for the functions
`my_script.py` --> added the materialization example
`notebook.ipynb` --> added the materialization example
`requirements-test.txt `--> added fsspec as an import since it is used for a type hint in the pandas_extensions.py file 
`test_load_from_data.orc` --> example file to read from

## How I tested this
I ran the materializations examples and the output was the same as the previous files. I also wrote tests for the new functionality 

image

## Notes
The `to_orc` was added to pandas in version 1.5.0 which requires `Python 3.8` or greater. For this reason, the tests will only for for the `PandasORCWriter` Class when the Python version is greater than or equal to 3.8. I suspect that this will not be too big of an issue since there are already plans to stop supporting `Python 3.7` (#441)

## Checklist


JoJo10Smith:
@skrawcz the `to_orc` method was introduced to pandas in version 1.5.0 (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_orc.html) 

`Pandas Version 1.5.0` requires `Python 3.8` (https://pypi.org/project/pandas/1.5.0/) so it is not available in `Python 3.7` - which is why the tests are failing for `Python 3.7`. The workaround was to only do the test if the `Python version is >= 3.8 `

The #284 mentions 'pandas prc' but I'm assuming it was meant to say ORC. 
Thanks 
Jordan

skrawcz:
> @skrawcz the `to_orc` method was introduced to pandas in version 1.5.0 (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_orc.html)
> 
> `Pandas Version 1.5.0` requires `Python 3.8` (https://pypi.org/project/pandas/1.5.0/) so it is not available in `Python 3.7` - which is why the tests are failing for `Python 3.7`. The workaround was to only do the test if the `Python version is >= 3.8 `
> 
> The #284 mentions 'pandas prc' but I'm assuming it was meant to say ORC. Thanks Jordan

Yep ORC. Yeah in pytest we can skip things for 3.7, and we can also skip registering the class unless we're 3.8+.

hamilton/plugins/pandas_extensions.py
```diff
@@ -8,6 +8,9 @@
 from pathlib import Path
 from typing import Any, Callable, Collection, Dict, Iterator, List, Optional, Tuple, Type, Union
 
+import fsspec
+import pyarrow.fs
```
skrawcz, line 12: These need to be moved after the pandas import check.

I believe pandas depends on fsspec, so it should be there. Otherwise pyarrow might not. So we should do an import check for it, and if it's not there use `Type` or something -- since it's only used as a type annotation.

skrawcz: actually it seems like fsspec might not be -- so we should check for it and use `Type` if it's not installed.

tests/plugins/test_pandas_extensions.py
```diff
@@ -204,3 +206,25 @@ def test_pandas_csv_writer(tmp_path: pathlib.Path) -> None:
     assert PandasCSVWriter.applicable_types() == [pd.DataFrame]
     assert file_path.exists()
     assert metadata["path"] == file_path
+
+
+def test_pandas_orc_writer(tmp_path: pathlib.Path) -> None:
```
skrawcz, line 211:
```suggestion
@pytest.mark.skipif(sys.version_info < (3, 8), reason="requires python3.8 or higher")
def test_pandas_orc_writer(tmp_path: pathlib.Path) -> None:
```

tests/plugins/test_pandas_extensions.py
```diff
@@ -204,3 +206,25 @@ def test_pandas_csv_writer(tmp_path: pathlib.Path) -> None:
...
+    # only avaible for python 3.8 forward
+    if sys.version_info >= (3, 8):
+        file_path = tmp_path / "test.orc"
+        writer = PandasORCWriter(path=file_path)
+        metadata = writer.save_data(pd.DataFrame(data={"col1": [1, 2], "col2": [4, 3]}))
+
+        assert PandasORCWriter.applicable_types() == [pd.DataFrame]
+        assert file_path.exists()
+        assert metadata["path"] == file_path
```
skrawcz, line 220: let's use pytest's functionality here :)

skrawcz (changes requested):
just need the import and test changes. Otherwise looks great, thanks @JoJo10Smith !

JoJo10Smith:
> just need the import and test changes. Otherwise looks great, thanks @JoJo10Smith !

No problem @skrawcz, I've made the changes. Please let me know if the import logic was what you had in mind. 
Thanks 
Jordan

skrawcz (approved):
LGTM :shipit:

PR #512, opened 2023-11-02
`Feature` Improved DAG visualization

zilto:
I tried to improve the DAG visualization using the graphviz library. The visualization should be less cluttered and visually represent the expressivity of the Hamilton DAG (inputs, config, functions, materializers, overrides, etc.). I added a legend directory in the viz for better readability.

## Changes
All the changes are made to the function `create_graphviz_graph()`. The algorithm follows the same general structure: build nodes, then build edges. Several utility functions were created to centralized relevant definitions:
- create node label (same for all nodes except input nodes)
- get the node type (input, config, function)
- get node modifiers (override, output, collect, expand, materializer)
- create a legend based on the node types of the DAG
- added arguments to the internal function `create_graphviz_graph()` these should be wired with other user-facing visualization functions

Removed:
- no longer display "Input" or "Override" prefix to reduce space used
- the base shape for functions are rectangles / rounded rectangles, which take significantly less space than ellipses.

## How I tested this
Manually tested it with several DAGs. I included DAGs that used a maximum number of features.

## Notes
- Currently doesn't implement the Parallel/Collect special edges (but both types can now be distinguished from the node style)
- Modifiers are applied sequentially and may override each other. Currently, it is not the case because each modifier acts on a different property (shape, fillcolor, periphery color, style)
- Graph, nodes, and edges have a `style` attribute which is a comma-separated list with heterogeneous attributes. These attributes are hard to manage and overriden as a whole when updating `style`

## Checklist

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -70,7 +70,9 @@ def add_dependency(
 
 
 def update_dependencies(
-    nodes: Dict[str, node.Node], adapter: base.HamiltonGraphAdapter, reset_dependencies: bool = True
+    nodes: Dict[str, node.Node],
+    adapter: base.HamiltonGraphAdapter,
+    reset_dependencies: bool = True,
 ):
     """Adds dependencies to a dictionary of nodes. If in_place is False,
     it will deepcopy the dict + nodes and return that. Otherwise it will
@@ -138,7 +140,7 @@ def create_graphviz_graph(
     graphviz_kwargs: dict,
     node_modifiers: Dict[str, Set[VisualizationNodeModifiers]],
     strictly_display_only_nodes_passed_in: bool,
-    left_to_right: bool = True,
+    orient: str = "LR",
     hide_inputs: bool = False,
     deduplicate_inputs: bool = False,
 ) -> "graphviz.Digraph":  # noqa: F821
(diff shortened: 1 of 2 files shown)


hamilton/graph.py
```diff
@@ -150,67 +151,233 @@ def create_graphviz_graph(
...
-            if is_expand_edge:
-                other_args["dir"] = "both"
-                other_args["arrowhead"] = "crow"
-                other_args["arrowtail"] = "none"
-            digraph.edge(d.name, n.name, **other_args)
+
+            node_type = _get_node_type(d)
+            if node_type == "input":
+                input_nodes.add(d)
+                continue
```
elijahbenizzy, line 351: Let's have some comments -- this is a bit of a nuanced approach (e.g. why wouldn't we have an edge style/an edge? Is this a bug?)

hamilton/graph.py
```diff
@@ -150,67 +151,233 @@ def create_graphviz_graph(
...
+        # skip input node creation
+        if hide_inputs:
+            continue
+
+        if len(input_nodes) > 0:
+            input_node_name = f"{n.name}_inputs"
+
+            # following block is for input node deduplication
+            input_nodes = frozenset(input_nodes)
+            if existing_input_name := input_sets.get(input_nodes):
```
elijahbenizzy, line 365: Walrus operator doesn't work for 3.7 -- we're going to kill support shortly, but TBD if you want to wait for that..

hamilton/graph.py
```diff
@@ -150,67 +151,233 @@ def create_graphviz_graph(
...
+
+            edge_style = _get_edge_style(d)
+            digraph.edge(d.name, n.name, **edge_style)
+
+        # skip input node creation
+        if hide_inputs:
+            continue
+
+        if len(input_nodes) > 0:
+            input_node_name = f"{n.name}_inputs"
```
elijahbenizzy, line 361: Use `__` or something that can't possibly conflict (hamilton nodes cannot have `__` generally...)

elijahbenizzy (commented):
Looking good -- a few comments:
1. Let's add some comments in the code to explain what we're doing -- not always clear
2. Walrus operator is not supported in 3.7 (maybe `__future__`?). Happy to use it but that makes this dependent on killing 3.7 which is coming soon, just waiting on pyarrow support for 3.12)
3. Unit tests seem to be failing -- we can probably update them to test this or remove the ones that were too specific in the first place.

hamilton/graph.py
```diff
@@ -150,67 +151,233 @@ def create_graphviz_graph(
...
+                style="rounded,filled",
+                fillcolor="#b4d8e4",
+                fontname=fontname,
+            )
+
+        return node_style
+
+    def _get_function_modifier_style(modifier: str):
+        if modifier == "output":
+            modifier_style = dict(fillcolor="#FFC857")
```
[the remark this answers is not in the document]
zilto: Yes, but the only type of node with a fillcolor is "function nodes" (the default).

I wonder how people design visualization software, but would it make sense to have a sort of lexicon of all the possible combinations for internal purposes?

hamilton/graph.py
```diff
@@ -150,67 +151,233 @@ def create_graphviz_graph(
...
+        legend_subgraph = graphviz.Digraph(
+            name="cluster_legend",  # needs to start with `cluster_` for graphviz layout
+            graph_attr=dict(
+                label="Legend",
+                rank="same",
+                fontname="helvetica",
+            ),
+        )
+
+        sorted_types = [
```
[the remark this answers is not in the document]
zilto: I moved materializer downwards

It's a bespoke ordering, but I thought about having `config` and `input` first because they most often appear at the top of the graph near the legend. Then, all others are of `function` type with modifiers.

hamilton/graph.py
zilto: Good catch! Didn't think of that. I will actually use `_` as prefix given no passed nodes should have that

zilto: I accordingly edited the name of the `legend_subgraph`

hamilton/graph.py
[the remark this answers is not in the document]
elijahbenizzy: 👍

hamilton/graph.py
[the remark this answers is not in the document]
elijahbenizzy: Yeah, I think I'm likely overthinking it though? As in, we can see what feedback we get?

tests/test_graph.py
```diff
@@ -501,8 +502,9 @@ def test_function_graph_has_cycles_false():
     assert fg.has_cycles(nodes, user_nodes) is False
 
 
-def test_function_graph_display():
+def test_function_graph_display(tmp_path: pathlib.Path):
```
zilto, line 505: This test was initially making two assertions:
1. is the content of the file valid
2. does it **not** create a file when passed `output_file_path = None`

I split this into two tests: `test_function_graph_display()` and `test_function_graph_display_no_dot_output()`

tests/test_graph.py
```diff
@@ -512,30 +514,100 @@ def test_function_graph_display():
...
+        all_nodes,
+        output_file_path=str(dot_file_path),
+        render_kwargs={"view": False},
+        node_modifiers=node_modifiers,
+    )
+
+    dot = dot_file_path.open("r").readlines()
+    dot_set = set(dot)
+
+    assert dot_set.issuperset(expected_set) and len(dot_set.difference(expected_set)) == 1
```
zilto, line 550: Because of the sorting issues for DOT file lines and now the content of input nodes (order of rows in the table), I used sets instead. The DOT lines `expected_set` have the input node commented out. Therefore, the newly produced DOT file `dot_set` should be a superset of `expected_set` with exactly one more line for the input node.

elijahbenizzy (approved):
LGTM! Let's 🚢 🇮🇹

PR #516, opened 2023-11-06
fixed type to string for Python <=3.9

zilto:
The recent change to `create_graphviz_graph()` in `graph.py` lead to failures for Python 3.9. More precisely, the function `_get_node_label()` was checking the `__name__` attribute of some type objects, resuling in AttributeError exceptions.

## Changes
A generic utility function `get_type_as_string` was added to `htypes.py` to get an arbitrary type as a string. The function uses different strategies to get well-formatted strings of the type otherwise return `None`.The function `create_graphviz_graph()` was changed accordingly. 

## How I tested this
A new test was added to `tests/test_type_utils.py` to check that several complex types weren't resulting in exceptions.

## Notes
- The test could check specifically for `AttributeError`
- The test covers the type to string conversion, but misshandling string and None values in `_get_node_label()` and `create_graphviz_graph()` could hypothetically lead to errors

## Checklist

### DIFF
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -2,7 +2,7 @@ import inspect
 import sys
 import typing
 from abc import ABC
-from typing import TYPE_CHECKING, Any, Generator, Tuple, Type, TypeVar
+from typing import TYPE_CHECKING, Any, Generator, Optional, Tuple, Type, TypeVar
 
 import typing_inspect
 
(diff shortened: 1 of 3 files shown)


hamilton/htypes.py
```diff
@@ -90,6 +90,25 @@ def custom_subclass_check(requested_type: Type, param_type: Type):
     return False
 
 
+def get_type_as_string(type_: Type) -> Optional[str]:
```
elijahbenizzy, line 93: Add ":param" string (even though its basic) for consistency

tests/test_type_utils.py
```diff
@@ -150,3 +150,46 @@ def test_validate_types_sad(type_):
 )
 def test__safe_subclass(candidate, type_, expected):
     assert htypes._safe_subclass(candidate, type_) == expected
+
+
+@pytest.mark.parametrize(
```
elijahbenizzy, line 155: This should assert the output if its easy...

zilto: The issue I fear is that the output depends on the Python version (might not be the case). I currently don't have the full CI/CD system locally to test on all Python versions efficiently and wanted to push a fix sooner than later

elijahbenizzy: Fair enough, let's leave it at this 👍

elijahbenizzy (approved):
🚢 🇮🇹

PR #517, opened 2023-11-06
Fix the displayed type for input node.

zilto:
The previous PR changed how types were displayed from `_get_node_label()` and `_get_input_label()`. The function `_get_input_label()` now referenced a variable that no longer existed in the function and it picked up values from the broader scope. The issue was fixed by renaming the variable within the scope of `_get_input_label()`.

## Changes
Changed the variable `n` (no longer existed within the context of the node) to `dep`. 

## How I tested this
When comparing before / after this change, I am able to reproduce the incorrect input node types and then fix the issue.

## Checklist

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -193,7 +193,7 @@ def create_graphviz_graph(
         rows = []
         for dep in input_nodes:
             name = dep.name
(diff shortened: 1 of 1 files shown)


elijahbenizzy (approved):
Test would be nice, but let's get this out then do that async so it doesn't drag on

PR #518, opened 2023-11-07
Fixes to the DAG path visualizations

zilto:
The viz would display the set of nodes + their immediate parents. The driver function would only pass the "nodes on the path" and have the viz draw the parents.

This caused an issue because nodes not explicitly passed didn't received styling. Instead, they were implicitly created when trying to create to them.

## Changes
There are 3 sets of changes:
1. Create_graphviz_graph(): added logic to highlight path in red
2. Driver visualizations: expose the new params (show_legend, orient, hide_inputs, deduplicate_inputs) to all visualization functions
3. Path visualizations (between, upstream, downstream): The changes force the parameter `strictly_display_only_passed_in_nodes=True` and handles the nodes to pass at the Driver level (i.e., includes the parents explicitly).

## How I tested this
The existing test suite still works. Changes only affect how the set of node is passed from the Driver to the `create_graphviz_graph()` function and how nodes / edges are displayed. 

## Checklist


skrawcz (approved):

hamilton/driver.py
```diff
@@ -839,14 +917,27 @@ def visualize_path_between(
...
         try:
             return self.graph.display(
-                nodes_for_path,
+                nodes_to_display,
                 output_file_path,
                 render_kwargs=render_kwargs,
                 graphviz_kwargs=graphviz_kwargs,
                 node_modifiers=node_modifiers,
-                strictly_display_only_passed_in_nodes=strict_path_visualization,
+                strictly_display_only_passed_in_nodes=True,
```
skrawcz, line 936:
```suggestion
                strictly_display_only_passed_in_nodes=strict_path_visualization,
```
this should be this right?

hamilton/driver.py
```diff
@@ -726,8 +789,12 @@ def display_upstream_of(
                 output_file_path,
                 render_kwargs=render_kwargs,
                 graphviz_kwargs=graphviz_kwargs,
-                strictly_display_only_passed_in_nodes=False,
+                strictly_display_only_passed_in_nodes=True,
```
skrawcz, line 792: why this change?

elijahbenizzy: @zilto we should add a comment. I think this is a no-op -- anything upstream will be covered by that. I think we'll want to remove that paramter entirely -- we can calculate it better outside of the function.

hamilton/driver.py
elijahbenizzy: Nope -- the only (relatively easy) way to do this was to calculate the entire set of nodes outside then only visualize those. We should remove this parameter and have the function only display the passed in ones. Otherwise you're choosing the set of included nodes in two places and it is quite confusing.

Issue #521, opened 2023-11-08
Data savers don't get registered


janhurst:
I have a custom `FeatureSaver(DataSaver)`. Prior to v1.32 I was able to engineer the registration of the saver early enough by manipulating some of my import hierarchy.

Since 1.32 I can't manage to get the saver to register. It seems very hard to not have the logic [here](https://github.com/DAGWorks-Inc/hamilton/blob/90bf57dd2ba3968358df538ac143db3c24a9290b/hamilton/io/materialization.py) executing before I can register my saver, so the attribute is not set.

This means when I try to build a materializer like `to.feature()` the `to` class has no `feature` attribute.

# Current behavior

Defining a materializer `to.<custom>()` results in an `AttributeError`

## Stack Traces
```
.venv/lib/python3.10/site-packages/hamilton/io/materialization.py:45)     ) from e

AttributeError: No data materializer named: feature. Available materializers are: dict_keys(['json', 'file', 'pickle', 'memory', 'csv', 'parquet', 'sql', 'xml', 'html', 'stata', 'feather', 'feature', 'clean_tag']). If you've gotten to this point, you either (1) spelled the loader name wrong, (2) are trying to use a loader that doesnot exist (yet). For a list of available materializers, see  https://hamilton.readthedocs.io/reference/io/available-data-adapters/#data-loaders
```

## Steps to replicate behavior
1. define a custom saver
2. register the saver
3. create a materializer `to.<customname>`

## Library & System Information
Ubuntu 20.04 based dev container docker image
Python 3.10.13
sf-hamilton 1.32+

# Expected behavior

Custom data saver can be registered and instantiated in materializer call.

elijahbenizzy:
@janhurst I think I've fixed it in the referenced PR (#523). 

For posterity/interest -- the regression in 1.32 is due to something clever  -- it dynamically assigned properties on import to the class, and set the signatures to be a certain value. This way, you could use a code-completing env (such as jupyter notebook) and get it to display the set of available materializers! That said, it turned the dynamic piece to static for these, which made custom ones harder.

In my personal development, I was sticking the import of the registering class prior to `to`, but that's ridiculous -- the order shouldn't matter. The PR ensures the order does not matter, automatically refreshing the available set if we detect that it should be present.

janhurst:
looking good on this side with 1.37.2rc0

skrawcz:
Closing because we fixed this.

PR #523, opened 2023-11-09
Allows for dynamic registration of materializer types

elijahbenizzy:
Before we were just registering once. Now we synchrionize all konwn types if we detect a missing type -- a type that is in the registry but not in the associated class. Note that we could actually just go to the saver registry directly, but we like to add to it dynamically so we can refer to it, say, in a jupyter notebook. If we don't add it dynamically, we'll never know the available set.

## Checklist

### DIFF
--- a/docs/reference/io/available-data-adapters.rst
+++ b/docs/reference/io/available-data-adapters.rst
@@ -43,6 +43,10 @@ If you want to extend these, see :doc:`/reference/io/available-data-adapters` fo
 and `the example <https://github.com/DAGWorks-Inc/hamilton/blob/main/examples/materialization/README.md>`_
 in the repository for an example of how to do so.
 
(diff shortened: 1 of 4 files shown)


elijahbenizzy:
Solves #521

docs/reference/io/available-data-adapters.rst
```diff
@@ -43,6 +43,10 @@ If you want to extend these, see :doc:`/reference/io/available-data-adapters` fo
 and `the example <https://github.com/DAGWorks-Inc/hamilton/blob/main/examples/materialization/README.md>`_
 in the repository for an example of how to do so.
 
+Note that you will need to call `registry.register_adapters` (or import a moule that does that)
```
skrawcz, line 46:
```suggestion
Note that you will need to call `registry.register_adapters` (or import a module that does that).
```

skrawcz (approved):

PR #526, opened 2023-11-09
added Matplotlib data saver

zilto:
Added a DataSaver to save matplotlib objects of type `matplotlib.figure.Figure`, i.e., resulting from `plt.figure()` and `plt.subplots()`.

## Changes
Added a `matplotlib_extensions.py` with the DataSaver class and added it to the registry

## How I tested this
Added a test to check the DataSaver properly writes to a file. I manually validated that the output file was correct.

## Notes
- The Matplotlib and the Plotly DataSaver accept as `path` argument objects of type `IO` (i.e., in-memory files, streams, etc.) which may fail when `hamilton.io.utils.get_file_metadata()` is called.

## Checklist

### DIFF
--- a/hamilton/function_modifiers/base.py
+++ b/hamilton/function_modifiers/base.py
@@ -21,6 +21,7 @@ if not registry.INITIALIZED:
     # right now. Side note: ray serializes things weirdly, so we need to do this here rather than in
     # in the other choice of hamilton/base.py.
     plugins_modules = [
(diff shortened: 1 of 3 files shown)


skrawcz:
> E   AttributeError: module 'matplotlib' has no attribute 'artist'

?

skrawcz (approved):

PR #536, opened 2023-11-14
Supports "Any" in loader types

elijahbenizzy:
There are certain cases where the loader can load any datatype , E.G. pickle-type loaders. Before this didn't work. The "proper" solution would be to change custom_subclass_check, but that has the possibility of breaking some cases of materialization as we have not always been consistent here. In this case, it *just* uses a loader with `Any` if another has not been specified. While this might casue the order to be slightly counter-intuitive, it will ensure full backwrads compatibility.

## Checklist

### DIFF
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -509,6 +509,10 @@ class SimplePythonGraphAdapter(SimplePythonDataFrameGraphAdapter):
 
     This executes the Hamilton dataflow locally on a machine in a single threaded, single process fashion. It allows\
     you to specify a ResultBuilder to control the return type of what ``execute()`` returns.
+
+    Currently this extends SimplePythonDataFrameGraphAdapter, although that's largely for legacy reasons (and can probably be changed).
+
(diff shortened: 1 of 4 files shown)


hamilton/base.py
```diff
@@ -509,6 +509,10 @@ class SimplePythonGraphAdapter(SimplePythonDataFrameGraphAdapter):
 
     This executes the Hamilton dataflow locally on a machine in a single threaded, single process fashion. It allows\
     you to specify a ResultBuilder to control the return type of what ``execute()`` returns.
+
+    Currently this extends SimpePythonDataFrameGraphAdapter, althoguh that's largely for legacy reasons (and can probaly be changed).
```
skrawcz, line 513:
```suggestion
    Currently this extends SimpePythonDataFrameGraphAdapter, although that's largely for legacy reasons (and can probably be changed).
```

skrawcz (approved):

PR #539, opened 2023-11-15
`example` added example for contrib dataflows

zilto:
Added a `examples/contrib` folder with a script that uses static installation and a notebook for dynamic installation and local copy.

## Notes
- We probably want to add links at the end of the README for the contribution instructions and PR links


examples/contrib/notebook.ipynb
```diff
@@ -0,0 +1,1293 @@
...
+   ],
+   "source": [
+    "dr = (\n",
+    "    driver.Builder()\n",
+    "    .with_modules(xgboost_optuna)\n",
+    "    .with_config(dict(task=\"classification\"))\n",
+    "    .build()\n",
+    ")\n",
+    "\n",
+    "dr.display_all_functions(None)"
```
skrawcz, line 444: you can drop None.

examples/contrib/notebook.ipynb
```diff
@@ -0,0 +1,1293 @@
...
```
skrawcz, line 1268: can drop `None`.

examples/contrib/run.py
```diff
@@ -0,0 +1,62 @@
+if __name__ == "__main__":
```
skrawcz, line 1: probably don't need to gate this if it's all under this if statement?

skrawcz (approved):
Two minor things. Otherwise looks good. We should add a link to the hub introduction README too since this is good documentation. 

Oh can the notebook link to the script too. e.g. have `1. Static install` and just link to the script?

skrawcz (approved):

PR #541, opened 2023-11-16
Update README.md

elijahbenizzy:
Missed updates from the upgrade


skrawcz (approved):
I think in the general docs pager there's also 3.7+ listed...

skrawcz (approved):

PR #543, opened 2023-11-18
Changes type annotation on DB connection to Any

skrawcz:
This was unnecessarily restrictive. The original
pandas doesn't have a type here. So need to put Any for our type system to work.

## Changes
 - changes type annotation
## How I tested this
 - locally

## Checklist


elijahbenizzy (approved):
LGTM

Issue #510, opened 2023-10-31
create_temporary_modules should error out with helper (`_`) functions


elijahbenizzy:
If we create_temporary_modules with an underscore function it should not work. It will not get turned into a function, and the function can refer to a helper function regardless.

This can help the ad-hoc case.

skrawcz:
I don't understand? You pass in the functions with create_temporary_modules? Why not just prevent people passing in `_` ones?

elijahbenizzy:
> I don't understand? You pass in the functions with create_temporary_modules? Why not just prevent people passing in `_` ones?

Yep, that's what this issue suggests. It curerntly does work, and gets weird results, but it shouldn't.

elijahbenizzy:
This is complete! https://github.com/DAGWorks-Inc/hamilton/pull/558

PR #558, opened 2023-11-27
Errors out when calling create_temporary_module with functions that

elijahbenizzy:
start with "_"

See https://github.com/DAGWorks-Inc/hamilton/issues/510

This is *not* backwards compatible, but this is a specific "ad-hoc" util and has an easy fix, so I'm OK not bumping versions accordingly. I would consider this previously undefined behavior, and this will make debugging easier.


elijahbenizzy:
See https://github.com/DAGWorks-Inc/hamilton/issues/510

skrawcz (approved):

PR #587, opened 2023-12-12
Adds examples showing printing of inputs with node error

skrawcz:
So this will:

1. print the name, and the original function name
2. print the node inputs

E.g.

********************************************************************************
> spend_std_dev [my_functions.spend_std_dev()] encountered an error            <
> Node inputs:
{'spend': 0    10
1    10
2    20
3    40
4    40
5    50
dtype: int64}
********************************************************************************

This will help make it easier to debug a function failure.

I chose pprint, since it has a mode to limit the width on a best effort basis to keep things looking reasonably tidy.

## Changes
 - graph_functions
 - @ config decorator
 - configuration 
 - function_modifiers.base

## How I tested this
 - locally

## Checklist

### DIFF
--- a/hamilton/execution/graph_functions.py
+++ b/hamilton/execution/graph_functions.py
@@ -146,8 +146,20 @@ def execute_subdag(
             try:
                 value = adapter.execute_node(node_, kwargs)
             except Exception:
-                message = f"> Node {node_.name} encountered an error <"
-                border = "*" * len(message)
+                if hasattr(node_.originating_functions[0], "__original_name__"):
+                    original_func_name = node_.originating_functions[0].__original_name__
+                else:
+                    original_func_name = node_.originating_functions[0].__name__
+                module = node_.originating_functions[0].__module__
+                message = f"> {node_.name} [{module}.{original_func_name}()] encountered an error"
+                padding = " " * (80 - len(message) - 1)
+                message += padding + "<"
+                import pprint
+
+                pp = pprint.PrettyPrinter(width=80)
+                input_string = pp.pformat(kwargs)
+                message += "\n> Node inputs:\n" + input_string
+                border = "*" * 80
                 logger.exception("\n" + border + "\n" + message + "\n" + border)
                 raise
         computed[node_.name] = value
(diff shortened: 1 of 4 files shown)


hamilton/execution/graph_functions.py
```diff
@@ -146,8 +146,30 @@ def dfs_traverse(
             try:
                 value = adapter.execute_node(node_, kwargs)
             except Exception:
-                message = f"> Node {node_.name} encountered an error <"
-                border = "*" * len(message)
+                # This code is coupled to how @config resolution works. Ideally it shouldn't be,
```
elijahbenizzy, line 149: Open an issue?

hamilton/function_modifiers/configuration.py
```diff
@@ -175,6 +175,10 @@ def _get_function_name(self, fn: Callable) -> str:
     def resolve(self, fn, config: Dict[str, Any]) -> Callable:
```
elijahbenizzy, line 175: Hmm, why do we have this? Sticking it in tags feels strange. Also, there are cases in which this will have *every* config item (E.G. with `@config.when(lambda x....)` -- the function approach)

hamilton/execution/graph_functions.py
skrawcz: https://github.com/DAGWorks-Inc/hamilton/issues/589

hamilton/function_modifiers/configuration.py
```diff
@@ -175,6 +175,9 @@ def _get_function_name(self, fn: Callable) -> str:
...
         if not self.does_resolve(config):
             return None
+        # attaches config keys used to resolve function
+        fn.__config_decorated__ = self._config_used if self._config_used else ["__UNKNOWN__"]
```
elijahbenizzy, line 179: `is not None` -- I don't think this will be empty but just in case

skrawcz: since we don't see a case for `[]` being passed, I think truthiness is fine here -- it's more concise and covers both.

PR #592, opened 2023-12-17
Lifecycle customization

elijahbenizzy:
Lifecycle Customizations, see #605 

## Checklist

- [ ] What remains?
  - [x] Implementation
  - [x] Documentation
  - [x] Unit tests/E2E tests
  - [x] Add a few user-facing ones, using the user-facing API
    - [x] println adapter -- prints information before/after
    - [x] debug adapter -- pause with pdb before/after
    - [ ] DDOG adapter -- logs to datadog 
    - [ ] OpenTel adapter -- logs to openTel


elijahbenizzy:
This is a WIP, not ready for feedback or comments

elijahbenizzy:
TODO prior to release:
- [ ] Sort out commits -- smaller pieces
- [ ] Ensure sphinx docs is good
- [ ] Add a few pre-canned examples, first-class
    - [ ] `PrintlnHook`
    - [ ] `PermissiveTypeCheckerHook`
    - [ ] ???

docs/concepts/customizing-execution.rst
```diff
@@ -35,57 +45,47 @@ something.
...
 Graph Adapters `adapt` the Hamilton DAG, and change how it is executed. They all implement a single interface called
 ``base.HamiltonGraphAdapter``. They are called internally by Hamilton at the right points in time to make execution
-work. The link with the Result Builders, is that GraphAdapters need to implement a ``build_result()`` function
+work. The link with the Result Builders, is that GraphAdapters (currently) need to implement a ``build_result()`` function
 themselves.
 
 .. code-block:: python
 
-    class HamiltonGraphAdapter(ResultMixin):
+    class HamiltonGraphAdapter(customization.ResultBuilder):
```
skrawcz, line 61: I think `lifecycle` is a better name than `customization`?

docs/concepts/customizing-execution.rst
```diff
@@ -35,57 +45,47 @@ something.
...
         # four functions not shown
 
-The default GraphAdapter is the ``base.SimplePythonDataFrameGraphAdapter`` which by default makes Hamilton try to build
-a ``pandas.DataFrame`` when ``.execute()`` is called.
-
-If you want to tell Hamilton to return something else, we suggest starting with the ``base.SimplePythonGraphAdapter``
-and writing a simple class & function that implements the ``base.ResultMixin`` interface and passing that in.  See
-:doc:`../reference/graph-adapters/index` and
-:doc:`../reference/result-builders/index` for options.
+The default performance is to join all data in a dictionary (unless you bypass `driver.Builder`), in which case it will
```
skrawcz, line 68:
```suggestion
The default behavior is to join all data in a dictionary (unless you bypass `driver.Builder`), in which case it will
```

docs/concepts/customizing-execution.rst
```diff
@@ -35,57 +45,47 @@ something.
...
+return a pandas dataframe. Should you find aspects you wish to customize, reach out to the maintainers. The odds are high
+that we already have the right abstraction and have not yet exposed it to the user.
```
skrawcz, line 70:
```suggestion
that we already have the right abstraction and have not yet exposed it via documentation & examples.
```

docs/reference/customizing-execution/index.rst
```diff
@@ -0,0 +1,29 @@
+==============
+GraphAdapters
+==============
+
+This section helps determine ways to execute Hamilton.
+Note that these are special cases of the Lifecycle Customizations described in the next section,
+meant to help with execution. They implement multiple lifecycle customizations at once.
```
skrawcz, line 7:
```suggestion
meant to help with execution. They implement multiple lifecycle customizations in a single place.
```

hamilton/customization/api.py
```diff
@@ -0,0 +1,306 @@
...
+        """This is run to check if edge types match. Note that this is an OR functionality
+        -- this is run after we do some default checks, so this can only be permissive.
+        Reach out if you want to be more restrictive than the default checks.
+
+        :param type_from: The type of the node that is the source of the edge.
+        :param type_to: The type of the node that is the destination of the edge.
+        :param kwargs: This is kept for future backwards compatibility.
+        :return: Whether or not the two node types form a valid edge.
+        """
+        return self.do_check_edge_types_match(type_from=type_from, type_to=type_to)
```
skrawcz, line 246: don't these two functions refer to each other? I'm confused.

hamilton/customization/api.py
```diff
@@ -0,0 +1,306 @@
...
+        """This is run to check if the input is valid for the node type. Note that this is an OR functionality
+        -- this is run after we do some default checks, so this can only be permissive.
+        Reach out if you want to be more restrictive than the default checks.
+
+        :param node_type: Type of the node that is accepting the input.
+        :param input_value: Value of the input
+        :param kwargs: Keyword arguments -- this is kept for future backwards compatibility.
+        :return: Whether or not the input is valid for the node type.
+        """
+        return self.do_validate_input(node_type=node_type, input_value=input_value)
```
skrawcz, line 263: same with these two?

hamilton/customization/api.py
```diff
@@ -0,0 +1,306 @@
...
+
+
+class NodeExecutionMethod(BaseDoNodeExecute):
+    """API for executing a node. This takes in tags, callable, node name, and kwargs, and is
+    responsible for executing the node and returning the result. Note this is not (currently)"""
```
skrawcz, line 268: unfinished comment.

hamilton/customization/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+
+# All the public-facing types go here
+__all__ = [
+    "LifecycleAdapter",
+    "LegacyResultMixin",
+    "ResultBuilder",
+    "GraphAdapter",
+    "NodeExecutionHook",
+    "EdgeConnectionHook",
+    "PrintLnHook",
```
skrawcz, line 19: is this missing the PDB one?

hamilton/customization/default.py
```diff
@@ -0,0 +1,303 @@
...
+        self,
+        node_filter: NodeFilter,
+        before: bool = False,
+        during: bool = False,
+        after: bool = False,
+    ):
+        """Creates a PDB debugger. This has three possible modes:
+        1. Before -- places you in a function with (a) node information, and (b) inputs
+        2. During -- runs the node with pdb.run. Note this may not always work or give what you expect as
+            node functions are often wrapped in multiple levels of input modifications/whatnot. That said, it should give you something.
```
skrawcz, line 294: To get around the partial/wrapped functions we could want to have a "hook" that would wrap functions at graph build time?

hamilton/driver.py
```diff
@@ -472,8 +546,11 @@ def raw_execute(
         overrides: Dict[str, Any] = None,
         display_graph: bool = False,
         inputs: Dict[str, Any] = None,
+        _fn_graph: graph.FunctionGraph = None,
     ) -> Dict[str, Any]:
         """Raw execute function that does the meat of execute.
+        This is being deprecated in favor of using Base.DictResult(), which will be the default, and already
```
skrawcz, line 552: what is "this" refer to here? Because it's not this function that's being deprecated right?

hamilton/customization/default.py
```diff
@@ -0,0 +1,303 @@
...
+
+        :param node_name: Name of the node
+        :param node_tags: Tags of the node
+        :param node_callable: Callable function of the node
+        :param node_kwargs: Keyword arguments passed to the node
+        :param task_id: ID of the task that the node is in, if any
+        :param future_kwargs: Additional keyword arguments that may be passed to the hook yet are ignored for now
+        :return: Result of the node
+        """
+        if not should_run_node(node_name, node_tags, self.node_filter):
```
skrawcz, line 175: there's also a few other things to consider:

1. you can skip stacks from some modules when you pass that as "skip" https://docs.python.org/3/library/pdb.html#pdb.Pdb. could be useful to skip most of hamilton stuff?
2. the post mortem option -- so could try/except around the function and then only debug on error.

docs/concepts/customizing-execution.rst
elijahbenizzy: Lifecycle is nebulous -- customization is what we're doing?

hamilton/customization/default.py
elijahbenizzy: Yeah skipping stacks is a more powerful way to do this I think? Its a little bit iffy though, so I'm happy to add over time. And yeah, saw the `.pm` option, which should be good for error. We can also add another param `error=True/False`. 

Going to make a comment that the API will likely remain backwards compatible, but the impact of it might not.

hamilton/customization/api.py
elijahbenizzy: Yeah this was a bug -- hadn't been using this class

hamilton/driver.py
elijahbenizzy: Yes, we don't want people to use raw_execute anymore, just the driver builder with `DictResult`

hamilton/driver.py
```diff
@@ -472,8 +546,11 @@ def raw_execute(
         overrides: Dict[str, Any] = None,
         display_graph: bool = False,
         inputs: Dict[str, Any] = None,
+        _fn_graph: graph.FunctionGraph = None,
     ) -> Dict[str, Any]:
-        """Raw execute function that does the meat of execute.
+        """Raw execute function that does the meat of execute. This entrypoint for running the graph
+        will be deprecated in favor of using Base.DictResult(), which will be the default,
+        and already forms the default in the new builder pattern.
```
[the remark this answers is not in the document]
elijahbenizzy: Calling `.raw_execute()`. Rather, you call:

```python
dr = driver.Builder()...
dr.execute(...)
```
Or explicitly pass in the `base.DictResult()` result builder

docs/concepts/customizing-execution.rst
elijahbenizzy: Changed to `lifecycle` (we discussed offline) -- more consistent and descriptive enough

hamilton/customization/default.py
elijahbenizzy: Yes, maybe, but its also a little complicated. E.G. how does that interact with decorators? 
The PDB debugger added the caveat that its not complete -- the API will be stable, but the implementation might change.

Some other thoughts:
1. Decorators are actually lifecycle methods -- this could fit there
2. Not every function is actually executed (although `@does` might be the only weird case..)
3. With decorators, its usually one node that actually executes the fn, although sometimes they don't

So yeah, probably a cleaner way to do this, but want to push that off till later

hamilton/driver.py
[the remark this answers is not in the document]
skrawcz: Right -- so the comment should direct users to what they should be doing. i.e. 
```suggestion
        """Raw execute function that does the meat of execute. 
        
        Don't use this entry point for execution directly. Always go through `.execute()`.
        In case you are using `.raw_execute()` directly, please switch to `.execute()` using a
        `base.DictResult()`. Note: `base.DictResult()` is the default return of execute if you are
        using the `driver.Builder()` class to create a `Driver()` object.
```
Otherwise I'm not worried, we didn't publicize using `raw_execute` much.

hamilton/customization/default.py
skrawcz: yes, not for this PR -- but for the future.

hamilton/lifecycle/api.py
```diff
@@ -0,0 +1,307 @@
...
+        :param input_value: Value of the input
+        :param kwargs: Keyword arguments -- this is kept for future backwards compatibility.
+        :return: Whether or not the input is valid for the node type.
+        """
+        pass
+
+
+class NodeExecutionMethod(BaseDoNodeExecute):
+    """API for executing a node. This takes in tags, callable, node name, and kwargs, and is
+    responsible for executing the node and returning the result. Note this is not (currently)"""
```
skrawcz, line 268: still unfinished comment.

hamilton/execution/executors.py
```diff
@@ -100,12 +100,49 @@ def base_execute_task(task: TaskImplementation) -> Dict[str, Any]:
...
+            task_id=task.task_id,
+            nodes=task.nodes,
+            inputs=task.dynamic_inputs,
+            overrides=task.overrides,
+        )
+    try:
+        out = execute_subdag(
+            nodes=task.nodes,
+            inputs=task.dynamic_inputs,
+            adapter=task.adapter,  # TODO -- wire through multiple graph adapters
```
skrawcz, line 116: this TODO still relevant?

hamilton/execution/graph_functions.py
```diff
@@ -144,11 +152,52 @@ def dfs_traverse(
                 if dependency.name in computed:
                     kwargs[dependency.name] = computed[dependency.name]
             try:
-                value = adapter.execute_node(node_, kwargs)
-            except Exception:
+                # TODO -- determine whether I want to check if they exist or not...
```
skrawcz, line 155: relevant?

hamilton/execution/graph_functions.py
```diff
@@ -144,11 +152,52 @@ def dfs_traverse(
...
+                if adapter.does_hook("pre_node_execute", is_async=False):
+                    adapter.call_all_lifecycle_hooks_sync(
+                        "pre_node_execute",
+                        run_id=run_id,
+                        node_=node_,
+                        kwargs=kwargs,
+                        task_id=task_id,
+                    )
+                # TODO -- validate
```
skrawcz, line 164: relevant? or add more details.

tests/lifecycle/test_lifecycle_adapters_end_to_end_task_based.py
```diff
@@ -0,0 +1,301 @@
...
+        "n_iters": 1,
+        "processed": 5,
+        "more_processed": 5,
+        "collect": 1,
+        "output": 1,
+    }
+    run_ids = {item.bound_kwargs["run_id"] for item in relevant_calls}
+    (run_id,) = run_ids
+    assert len(run_ids) == 1
+    assert len(run_id) > len("TODO -- add a run ID")  # I'd hope it has more entropy than this...
```
skrawcz, line 88: what are you testing for here?

hamilton/driver.py
```diff
@@ -472,38 +546,79 @@ def raw_execute(
...
+                    "post_graph_execute",
+                    run_id=run_id,
+                    graph=function_graph,
+                    success=False,
+                    error=e,
+                    results=None,
+                )
+            raise e
+
+        # TODO -- add error case
```
skrawcz, line 620: what  error case needs to be caught here?

hamilton/execution/executors.py
elijahbenizzy: Nope

tests/lifecycle/test_lifecycle_adapters_end_to_end_task_based.py
elijahbenizzy: Its a UUID, it should have more chars than the p;laceholder I had before :) i'll change

hamilton/driver.py
elijahbenizzy: This is done

skrawcz (approved):
I haven't gone over this with a fine toothed comb, but I have run the branch locally and it seems to not introduce any regressions.

elijahbenizzy:
> I haven't gone over this with a fine toothed comb, but I have run the branch locally and it seems to not introduce any regressions.

I'm doing a detailed pass to form another review

PR #593, opened 2023-12-18
Do not fail default NaN validation on empty series

elshize:
If the validated series is empty, then it makes sense to treat it as if 0% of the values are NaN.

## Changes

The only change is to set the nan fraction to 0 if the series is empty.

## How I tested this

Unit tests are added for the case of empty series.

## Checklist

### DIFF
--- a/hamilton/data_quality/default_validators.py
+++ b/hamilton/data_quality/default_validators.py
@@ -198,7 +198,7 @@ class MaxFractionNansValidatorPandasSeries(base.BaseDefaultValidator):
     def validate(self, data: pd.Series) -> base.ValidationResult:
         total_length = len(data)
         total_na = data.isna().sum()
-        fraction_na = total_na / total_length
+        fraction_na = total_na / total_length if total_length > 0 else 0
         passes = fraction_na <= self.max_fraction_nans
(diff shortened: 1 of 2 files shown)


elshize:
Not sure if there's any documentation that goes so deep into details, I couldn't find any so I did not update anything.

I think this behavior is the intuitive one, but let me know if you disagree.

skrawcz:
@elshize you specifically mean this fails:

```python

@check_output(max_fraction_nans=0.2)
def empty() -> pd.Series:
   return pd.Series([])

```
?

elshize:
Yes, though of course I'm thinking when e.g. the loaded data turns out to be empty or there's a filtering step.

skrawcz:
@elshize easy - https://github.com/DAGWorks-Inc/hamilton/pull/594

elshize:
@skrawcz That works :)

skrawcz (commented):
I like this better -- sorry, mind reopening? I'll merge this one instead of mine.

elshize:
Ah, sure, no problem.

skrawcz (approved):

skrawcz:
Published in `1.40.1`

PR #596, opened 2023-12-19
Extends Pandera validator to handle more DF types

skrawcz:
We were assuming only pandas annotated functions. This changes that and ensure that functions annotated with a dask datatype will work.

Note, added pyspark without adding a test.

## Changes
 - pandera validator

## How I tested this
 - locally via unit tests

## Notes
 - this uses the registered dataframe and column types from registry filtering on an explicit list that pandera supports.

## Checklist


elijahbenizzy (approved):
LGTM

PR #597, opened 2023-12-20
Exposes ray init functionality

skrawcz:
This was bothering me that we hadn't exposed a few things, this fixes that. It means that someone can now connect to a ray cluster for example.

## Changes
 - h_ray.py task executor

## How I tested this
 - locally

## Checklist

### DIFF
--- a/hamilton/plugins/h_ray.py
+++ b/hamilton/plugins/h_ray.py
@@ -209,20 +209,33 @@ class RayTaskExecutor(executors.TaskExecutor):
     This is still experimental, so the API might change.
     """
 
-    def __init__(self, num_cpus: int):
+    def __init__(
+        self,
+        num_cpus: int = None,
+        config: typing.Dict[str, typing.Any] = None,
+        skip_init: bool = False,
+    ):
         """Creates a ray task executor. Note this will likely take in more parameters. This is
         experimental, so the API will likely change, although we will do our best to make it
         backwards compatible.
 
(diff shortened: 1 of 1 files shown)


hamilton/plugins/h_ray.py
```diff
@@ -209,20 +209,33 @@ class RayTaskExecutor(executors.TaskExecutor):
     This is still experimental, so the API might change.
     """
 
-    def __init__(self, num_cpus: int):
+    def __init__(
+        self,
+        num_cpus: int = None,
```
elijahbenizzy, line 214: We expose num_cpus -- anything else we want to expose? Or do we just want to keep that there for backwards compatibility?

hamilton/plugins/h_ray.py
```diff
@@ -209,20 +209,33 @@ class RayTaskExecutor(executors.TaskExecutor):
...
+        config: typing.Dict[str, typing.Any] = None,
```
elijahbenizzy, line 215: `ray_init_config` might be more descriptive

elijahbenizzy (commented):
A few nits but let's ship.

hamilton/plugins/h_ray.py
skrawcz: backwards compatibility.

hamilton/plugins/h_ray.py
skrawcz: done

elijahbenizzy (approved):

PR #601, opened 2023-12-21
Add missing import to fix test

roelbertens:
When running the tests locally I needed to add one import to get all tests running.


skrawcz (approved):
yes thanks -- it seems that because we run unit tests from root it just happens that it's imported and why it hasn't broken earlier.

PR #607, opened 2023-12-24
Schema management for dataset-types

elijahbenizzy:
## Checklist


elijahbenizzy:
POC taking @roelbertens's implementation and making a little more general. 

## Design decisions to enable this:
1. This is going to take the form of a decorator
2. The data, for now, is not going to be accessible
3. We store the data under tags in the node as a serialized format. Under `hamilton.internal.schema=...`

## Further questions:
1. How do we allow for API change?
    - Keep with this API
    - Have the parameter not be `**kwargs`  -- be able to switch it later
    - Change this to `dataset_schema` or something like that, use `schema` later
    - Keep in experimental/call it `schema_beta`
2. Should schemas allow order?
3. Should we be able to build schemas for specific libraries (spark, pydantic, etc...)?
    - Probably yes -- we should have a schema class
    - This schema class will have the ability to specify a "flavor", which will be generic
    - It will allow for column names/types
    - We should be able to convert between them/check compatibility with the right plugin
4. How to decorate input types?
    - Type-annotation

## TODO remaining:
- [ ] Add some basic unit tests for schema
- [ ] Add some documentation
- [ ] Add/fix some tests for viz
- [ ] Ensure that the node modifiers work together (E.G. we have some colors that we need to use with the different node modifiers)

examples/spark/pyspark_feature_catalog/features.py
```diff
@@ -2,28 +2,28 @@
 from pyspark.sql import functions as sf
 from with_columns import darkshore_flag, durotar_flag
 
-from hamilton.function_modifiers import tag
+from hamilton.function_modifiers.metadata import schema
 from hamilton.plugins.h_spark import with_columns
 
-WORLD_OF_WARCRAFT__SCHEMA = "zone:str, level:int, avatarId:int"
+WORLD_OF_WARCRAFT_SCHEMA = {"zone": str, "level": int, "avatarId": int}
```
skrawcz, line 8: Why the need to constrain to python types? 

E.g. someone should be able to specify things like:
```python
"zone:str, level:int64, avatarId:int32"
```

I mean, this should be one of the possible ways to specify it, but I think as the only way, it's too restrictive.

hamilton/function_modifiers/metadata.py
```diff
@@ -166,3 +167,49 @@ def decorate_node(self, node_: node.Node) -> node.Node:
...
+        return type_.__name__
+    elif type_ == datetime.datetime:
+        return "datetime"
+    elif type_ == datetime.date:
+        return "date"
+    raise ValueError(f"Type {type_} is not supported")
+
+
+class schema(tag):
+    """Represents a schema for a node. This is implemented using tags, but that might change."""
```
skrawcz, line 201:
```suggestion
    """Represents an output schema for a node. This is implemented using tags, but that might change."""
```

hamilton/function_modifiers/metadata.py
```diff
@@ -166,3 +167,49 @@ def decorate_node(self, node_: node.Node) -> node.Node:
...
+
+    INTERNAL_SCHEMA_KEY = "hamilton.internal.schema"
```
skrawcz, line 203: Probably want to call this out explicitly. 
```suggestion
    INTERNAL_SCHEMA_KEY = "hamilton.internal.output_schema"
```

examples/spark/pyspark_feature_catalog/features.py
elijahbenizzy: Sure, we can always add more. And `int32` versus `int64` are generally specific to certain frameworks (we wouldn't, likely, have this in, say, pydantic). Hence `int` as the common one. This was also the goal to have the different flavors, which map to each other in certain ways.

hamilton/function_modifiers/metadata.py
```diff
@@ -166,3 +167,94 @@ def decorate_node(self, node_: node.Node) -> node.Node:
...
+    "int",
+    "float",
+    "str",
+    "bool",
+    "dict",
+    "list",
+    "object",
+    "datetime",
+    "date",
+]
```
skrawcz, line 182: Some extras:

```suggestion
SchemaTypes = Literal[
    "int",             # Standard integer
    "float",           # Floating point number
    "str",             # String
    "bool",            # Boolean
    "dict",            # Dictionary
    "list",            # List
    "list[int]", "list[float]", "list[str]", "list[bool]", "list[dict]", "list[datetime]", "list[date]", "list[complex]", "list[list]", 
    "object",          # Generic object
    "datetime",        # Datetime object
    "date",            # Date object
    "complex",         # Complex number
    "ndarray",         # Numpy array
    "timedelta",       # Time delta object
    "bytes",           # Bytes
    "Decimal",         # Decimal numbers for high precision
    "Fraction",        # Fraction type
    'int8', 'int16', 'int32', 'int64',
        'uint8', 'uint16', 'uint32', 'uint64',
        'float16', 'float32', 'float64', 'float128',
        'complex64', 'complex128', 'complex256',
        'timedelta64', 'datetime64', 'timestamp',
]
```
otherwise how do we want to deal with nested structures properly? 

I'm almost inclined to say don't do any validation, but allow people to register their own, or by default register our one here that they can override and/ or extend.

skrawcz: nvm -- I see this isn't validated against...

examples/spark/pyspark_feature_catalog/features.py
```diff
@@ -0,0 +1,39 @@
+import pyspark.sql as ps
+from pyspark.sql import functions as sf
+from with_columns import darkshore_flag, durotar_flag
+
+from hamilton.function_modifiers.metadata import schema
```
skrawcz, line 5: shouldn't this be imported from  `hamilton.function_modifiers` ?

examples/spark/pyspark_feature_catalog/features.py
```diff
@@ -0,0 +1,39 @@
...
+from hamilton.plugins.h_spark import with_columns
+
+WORLD_OF_WARCRAFT_SCHEMA = (("zone", "str"), ("level", "int"), ("avatarId", "int"))
+
+
+def spark_session() -> ps.SparkSession:
+    return ps.SparkSession.builder.master("local[1]").getOrCreate()
+
+
+@schema(*WORLD_OF_WARCRAFT_SCHEMA)
```
skrawcz, line 15: So this would become `@schema.output(...)` right?

hamilton/function_modifiers/metadata.py
```diff
@@ -166,3 +166,91 @@ def decorate_node(self, node_: node.Node) -> node.Node:
...
+    "int",
+    "float",
+    "str",
+    "bool",
+    "dict",
+    "list",
+    "object",
+    "datetime",
+    "date",
+)
```
skrawcz, line 185: where is this used?

hamilton/function_modifiers/metadata.py
```diff
@@ -166,3 +166,91 @@ def decorate_node(self, node_: node.Node) -> node.Node:
...
+        This is implemented using tags, but that might change. Thus you should not
+        rely on the tags created by this decorator (which is why they are prefixed with `internal`).
+
+        To use this, you should decorate a node with `@schema.output`
+
+        Example usage:
+
+        .. code-block:: python
+
+           @schema.output(a=int, b=float, c=str)
```
skrawcz, line 250:
```suggestion
           @schema.output(("a", "int"), ("b", "float"), ("c", "str"))
```

hamilton/function_modifiers/metadata.py
elijahbenizzy: Its not -- I'll comment it out but leave it in for now. See comment above.

skrawcz (approved):

PR #618, opened 2024-01-02
Moves automatic registration from h_spark to spark_extensions

elijahbenizzy:
It was in the wrong place -- in the past this was in h_spark, which doesn't get automatically registered. Thus this had to be moved to spark_extensions, which *does* get automatically registered.

The trickier part is that we have to do some odd things because the system assumes that all extensions have dataframe types and column types, or neither. While we'll likely want to change the assumptions we make, this should work for now.

## Checklist


elijahbenizzy:
TODO:
- [ ] Add E2E test to ensure it works with proper imports
- [ ] Fix polars issue

skrawcz (approved):

Issue #620, opened 2024-01-06
Allow non-strings in @tag(...)


kzk2000:
```@tag(business_line=<must be a string here>)``` doesn't currently allow for non-strings.
I'd would be great if, say, a list of strings would also work, so I could do something this ```@tag(business_line=["US", "CA"])``` 

and then leverage that later with a wrapper like this ```dr.list_available_variables(tag_filter={"business_line" : "US"})``` or similar

**Describe alternatives you've considered**
Write my own ```@custom_tag()``` derived from ```@tag``` class that disables the "str" type check. Doable but preferably I don't go off tangent with custom code and Hamilton just supports this out of the box.

**Comment by elijahbenizzy** [comment]
Agreed that we should allow for some structure here. I think we have two choices:
1. Allow any arbitrary json here
2. Allow some specific structure that has semantic meaning.

I'm going to recommend (2). In this case:
- strings assign a value to the tag
- lists assign multiple values to the tag

Then we have a query system (suggested in [here](https://github.com/DAGWorks-Inc/hamilton/issues/37)) that allows you to query for nodes that match tags. With simple `=`, `in`, `not =`, `not in` type operations. The driver maintains a list and can do the queries for you. I don't really foresee the need for anything more complex, and this matches the common configuration patterns...

**Comment by kzk2000** [comment]
> [...] that match tags with simple =, in, not =, not in type operations.

this was exactly where I was going with this, works for me!

**Comment by skrawcz** [comment]
also - https://github.com/DAGWorks-Inc/hamilton/issues/293.

**Comment by skrawcz** [comment]
Otherwise I think lists of strings seem to be fine. Can't seem to recall why we didn't do this in the first place.

The other -- nested dicts can already be done with the namespacing/dot stuff.

```
@tag(**{'my.tag': 'tag_value'})
       def my_function(...) -> ...:
```

skrawcz:
@kzk2000 want to take https://github.com/DAGWorks-Inc/hamilton/pull/622 for a spin?

PR #622, opened 2024-01-08
Feature/620

skrawcz:
Implements a version of #620 .

```python
@tag( business_value=["CA", "US"] )`
def combo_node():
    ...

@tag( business_value=["US"] )`
def only_us_node():
    ...

@tag( business_value=["CA"], some_other_tag="BAR" )`
def only_ca_node():
    ...
```
then the following queries will return:
```python
# case A: returns only_ca_node() and combo_node()
dr.list_available_variables(tag_filter=dict(business_lines=["CA"]))
# case B: returns only_us_node() and combo_node()
dr.list_available_variables(tag_filter=dict(business_lines="US"))
# case C: returns all 3 nodes
dr.list_available_variables(tag_filter=dict(business_lines=["CA", "US"] ))
# case D: returns all 3 nodes
dr.list_available_variables(tag_filter=dict(business_lines=None ))
# case E: returns 0 nodes
dr.list_available_variables(tag_filter=dict(business_lines="UK" ))
# case F: returns 0 nodes
dr.list_available_variables(tag_filter=dict(business_lines="US", some_other_tag="FOO" )
# case G: returns 1 nodes
dr.list_available_variables(tag_filter=dict(business_lines="CA", some_other_tag="BAR" )
```

## Changes
 - `@tag` now can take in a list of strings
 -  `driver.list_available_variables()` now takes in a tag_filter argument
 - node.py has new function
 - tests

## How I tested this
 - locally
 - via unit tests

## Notes
 - this does not implement anything fancy with respect to query logic. We should implement more as the use case presents itself.

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -638,20 +638,57 @@ class Driver:
         return out
 
     @capture_function_usage
-    def list_available_variables(self) -> List[Variable]:
+    def list_available_variables(
+        self, *, tag_filter: Dict[str, Union[Optional[str], List[str]]] = None
+    ) -> List[Variable]:
         """Returns available variables, i.e. outputs.
 
-        These variables corresond 1:1 with nodes in the DAG, and contain the following information:
-        1. name: the name of the node
-        2. tags: the tags associated with this node
-        3. type: The type of data this node returns
-        4. is_external_input: Whether this node represents an external input (required from outside),
-        or not (has a function specifying its behavior).
+        These variables correspond 1:1 with nodes in the DAG, and contain the following information:
+
+            1. name: the name of the node
+            2. tags: the tags associated with this node
+            3. type: The type of data this node returns
+            4. is_external_input: Whether this node represents an external input (required from outside), \
+            or not (has a function specifying its behavior).
 
 
+        .. code-block:: python
+
+            # gets all
+            dr.list_available_variables()
+            # gets exact matching tag name and tag value
+            dr.list_available_variables({"TAG_NAME": "TAG_VALUE"})
+            # gets all matching tag name and at least one of the values in the list
+            dr.list_available_variables({"TAG_NAME": ["TAG_VALUE1", "TAG_VALUE2"]})
+            # gets all with matching tag name, irrespective of value
+            dr.list_available_variables({"TAG_NAME": None})
+            # AND query between the two tags (i.e. both need to match)
+            dr.list_available_variables({"TAG_NAME": "TAG_VALUE", "TAG_NAME2": "TAG_VALUE2"}
+
+        :param tag_filter: A dictionary of tags to filter by. Only nodes matching the tags and their values will
+            be returned. If the value for a tag is None, then we will return all nodes with that tag. If the value
+            is non-empty we will return all nodes with that tag and that value.
         :return: list of available variables (i.e. outputs).
         """
-        return [Variable.from_node(n) for n in self.graph.get_nodes()]
+        all_nodes = self.graph.get_nodes()
+        if tag_filter:
+            valid_filter_values = all(
+                map(
+                    lambda x: isinstance(x, str)
+                    or (isinstance(x, list) and len(x) != 0)
+                    or x is None,
+                    tag_filter.values(),
+                )
+            )
+            if not valid_filter_values:
+                raise ValueError("All tag query values must be a string or list of strings")
+            results = []
+            for n in all_nodes:
+                if node.matches_query(n.tags, tag_filter):
+                    results.append(Variable.from_node(n))
+        else:
+            results = [Variable.from_node(n) for n in all_nodes]
+        return results
 
     @capture_function_usage
     def display_all_functions(
--- a/hamilton/node.py
+++ b/hamilton/node.py
@@ -370,3 +370,36 @@ class Node(object):
             return __transform(self.callable(**kwargs), kwargs)
 
         return self.copy_with(callabl=new_callable, typ=__output_type)
+
+
+def matches_query(
+    tags: Dict[str, Union[str, List[str]]], query_dict: Dict[str, Optional[Union[str, List[str]]]]
+) -> bool:
+    """Check whether a set of node tags matches the query based on tags.
+
+    An empty dict of a query matches all tags.
+
+    :param tags: the tags of the node.
+    :param query_dict: of tag to value. If value is None, we just check that the tag exists.
+    :return: True if we have tags that match all tag queries, False otherwise.
+    """
+    # it's an AND clause between each tag and value in the query dict.
+    for tag, value in query_dict.items():
+        # if tag not in node we can return False immediately.
+        if tag not in tags:
+            return False
+        # if value is None -- we don't care about the value, just that the tag exists.
+        if value is None:
+            continue
+        node_tag_value = tags[tag]
+        if not isinstance(node_tag_value, list):
+            node_tag_value = [node_tag_value]
+        if not isinstance(value, list):
+            value = [value]
+        if set(value).intersection(set(node_tag_value)):
+            # if there is some overlap, we're good.
+            continue
+        else:
+            # else, return False.
+            return False
+    return True
--- a/tests/test_hamilton_driver.py
+++ b/tests/test_hamilton_driver.py
@@ -121,7 +121,58 @@ def test_driver_variables_exposes_tags():
     assert tags["a"] == {"module": "tests.resources.tagging", "test": "a"}
     assert tags["b"] == {"module": "tests.resources.tagging", "test": "b_c"}
     assert tags["c"] == {"module": "tests.resources.tagging", "test": "b_c"}
-    assert tags["d"] == {"module": "tests.resources.tagging"}
+    assert tags["d"] == {"module": "tests.resources.tagging", "test_list": ["us", "uk"]}
+
+
+@pytest.mark.parametrize(
+    "filter,expected",
+    [
+        (None, {"a", "b_c", "b", "c", "d"}),  # no filter
+        ({}, {"a", "b_c", "b", "c", "d"}),  # empty filter
+        ({"test": "b_c"}, {"b", "c"}),
+        ({"test": None}, {"a", "b", "c"}),
+        ({"module": "tests.resources.tagging"}, {"a", "b_c", "b", "c", "d"}),
+        ({"test_list": "us"}, {"d"}),
+        ({"test_list": "uk"}, {"d"}),
+        ({"test_list": ["uk"]}, {"d"}),
+        ({"module": "tests.resources.tagging", "test": "b_c"}, {"b", "c"}),
+        ({"test_list": ["nz", "uk"]}, {"d"}),
+        ({"test_list": ["us", "uk"]}, {"d"}),
+        ({"test_list": ["uk", "us"]}, {"d"}),
+        ({"test": ["b_c"]}, {"b", "c"}),
+        ({"test": ["b_c", "foo"]}, {"b", "c"}),
+    ],
+    ids=[
+        "filter with None passed",
+        "filter with empty filter",
+        "filter by single tag with extract decorator",
+        "filter with None value",
+        "filter with specific value",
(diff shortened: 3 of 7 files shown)


kzk2000:
@skrawcz haven't tried it yet, but just flagging my thoughts:

Do both of these return the same thing nodes?
```

dr.list_available_variables({"business_line": ["US", "UK"]}) 
dr.list_available_variables({"business_line": [ "UK", "US"]}) 
```

skrawcz:
> @skrawcz haven't tried it yet, but just flagging my thoughts:
> 
> Do both of these return the same thing nodes?
> 
> ```
> 
> dr.list_available_variables({"business_line": ["US", "UK"]}) 
> dr.list_available_variables({"business_line": [ "UK", "US"]}) 
> ```

yep it's order invariant.

I even handle the case:

```python
@tag(business_line="US")
def my_function() -> ...:
     ...
```

```python
# does not match
dr.list_available_variables({"business_line": ["US", "UK"]}) 
# does match
dr.list_available_variables({"business_line": ["US"]}) 
```

hamilton/driver.py
```diff
@@ -638,20 +638,80 @@ def raw_execute(
         return out
 
     @capture_function_usage
-    def list_available_variables(self) -> List[Variable]:
+    def list_available_variables(
```
elijahbenizzy, line 641: Feels like this should be a `query_available_variables` function, although it could work either way

hamilton/driver.py
```diff
@@ -638,20 +638,80 @@ def raw_execute(
...
+        .. code-block:: python
 
+            # gets all
+            dr.list_available_variables()
+            # gets exact matching
+            dr.list_available_variables({"TAG_NAME": "TAG_VALUE"})
+            # gets exact matching -- all tags must match, order invariant
+            dr.list_available_variables({"TAG_NAME": ["TAG_VALUE1", "TAG_VALUE2"]})
+            # gets all with matching tag, irrespective of value
+            dr.list_available_variables({"TAG_NAME": None})
```
elijahbenizzy, line 663: What about `or` gates? or are we just assuming `and` for now? I guess you can always list everything and query through...

hamilton/driver.py
```diff
@@ -638,20 +638,80 @@ def raw_execute(
...
+
+        :param tag_filter: A dictionary of tags to filter by. Only nodes matching the tags and their values will
+            be returned. If the value for a tag is None, then we will return all nodes with that tag. If the value
+            is non-empty we will return all nodes with that tag and that value.
         :return: list of available variables (i.e. outputs).
         """
-        return [Variable.from_node(n) for n in self.graph.get_nodes()]
+
+        def exclude_node(n: node.Node, query_dict: Dict[str, Optional[str]]) -> bool:
```
elijahbenizzy, line 671: This should probably be a separate, unit-testable function?

hamilton/driver.py
```diff
@@ -638,20 +638,80 @@ def raw_execute(
...
+            3. type: The type of data this node returns
+            4. is_external_input: Whether this node represents an external input (required from outside),
+            or not (has a function specifying its behavior).
 
+        .. code-block:: python
 
+            # gets all
+            dr.list_available_variables()
+            # gets exact matching
+            dr.list_available_variables({"TAG_NAME": "TAG_VALUE"})
```
elijahbenizzy, line 659: This API is confusing me quite a bit TBH. `TAG_VALUE` is a list means that they *all* have to match? BUt if its an individual it means they have to match any?

We should be opinionated and simple here. If not, we should leave querying up to the user... I was thinking:
1. Only query for individual tag_values in dict. Dicts imply `and`, and an individual one means its one of the decorated values
2. We leave open the option for more kw params later

```python
dr.select_variables(tag_match={"foo" : "bar", "baz": "qux"})
```

This would match:
- `@tag(foo="bar")`
- `@tag(baz="qux")`
- `@tag(foo=["bar", "qux"])

But not:
- `@tag(foo="not_bar")`
- `@tag(foo=["not_baz", "not_bar"])`
- `@tag(foo="bar", baz="not_qux")`
- `@tag(foo=["bar", "qux"], baz=["bar", "not_qux"])`

This is being opinionated that matching means any do exist, anything else is up to the user to specify/query independently.

tests/test_hamilton_driver.py
```diff
@@ -121,7 +121,56 @@ def test_driver_variables_exposes_tags():
...
+        "filter tag with list values - value 2",
+        "filter with two filter clauses",
+        "filter with with list values not exact",
+        "filter with with list values exact",
+        "filter with with list values exact order invariant",
+        "filter with with list values edge case one item match",
+        "filter with with list values edge case no results",
+    ],
+)
+def test_driver_variables_filters_tags(filter, expected):
```
elijahbenizzy, line 160: Looking at this test, we should have a function that takes in filters and node tags and tells whether they match. We shouldn't have to test the query function to do that with a driver and all.

elijahbenizzy (commented):
A few thoughts -- largely on API

tests/test_hamilton_driver.py
skrawcz: Fair.

hamilton/driver.py
skrawcz: @kzk2000 thoughts on a new or old function? 

I'm more inclined to have one function.

hamilton/driver.py
skrawcz: The issue is how do you interpret a list of tag values?

E.g. Given:

```python
@tag(business_lines=["CA", "TX"])
def my_func() -> ...:
   pass

```
```python
# e.g. I want to get all nodes with a particular business line (it could be variable in length)
# should this be an exact match if someone passes a single item list? 
dr.list_available_variables(tag_filter=dict(business_lines=["CA"]))

# to me the above is equivalent to this -- because a user could also write this:
dr.list_available_variables(tag_filter=dict(business_lines="CA"))

# Why?
# Well, why would someone put a tag on a function with only a single value in it?
@tag(business_lines=["CA"]) # < -- this wouldn't make sense to me.
def my_func() -> ...:
   pass

# e.g. I want to get all nodes that correspond to Cali AND Texas AND Florida:
# what about this? I think this is an exact match requirement (i.e. AND), not an OR.
dr.list_available_variables(tag_filter=dict(business_lines=["CA", "TX", "FL"]))

# e.g. I want to get all nodes that hit Cali OR Texas OR Florida
# If you want an OR you would have to do multiple calls and union the results.
(
 dr.list_available_variables(tag_filter=dict(business_lines="CA"))  | 
 dr.list_available_variables(tag_filter=dict(business_lines="TX"))  | 
 dr.list_available_variables(tag_filter=dict(business_lines="FL"))  | 
)
```

skrawcz: @kzk2000 any thoughts on lists of values? It would be simpler to not allow it.

kzk2000: @skrawcz Whatever feels more intuitive is all right by me.

Say
```
@tag( business_value=["CA", "US"] )`
def combo_node():
    ...

@tag( business_value=["US"] )`
def only_us_node():
    ...

@tag( business_value=["CA"] )`
def only_ca_node():
    ...
```


That is, I would treat the last case as OR operator, if any one of the list entries matches, return that node.
I'm totally fine with making this more explicit as you suggested here.
```
(
 dr.list_available_variables(tag_filter=dict(business_lines="CA"))  | 
 dr.list_available_variables(tag_filter=dict(business_lines="TX"))  | 
 dr.list_available_variables(tag_filter=dict(business_lines="FL"))  | 
) # you'd have to make them sets for this to work but you get the picture
```

Bottom line as long as `dr.list_available_variables()``` returns all nodes, I can easily run any list comprehension filter like this myself.

[ x for x in  dr.list_available_variables() if x.tag.get('business_line", None) in ['US', 'CA']]

or use sets as you suggested

skrawcz: Okay I think I might switch to the OR interpretation then.

skrawcz: Will try to have something by Friday

skrawcz: @elijahbenizzy @kzk2000 

For this case:
```python
dr.list_available_variables(tag_filter=dict(business_lines=["CA"], some_other_tag="foo" ))
```
I'm interpreting it as an AND query, while the values in the list are OR. Thoughts?

tests/test_hamilton_driver.py
skrawcz: done. moved the function to node module.

hamilton/driver.py
skrawcz: done.

hamilton/driver.py
skrawcz: changed in latest commit to match more of the discussion below.

hamilton/driver.py
elijahbenizzy: So yeah, offline we've synced up and decided that tags are *or* queries. This leaves the door open with:
```python
tag_filter=dict(foo=or_("bar", "baz"))
tag_filter=dict(foo=and_("bar", "baz"))
```

or something like this

hamilton/node.py
```diff
@@ -370,3 +370,55 @@ def new_callable(**kwargs) -> Any:
...
+def matches_query(
+    tags: Dict[str, Union[str, List[str]]], query_dict: Dict[str, Optional[Union[str, List[str]]]]
+) -> bool:
+    """Check whether a set of node tags matches the query based on tags.
+
+    :param tags: the tags of the node.
+    :param query_dict: of tag to value. If value is None, we just check that the tag exists.
+    :return: True if we have tags that match all tag queries, False otherwise.
+    """
+    if not query_dict:
```
elijahbenizzy, line 384: Feels like `tag_filter={}` should trigger on everything? Or not be allowed, cause its pointless? No filter IMO means it matches. Also that means you can just remove this.

hamilton/driver.py
```diff
@@ -638,20 +638,57 @@ def raw_execute(
...
+            dr.list_available_variables({"TAG_NAME": "TAG_VALUE", "TAG_NAME2": "TAG_VALUE2"}
+
+        :param tag_filter: A dictionary of tags to filter by. Only nodes matching the tags and their values will
+            be returned. If the value for a tag is None, then we will return all nodes with that tag. If the value
+            is non-empty we will return all nodes with that tag and that value.
         :return: list of available variables (i.e. outputs).
         """
-        return [Variable.from_node(n) for n in self.graph.get_nodes()]
+        all_nodes = self.graph.get_nodes()
+        if tag_filter:
```
elijahbenizzy, line 674: This is either empty, or `None`, which then conflicts (although not seriously) with the `matches_query` function. Let's change `matches_query` to return true if its empty then at least it is consistent, the truthiness here is fine then (cause its the same implied behavior).

hamilton/node.py
```diff
@@ -370,3 +370,55 @@ def new_callable(**kwargs) -> Any:
...
+    # it's an AND clause between each tag and value in the query dict.
+    for tag, value in query_dict.items():
+        # if tag not in node we can return False immediately.
+        if tag not in tags:
+            return False
+        # if value is None -- we don't care about the value, just that the tag exists.
+        if value is None:
+            continue
+        node_tag_value = tags[tag]
+        if isinstance(node_tag_value, list) and isinstance(value, list):
```
elijahbenizzy, line 395: cleaner way to do this is to make the node_tag_value a list regardless, then you only have one code path

elijahbenizzy (approved):
LGTM

hamilton/node.py
skrawcz: oh sure -- can do that.

hamilton/node.py
skrawcz: done.

hamilton/driver.py
skrawcz: To confirm:


> dr.list_available_variables(tag_filter=dict(business_lines=["CA", "US"], some_other_tag="foo" ))

this to me reads:
    business_lines=CA  OR business_lines=US 
AND 
    some_other_tag=foo

PR #625, opened 2024-01-10
Adds Simple MLFlow materializer

skrawcz:
This isn't ready to be shipped. But shows how one could create something that saves to MLFlow easily.

Things to consider -- what is the structure a model function should output if we want to capture extra things?

## Checklist


zilto:
Closing in favor of #945

PR #632, opened 2024-01-12
Adds TQDM adapter to show progress

skrawcz:
This is not added by default. But something people can add to their driver. This will only work
if you have TQDM installed.

```python

from hamilton import lifecycle

dr = (
    driver.Builder()
    .with_config({})
    .with_modules(some_modules)
    .with_adapters(lifecycle.TQDM(desc="DAG-NAME"))
    .build()
)

```

## Changes
 - adds TQDM progress bar adapter

## How I tested this
 - locally

## Notes
 - this is conditionally available based on whether TQDM is installed or not.
 - [ ] need to update docs somewhere I think.

## Checklist

### DIFF
--- a/hamilton/lifecycle/__init__.py
+++ b/hamilton/lifecycle/__init__.py
@@ -13,15 +13,6 @@ from .default import PDBDebugger, PrintLn  # noqa: F401
 
 PrintLnHook = PrintLn  # for backwards compatibility -- this will be removed in 2.0
 
-try:
-    from .conditional_tqdm import TQDM  # noqa: F401
-except ImportError:
-    TQDM = None
-
-optional = []
-if TQDM is not None:
-    optional.append("TQDM")
-
 # All the following types are public facing
 __all__ = [
     "LifecycleAdapter",
@@ -36,4 +27,4 @@ __all__ = [
     "GraphExecutionHook",
     "NodeExecutionMethod",
     "StaticValidator",
-] + optional
+]
--- a/hamilton/lifecycle/__init__.py
+++ b/hamilton/lifecycle/__init__.py
@@ -1,4 +1,4 @@
-from .api import (
+from .api import (  # noqa: F401
     EdgeConnectionHook,
     GraphAdapter,
     GraphExecutionHook,
@@ -8,8 +8,17 @@ from .api import (
     ResultBuilder,
     StaticValidator,
 )
-from .base import LifecycleAdapter
-from .default import PDBDebugger, PrintLnHook
+from .base import LifecycleAdapter  # noqa: F401
+from .default import PDBDebugger, PrintLnHook  # noqa: F401
+
+try:
+    from .conditional_tqdm import TQDMHook  # noqa: F401
+except ImportError:
+    TQDMHook = None
+
+optional = []
+if TQDMHook is not None:
+    optional.append("TQDMHook")
 
 # All the following types are public facing
 __all__ = [
@@ -24,4 +33,4 @@ __all__ = [
     "GraphExecutionHook",
     "NodeExecutionMethod",
     "StaticValidator",
-]
+] + optional
--- a/hamilton/lifecycle/__init__.py
+++ b/hamilton/lifecycle/__init__.py
@@ -9,7 +9,9 @@ from .api import (  # noqa: F401
     StaticValidator,
 )
 from .base import LifecycleAdapter  # noqa: F401
-from .default import PDBDebugger, PrintLnHook  # noqa: F401
+from .default import PDBDebugger, PrintLn  # noqa: F401
+
+PrintLnHook = PrintLn  # for backwards compatibility -- this will be removed in 2.0
 
 try:
     from .conditional_tqdm import TQDMHook  # noqa: F401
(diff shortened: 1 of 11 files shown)


hamilton/lifecycle/conditional_tqdm.py
```diff
@@ -0,0 +1,105 @@
...
+from hamilton.lifecycle import base
+
+
+class TQDMHook(
+    base.BasePreGraphExecute,
+    base.BasePreNodeExecute,
+    base.BasePostNodeExecute,
+    base.BasePostGraphExecute,
+):
+    """A hook that uses tqdm to show progress bars for the graph execution.
```
elijahbenizzy, line 15: Lingo question -- I've been using adapter for a combination of hooks/methods. But this is just hooks so maybe its a hook?

hamilton/lifecycle/conditional_tqdm.py
```diff
@@ -0,0 +1,105 @@
+from typing import Any, Dict, List, Optional
+
+import tqdm
+
+from hamilton import graph, node
+from hamilton.lifecycle import base
+
+
+class TQDMHook(
```
elijahbenizzy, line 9: Maybe call it `ProgressBarHook`?

hamilton/lifecycle/conditional_tqdm.py
```diff
@@ -0,0 +1,105 @@
...
+        node_: node.Node,
+        kwargs: Dict[str, Any],
+        success: bool,
+        error: Optional[Exception],
+        result: Optional[Any],
+        task_id: Optional[str] = None,
+    ):
+        self.progress_bar.update(1)
+
+    def post_graph_execute(
```
elijahbenizzy, line 92: To use external APIs, all we need is this + `pre` right?

hamilton/lifecycle/__init__.py
```diff
@@ -1,12 +1,21 @@
...
-from .default import PDBDebugger, PrintLnHook
+from .base import LifecycleAdapter  # noqa: F401
+from .default import PDBDebugger, PrintLnHook  # noqa: F401
+
+try:
+    from .conditional_tqdm import TQDMHook  # noqa: F401
+except ImportError:
+    TQDMHook = None
+
+optional = []
```
elijahbenizzy, line 16: Kinda think this should be default in the driver, no? Only if TQDM is installed?

hamilton/lifecycle/conditional_tqdm.py
elijahbenizzy: TQDM is an implementation detail... OTOH we could have other progress bars.

hamilton/lifecycle/conditional_tqdm.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        *,
+        run_id: str,
+        graph: graph.FunctionGraph,
+        success: bool,
+        error: Optional[Exception],
+        results: Optional[Dict[str, Any]],
+    ):
+        name_part = "Execution Complete!"
+        padding = " " * (self.node_name_target_width - len(name_part))
```
elijahbenizzy, line 102: This could end up negative right? Guess that'll still work:

```
>>> "_"*-1
''
```

hamilton/lifecycle/conditional_tqdm.py
```diff
@@ -0,0 +1,105 @@
...
+        self.progress_bar.set_description_str(f"{self.desc} -> {name_part + padding}")
+        self.progress_bar.set_postfix({})
```
elijahbenizzy, line 104: OK, so just to be clear, you *don't* like having the node name as part of the postfix, right? I like it but either way is fine.

hamilton/lifecycle/__init__.py
skrawcz: can people remove it? if so, how?

hamilton/lifecycle/conditional_tqdm.py
skrawcz: it goes in the adapter argument of the driver though.

Adapters have types is the way I see it.

hamilton/lifecycle/conditional_tqdm.py
skrawcz: sure -- if someone puts in something small.

hamilton/lifecycle/conditional_tqdm.py
skrawcz: Simplified to just TQDM

hamilton/lifecycle/conditional_tqdm.py
skrawcz: Nah I think it's specific because it requires a dependency.

hamilton/lifecycle/conditional_tqdm.py
skrawcz: yes

hamilton/lifecycle/conditional_tqdm.py
skrawcz: I have both here just to show what it would look like.

elijahbenizzy (approved):

PR #633, opened 2024-01-12
Lifecycle validators

elijahbenizzy:
Adds a lifecycle validation capability. We had two types of lifecycle customizations:
1. Methods (replaced some component)
2. Hooks (augmented some components)

Now we have validators, which return a [bool, str] result, consisting of [success, message]. These allow erroring out if conditions are not met. We add one for validating nodes, and one for validating DAGs.

We also do a bunch of repo hygiene/add a few more public-facing capabilities.

https://github.com/DAGWorks-Inc/hamilton/issues/631

## Checklist

- [x] PR has an informative and human-readable title (this will be pulled into the release notes)
- [ ] Changes are limited to a single goal (no scope creep)
- [x] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [x] Any _change_ in functionality is tested
- [x] New functions are documented (with a description, list of inputs, and expected output)
- [x] Placeholder code is flagged / future TODOs are captured in comments
- [x] Project documentation has been updated if adding/changing functionality.

PR #635, opened 2024-01-14
export_json method in Driver.

alechewitt:
This PR creates a new method on the driver `export_json`. The method returns a JSON string representation of the graph.

The method is intended to be used by clients wishing to view and analyze their graphs with other visualizations tools. 

Idea referenced in discussion here: https://github.com/DAGWorks-Inc/hamilton/discussions/624

## Changes
 - New method `export_json` created in `driver.py`
 - New method `as_dict` created in `node.py`. This returns a JSON serializable dictionary of the Node.
 
## How I tested this
New tests created for both new methods

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1,5 +1,6 @@
 import abc
 import functools
+import json
 import logging
 import operator
 import sys
@@ -16,6 +17,7 @@ import pandas as pd
 
 from hamilton import common, graph_types, htypes
 from hamilton.execution import executors, graph_functions, grouping, state
+from hamilton.graph_types import HamiltonNode
 from hamilton.io import materialization
 from hamilton.io.materialization import ExtractorFactory, MaterializerFactory
 from hamilton.lifecycle import base as lifecycle_base
@@ -897,6 +899,28 @@ class Driver:
             deduplicate_inputs=deduplicate_inputs,
             show_schema=show_schema,
         )
+
+    @capture_function_usage
+    def export_execution(
+        self,
+        final_vars: List[str],
+        inputs: Dict[str, Any] = None,
+        overrides: Dict[str, Any] = None,
+    ) -> str:
+        """Method to create JSON representation of the Graph.
+
+        :param final_vars: The final variables to compute.
+        :param inputs: Optional. The inputs to the DAG.
+        :param overrides: Optional. Overrides to the DAG.
+        :return: JSON string representation of the graph.
+        """
+        nodes, user_nodes = self.graph.get_upstream_nodes(final_vars, inputs, overrides)
+        Driver.validate_inputs(self.graph, self.adapter, user_nodes, inputs, nodes)
+        all_nodes = nodes | user_nodes
+
+        hamilton_nodes = [HamiltonNode.from_node(n).as_dict() for n in all_nodes]
+        sorted_nodes = sorted(hamilton_nodes, key=lambda x: x["name"])
+        return json.dumps({"nodes": sorted_nodes})
 
     @capture_function_usage
     def has_cycles(
--- a/hamilton/graph_types.py
+++ b/hamilton/graph_types.py
@@ -1,8 +1,10 @@
 """Module for external-facing graph constructs. These help the user navigate/manage the graph as needed."""
+import inspect
 import typing
 from dataclasses import dataclass
 
 from hamilton import htypes, node
+from hamilton.htypes import get_type_as_string
 
 # This is a little ugly -- its just required for graph build, and works
 # This indicates a larger smell though -- we need to have the right level of
@@ -27,6 +29,20 @@ class HamiltonNode:
     documentation: typing.Optional[str]
     required_dependencies: typing.Set[str]
     optional_dependencies: typing.Set[str]
+
+    def as_dict(self):
+        """Create a dictionary representation of the Node that is JSON serializable"""
+        return {
+            "name": self.name,
+            "tags": self.tags,
+            "output_type": get_type_as_string(self.type) if get_type_as_string(self.type) else "",
+            "required_dependencies": sorted(self.required_dependencies),
+            "optional_dependencies": sorted(self.optional_dependencies),
+            "source": inspect.getsource(self.originating_functions[0])
+            if self.originating_functions
+            else None,
+            "documentation": self.documentation,
+        }
 
     @staticmethod
     def from_node(n: node.Node) -> "HamiltonNode":
--- a/tests/test_driver_export.py
+++ b/tests/test_driver_export.py
@@ -1,0 +1,96 @@
+import json
+
+from hamilton import driver
+
+from tests.resources.dynamic_parallelism import no_parallel
+
+EXPECTED_JSON = {
+    "nodes": [
+        {
+            "name": "final",
+            "tags": {"module": "tests.resources.dynamic_parallelism.no_parallel"},
+            "output_type": "int",
+            "required_dependencies": ["sum_step_squared_plus_step_cubed"],
+            "optional_dependencies": [],
+            "source": (
+                "def final(sum_step_squared_plus_step_cubed: int) -> int:\n"
+                '    print("finalizing")\n'
+                "    return sum_step_squared_plus_step_cubed\n"
+            ),
+            "documentation": "",
+        },
+        {
+            "name": "number_of_steps",
+            "tags": {"module": "tests.resources.dynamic_parallelism.no_parallel"},
+            "output_type": "int",
+            "required_dependencies": [],
+            "optional_dependencies": [],
+            "source": "def number_of_steps() -> int:\n    return 5\n",
+            "documentation": "",
+        },
+        {
+            "name": "step_cubed",
+            "tags": {"module": "tests.resources.dynamic_parallelism.no_parallel"},
+            "output_type": "int",
+            "required_dependencies": ["steps"],
+            "optional_dependencies": [],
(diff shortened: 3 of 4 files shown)


skrawcz:
We'll push https://github.com/DAGWorks-Inc/hamilton/pull/633 first, and then this will require some changes.

elijahbenizzy:
> We'll push #633 first, and then this will require some changes.

To clarify, we're now exposing a `HamiltonGraph` object to the user. I think its clean if the driver's `export_json` function just spits out a json version of this. Planning to release shortly, then we can rebase this off of main and I can point you in the right direction, should be a very quick change! 

Thanks for the contribution -- quite excited by this.

zilto:
Sorry for being late to the discussion. I know we purposefully avoid having Hamilton depend on Networkx, but I previously had success exporting the Hamilton DAG by using `hamilton.graph.create_networkx_graph()` ([ref](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/graph.py#L552C5-L552C26) and then using the [networkx supported formats](https://networkx.org/documentation/stable/reference/readwrite/index.html) (GEXF, GML, GraphML, JSON, LEDA, etc.). I used the [cytoscapejs format](https://js.cytoscape.org/).

For instance, the following code worked nicely in my previous projects:

```python
from hamilton.graph import FunctionGraph, create_networkx_graph
import networkx as nx

import my_functions  # a module you would pass to the Driver

fgraph = FunctionGraph(my_functions, config=dict())  # can pass config
nodes = fgraph.get_nodes()
nx_graph = create_networkx_graph(set(nodes), set(), "need-a-name-argument")
graph_json = networkx.cytoscape.cytoscape_data(nx_graph)
```

hamilton/driver.py
```diff
@@ -900,6 +901,30 @@ def visualize_execution(
...
+        nodes, user_nodes = self.graph.get_upstream_nodes(final_vars, inputs, overrides)
+        Driver.validate_inputs(self.graph, self.adapter, user_nodes, inputs, nodes)
+        all_nodes = nodes | user_nodes
+
+        # Create JSON representation of the graph
+        # TODO: Do we also want to include information on whether the node is IS_USER_INPUT and IS_OUTPUT
+        node_dicts = [node.as_dict() for node in all_nodes]
+        sorted_nodes = sorted(node_dicts, key=lambda x: x["name"])
+        return json.dumps(sorted_nodes)
+
```
skrawcz, line 927: So we want this to:

1. Use https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/graph_types.py#L17
2. Add the as_dict() function to HamiltonNode
3. change this function to mention it is exporting only the what would be executed given the final vars, inputs, and overrides. Maybe this means the name should incorporate `execution` in it to mirror the other functions in the driver?

@elijahbenizzy anything to add/disagree with?

elijahbenizzy: @skrawcz mostly. I think that it could be one step cleaner -- we have `HamiltonGraph`, which has a `from_graph` function. We should have an `as_dict` function there that delegates to the one in `HamiltonNode`. Then this just has to call `from_graph` on `self.graph` and it should be good!

skrawcz: From graph wouldn't help here because the graph is being filtered down though?

alechewitt: Yes, my thoughts were this method would be similar to the `visualize_execution` method. It accepts a list of `final_vars` and then constructs a JSON representation of the graph that you would see with the `visualize_execution` method, rather than a JSON representation of the entire `FunctionGraph`.

On adding the `as_dict()` method to the `HamiltonNode`. There are some things omitted from the HamiltonNode that are in the `Node` that I was originally exporting. Specfically:
 - [source](https://github.com/DAGWorks-Inc/hamilton/pull/635/files#diff-6b4a15b6f5263150ab8674431da707b5ea1c4a855e07a42ffb79ff5aa56cea86R224) - This is the source code of the `callable` function. I notice the `callable` isn't in the the `HamiltonNode`, is this by design?  There is the `originating_functions`, However I could not work out what these functions are, I think I need to spend longer reading the source code.
 - [depended_on_by](https://github.com/DAGWorks-Inc/hamilton/pull/635/files#diff-6b4a15b6f5263150ab8674431da707b5ea1c4a855e07a42ffb79ff5aa56cea86R223) - This is also not in the HamiltonNode.

On the name, I agree it is a bit confusing at the minute. Some other suggestions (from chatgpt):
 - export_execution
 - serialize_execution
 - jsonify_execution_path

skrawcz: > There is the originating_functions,

Node can have a bunch of wrapped functions. Originating functions is what decorators add to. So the source code you're after will be in the first entry that you'd want (IIRC).

> [depended_on_by](https://github.com/DAGWorks-Inc/hamilton/pull/635/files#diff-6b4a15b6f5263150ab8674431da707b5ea1c4a855e07a42ffb79ff5aa56cea86R223) - This is also not in the HamiltonNode.

That's true, though it's not required to recreate the graph. dependend_on_by is a mirror of dependencies, I just decided to have two to know which way you're traversing the graph. So you could reconstruct this part yourself.

elijahbenizzy: Ahh yep, see your point. Then yeah, I'm a fan of `export_execution`, and using the nodes... 

To add to what @skrawcz said, you can get the following:

1. `source` -- from `inspect.getsourcelines(originating_functions[0]) if originating_functions else None` (or something like that).  This has to do with the way hamilton works - the `callable` does not necessarily correspond to the actual function. Instead, it is sometimes wrapped with some special logic. The originating_functions are the one you want, as they're guarenteed to be the ones you defined (if they exist, there are some edge-cases of dynamically generated nodes). IIRC it is always the first. In 99% of the cases there is just one, so use the first in the list.
2. `depended_on_by` -- I didn't want to include this as it encodes the same information twice -- the fact that we included it is a little odd TBH (although it does make traversing it easier).

I think this should give you enough to work through it?

hamilton/driver.py
```diff
@@ -898,6 +900,28 @@ def visualize_execution(
...
+        :param overrides: Optional. Overrides to the DAG.
+        :return: JSON string representation of the graph.
+        """
+        nodes, user_nodes = self.graph.get_upstream_nodes(final_vars, inputs, overrides)
+        Driver.validate_inputs(self.graph, self.adapter, user_nodes, inputs, nodes)
+        all_nodes = nodes | user_nodes
+
+        hamilton_nodes = [HamiltonNode.from_node(n).as_dict() for n in all_nodes]
+        sorted_nodes = sorted(hamilton_nodes, key=lambda x: x["name"])
+        return json.dumps(sorted_nodes)
```
elijahbenizzy, line 923: Thoughts -- should we put this under a key "nodes"? That way it leaves room for other things (not sure what they would be but flexibility is nice).

hamilton/graph_types.py
```diff
@@ -28,6 +30,20 @@ class HamiltonNode:
     required_dependencies: typing.Set[str]
     optional_dependencies: typing.Set[str]
 
+    def as_dict(self):
```
elijahbenizzy, line 33: `dataclasses.as_dict` works  as well, but given that we handle other serialization pieces, I think this is better.

elijahbenizzy (commented):
Looks good! Only thing I'd change is the shape of the export to have the `nodes` key, but I'd be curious as to your thoughts.

hamilton/driver.py
alechewitt: Yes I agree. As this is a public method it will be harder to change in the future so we should make it extensible. I also don't see any downsides to putting them under the key `nodes`. I'll update the PR later today.

hamilton/graph_types.py
alechewitt: In the past I have had success with using [Pydantic](https://docs.pydantic.dev/latest/concepts/serialization/#custom-serializers) to define custom serialization for an individual field. However I don't think it is worth adding adding a dependency for just this use case.

elijahbenizzy (approved):
Looks good! Thanks for the contribution.

skrawcz:
just thought that we probably want `json` in the function name?  Could there be other export formats? 🤔

alechewitt:
> just thought that we probably want `json` in the function name? Could there be other export formats? 🤔

I think it depends on how you see other export formats being created in the future. I see two options, 
 1. Modify this function to have a parameter to define the format eg: `def export_execution(format='JSON')`
 1. Create a separate method for each export format as and when they arise. eg: `export_execution_json`, `export_execution_xml` etc.

I'm happy to change the name to whatever you think is most appropriate. However, if no strong opinion now, I think it will be fine to make the decision when the next format comes along without any issue.

elijahbenizzy:
> > just thought that we probably want `json` in the function name? Could there be other export formats? 🤔
> 
> I think it depends on how you see other export formats being created in the future. I see two options,
> 
> 1. Modify this method to have a parameter to define the format eg: `def export_execution(format='JSON')`
> 2. Create a separate method for each export format as and when they arise. eg: `export_execution_json`, `export_execution_xml` etc.
> 
> I'm happy to change the name to whatever you think is most appropriate. However, if no strong opinion now, I think it will be fine to make the decision when the next format comes along without any issue.

@skrawcz I liked `export_execution`, and I think @alechewitt you have a good point on the parameter. Also, `str` is a reasonable return type -- pretty much every export could be represented as a `str`. So yeah, should we want to, we can later do `export_execution(format=...)`, where `format` defaults to `json`

Issue #637, opened 2024-01-15
Schema only displayed for one node when multiple nodes have the same schema


roelbertens:
# Current behavior

When two nodes have the same schema it is only displayed for one of the nodes.
It should of course be displayed for both of them.

Note: sometimes it displays the column at the first node and sometimes at the second.

## Screenshots
<img width="842" alt="image" src="https://github.com/DAGWorks-Inc/hamilton/assets/18281928/fcda81a6-161e-4601-bd21-cf78f6d43252">

## Steps to replicate behavior
Module debug.py
```
import pyspark.sql as ps
from pyspark.sql import functions as sf
from hamilton.function_modifiers import schema
WORLD_OF_WARCRAFT__SCHEMA = [("zone", "str"), ("level", "int"), ("avatarId", "int")]

def spark_session() -> ps.SparkSession:
    return ps.SparkSession.builder.master("local[1]").getOrCreate()

@schema.output(*WORLD_OF_WARCRAFT__SCHEMA)
def world_of_warcraft(spark_session: ps.SparkSession) -> ps.DataFrame:
    return spark_session.read.parquet("data/wow.parquet")

@schema.output(*WORLD_OF_WARCRAFT__SCHEMA)
def world_of_warcraft_level10(world_of_warcraft: ps.DataFrame) -> ps.DataFrame:
    return world_of_warcraft.filter(sf.col("level") == 10)
```

And then run
```
debug = importlib.import_module("debug")
dr = driver.Driver({}, debug)
dr.display_all_functions(show_schema=True)
```

## Library & System Information
sf-hamilton-1.44.0

elijahbenizzy:
Nice find! Bug here is that it is uniquely keyed on the name inside the cluster, which is not unique. I think the solution is to separate out the name from the label, pretty sure that's easy.

Evidence -- if I make it a UUID, it works:
![image](https://github.com/DAGWorks-Inc/hamilton/assets/3886307/14d88c86-419b-4295-84e5-941a38ed2736)

elijahbenizzy:
@roelbertens got a fix! Releasing with the next batch (plan is today). https://github.com/DAGWorks-Inc/hamilton/pull/638

PR #638, opened 2024-01-15
Fixes schema viz so it doesn't drop duplicates

elijahbenizzy:
See #637. This was a bug in which we were reusing the global namespace for columns. This fixes it by prepending with a unique name.

## Checklist


hamilton/graph.py
```diff
@@ -487,7 +487,10 @@ def _insert_space_after_colon(col: str) -> str:
                 c.attr(**cluster_node_style)
                 cols = _create_equal_length_cols(n.tags.get(schema.INTERNAL_SCHEMA_OUTPUT_KEY))
                 for i in range(len(cols)):
-                    c.node(cols[i], **field_node_style)
+                    # We may want to change the name. We have a difference name than label,
```
skrawcz, line 490:
```suggestion
                    # We may want to change the name. We have a different name than label,
```

PR #642, opened 2024-01-15
Prototypes two approaches to customizing graphviz

skrawcz:
1. Using the `@style` decorator. The assumption is the key is a node style attribute, and the value is a valid value.
2. Using a passed in "style" function. It can operate over the node name, node output type, tags, and node type (function, vs config).

The example image shows red coming from the style tag. Blue coming from the function.

This only wires up the `display_all_functions()` method in the driver to pass through the custom style function.

What's not included -- styling the legend.
The legend seems a bit more complex to customize:

 - how do you add nodes to it?
 - how do you style those nodes?

Other assumptions:

 - people know graphviz -- will skip validating values for now.
 - custom styles are applied last

## Changes
 - graph
 - driver
 - metadata tags

## How I tested this
 - locally

## Notes
 - this is just a prototype POC
 - we'd probably want the custom function to be kwargs only, so that we could add things as required? e.g. what if someone wanted to display source code -- how would we facilitate that?

See these -- and then look at the code:
a_path dot
my_dag dot
my_full_dag dot

## Checklist


skrawcz:
Notes:
 - `@style` would need `@style_outputs` equivalent.
 -  the custom function helps decouple style from graph -- and we think this is our preference for people
 
TODO:
 - finish wiring through custom style function
 - have @roelbertens try things out -- we'd like feedback on decorator approach, versus the style function.

examples/hello_world/my_script.py
```diff
@@ -41,4 +43,13 @@
...
     strict_path_visualization=False,
 )
-# dr.display_all_functions("./my_full_dag.dot", {"format": "png"})
+
+
+def custom_style(name: str, output_type: Type, tags: dict, node_type: str) -> dict:
+    print(name, output_type, tags, node_type)
+    if tags.get("some_key") == "some_value":
+        return {"fillcolor": "blue"}
+    return {}
```
skrawcz, line 52: To hack in handling legend stuff:
```suggestion
def custom_style(name: str, output_type: Type, tags: dict, node_type: str) -> dict:
    print(name, output_type, tags, node_type)
    if node_type == "legend":
        return {"LLM": {"fillcolor": "blue"}, ...}
    else:
        if tags.get("some_key") == "some_value":
            return {"fillcolor": "blue"}
    return {}
```

skrawcz: Or this outputs

```
{style}, name_of_style | None
```
We then use name_of_style for legend, if None, we don't include.

elijahbenizzy: Pass in the new `HamiltonNode` object -- has all this data + more.

skrawcz:
@roelbertens mind playing around with this branch and seeing if you can do what you want? I would like to see both approaches to see what's better.

Note: I'm leaning towards _not_ supporting the decorator because it doesn't leave a good avenue for modifying the legend. Instead having you pass in a styling function that can inspect the node and also add to the legend.

By using a function you can annotate nodes with information that's useful via tags, and then use that to determine the style. This then means no code modification is necessary if you want to change how you display the graph...

roelbertens:
Hi @skrawcz thanks for all your work in this, really appreciated!

I would prefer the approach using `custom_style` because it is more flexible and also changes the legend.
I think with some simple example it will be understandable for people to use.
I would directly start using it already.

skrawcz:
> Hi @skrawcz thanks for all your work in this, really appreciated!
> 
> I would prefer the approach using `custom_style` because it is more flexible and also changes the legend. I think with some simple example it will be understandable for people to use. I would directly start using it already.

Cool! I propose:

1. Dropping the decorators
2. Only allowing the custom style function -- need to document it more.
3. Probably refactoring some internal functions to expose the constants that people might want to depend on.
4. We should be able to ship this next week.

zilto:
I believe this can be closed since #663 was merged
.

PR #644, opened 2024-01-16
DDOG Integration

elijahbenizzy:
Iterating -- plan of attack/thoughts:

- [x] Opted to store a cache myself and prune rather than using context as I want to leave the door open for more clever uses of this. I just link parents and store a dict
- [x] Should be easy to get tasks working, just need to expose the API
- [x] Should also be possible to add causal links between spans as well. TBD on how to use the DDOG API exactly.
- [x] Add all node tags to DDOG cause why not :shrug:
- [ ] do OpenTel (out of scope for this, that's next)

What this looks like:

image

## Changes
- DDOG
- An extra (easter-egg) hook
- Improvements to running hooks in the case of exceptions

## How I tested this
- ran ddog multiple times locally under different contexts
- tests passed for existing ones (what I was fixing for was not worth testing for TBH, parity with current behavior is good)

## Checklist

### DIFF
--- a/hamilton/lifecycle/base.py
+++ b/hamilton/lifecycle/base.py
@@ -751,19 +751,21 @@ class LifecycleAdapterSet:
     def _get_lifecycle_hooks(
         self,
     ) -> Tuple[Dict[str, List[LifecycleAdapter]], Dict[str, List[LifecycleAdapter]]]:
-        sync_hooks = collections.defaultdict(set)
-        async_hooks = collections.defaultdict(set)
+        sync_hooks = collections.defaultdict(list)
+        async_hooks = collections.defaultdict(list)
         for adapter in self.adapters:
             for cls in inspect.getmro(adapter.__class__):
                 sync_hook = getattr(cls, SYNC_HOOK, None)
                 if sync_hook is not None:
-                    sync_hooks[sync_hook].add(adapter)
+                    if adapter not in sync_hooks[sync_hook]:
+                        sync_hooks[sync_hook].append(adapter)
                 async_hook = getattr(cls, ASYNC_HOOK, None)
                 if async_hook is not None:
-                    async_hooks[async_hook].add(adapter)
+                    if adapter not in async_hooks[async_hook]:
+                        async_hooks[async_hook].append(adapter)
         return (
-            {hook: list(adapters) for hook, adapters in sync_hooks.items()},
-            {hook: list(adapters) for hook, adapters in async_hooks.items()},
+            {hook: adapters for hook, adapters in sync_hooks.items()},
+            {hook: adapters for hook, adapters in async_hooks.items()},
         )
 
     def _get_lifecycle_methods(
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -2,6 +2,7 @@
 import logging
 import pdb
 import pprint
+import random
 import time
 from typing import Any, Callable, Dict, List, Optional, Union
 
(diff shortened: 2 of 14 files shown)


hamilton/lifecycle/base.py
```diff
@@ -751,19 +751,21 @@ def _get_lifecycle_validators(
...
-        async_hooks = collections.defaultdict(set)
+        sync_hooks = collections.defaultdict(list)
+        async_hooks = collections.defaultdict(list)
         for adapter in self.adapters:
             for cls in inspect.getmro(adapter.__class__):
                 sync_hook = getattr(cls, SYNC_HOOK, None)
                 if sync_hook is not None:
-                    sync_hooks[sync_hook].add(adapter)
+                    if adapter not in sync_hooks[sync_hook]:
+                        sync_hooks[sync_hook].append(adapter)
```
elijahbenizzy, line 761: OK, this is not ironed into the contract, and we'll likely want to switch around order, but I think for the future we need to think about this a bit -- my instinct says to do `pre_` in order and `post_` in reverse order, but that's complicated and ambiguous to implement.

hamilton/plugins/h_ddog.py
```diff
@@ -0,0 +1,259 @@
+from typing import Any, Dict, Optional
+
+from ddtrace import Span, context, tracer
+
+from hamilton import lifecycle
+
+
+class DDOGTracer(
```
elijahbenizzy, line 8: Could add testing but I don't think its worth it, I'll just fix forward and if its a pain add testing

hamilton/lifecycle/default.py
```diff
@@ -307,3 +308,33 @@ def run_after_node_execution(
...
+
+
+def wait_random(mean: float, stddev: float):
+    sleep_time = random.gauss(mu=mean, sigma=stddev)
+    if sleep_time < 0:
+        sleep_time = 0
+    time.sleep(sleep_time)
+
+
+class SlowDownYouMoveTooFast(NodeExecutionHook):
```
elijahbenizzy, line 320: Easter egg

hamilton/lifecycle/base.py
skrawcz: yes we should do it that way ideally

hamilton/plugins/h_ddog.py
```diff
@@ -1,78 +1,192 @@
-from typing import Any, Dict, Optional, Tuple
+from typing import Any, Dict, Optional
 
-from ddtrace import tracer
+from ddtrace import Span, context, tracer
```
skrawcz: surround with try/except and print what to install to use this ?

hamilton/plugins/h_ddog.py
```diff
@@ -1,78 +1,192 @@
...
-        )  # Cache of (run_id, task_id, node_id) tuples -- carries multiple levels
+        )  # Cache of run_id -> [task_id, node_id] -> span. We use this to open/close general traces
 
     @staticmethod
-    def _span_key(
-        run_id: str, task_id: Optional[str], node_id: Optional[str]
-    ) -> Tuple[Optional[str], ...]:
-        return run_id, task_id, node_id
+    def _serialize_span_dict(span_dict: Dict[str, Span]):
+        """Serializes to a readable format. We're not propogating links, but that's fine (for now).
```
skrawcz, line 49: what do you mean link?

skrawcz (approved):

hamilton/plugins/h_ddog.py
elijahbenizzy: Span links -- see notes elsewhere

PR #648, opened 2024-01-18
type annotations edited to fix linting issue

zilto:
Pylance (VSCode linter) would underline the entirety of materializers. 

## Changes
Elijah found the bug and asked to edit the type annotations of two methods

## How I tested this
- Pylance no longer throws errors after local changes. 
- Reran the test suite (except `tests/integrations` and `tests/plugins`) successfully and

### DIFF
--- a/hamilton/io/materialization.py
+++ b/hamilton/io/materialization.py
@@ -26,7 +26,7 @@ class materialization_meta__(type):
         clsobj.__annotations__ = {}
         return clsobj
 
-    def __getattr__(cls, item: str) -> "MaterializerFactory":
+    def __getattr__(cls, item: str) -> "_MaterializerFactoryProtocol":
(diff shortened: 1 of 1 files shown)


hamilton/io/materialization.py
```diff
@@ -26,7 +26,7 @@ def __new__(cls, name, bases, clsdict):
         clsobj.__annotations__ = {}
         return clsobj
 
-    def __getattr__(cls, item: str) -> "MaterializerFactory":
```
elijahbenizzy, line 29: We need to move these classes around too -- I can do that if you're slammed

elijahbenizzy: Hmm I think we don't -- this is now delayed in evaluation. So long as this doesn't break/complain if you put `from __future__ import annotations` then it should be fine!

PR #649, opened 2024-01-18
Infer visualization format from output path

zilto:
Previously, visualizations functions needed an argument `output_file_path="/path/to/my_viz"` and the format was specified via `viz_func(render_kwargs=dict(format="png"))`. 

## Motivation
- for the path `/path/to/my_viz``, `my_viz` actually specifies the file name even though it doesn't have an extension. Graphviz render engine would append the format to the file name. It leads to unintuitive behavior where `/path/to/my_viz.png` would get appended another format `/path/to/my_viz.png.pdf`
- the default is `pdf` which unlikely to be the most used format
- you really need to dig around to figure out how to specify `render_kwargs=dict(format="png"))` and it decreases readability of visualization function calls.

## Changes
- added a default `kwargs` value to have `png` as the default format instead of `pdf`
- added a step to infer the file format from the `output_file_path`. This will overwrite the default `png`. Then, the `render_kwargs=dict(format=...))` value will overwrite the value again if specified (backward compatibility)
- if a file format is inferred (e.g., `svg` for `/path/to/my_viz.svg`), the `.svg` suffix is removed from the `output_file_path` to allow the graphviz render engine to add it back (prevent `/path/to/my_viz.svg.svg`)

## How I tested this
- needed to change a few tests that inspected the DOT file from `dag.DOT` to `dag`
- ran the test suite successfully
- I manually tested using the default png, the inferred, and the explicit formats, but didn't add tests for it

## Checklist

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -6,6 +6,7 @@ It should only house the graph & things required to create and traverse one.
 Note: one should largely consider the code in this module to be "private".
 """
 import logging
+import pathlib
 import uuid
 from enum import Enum
 from types import ModuleType
@@ -787,8 +788,14 @@ class FunctionGraph:
             deduplicate_inputs,
             display_fields=display_fields,
         )
-        kwargs = {"view": True}
-        if render_kwargs and isinstance(render_kwargs, dict):
+        kwargs = {"view": False, "format": "png"}  # default format = png
+        if output_file_path:  # infer format from path
+            if suffix := pathlib.Path(output_file_path).suffix:
+                inferred_format = suffix.partition(".")[-1]
+                kwargs.update(format=inferred_format)
+                # remove suffix if exist because dot.render() will append the format
+                output_file_path = str(pathlib.Path(output_file_path).stem)
+        if render_kwargs and isinstance(render_kwargs, dict):  # accept explicit format
             kwargs.update(render_kwargs)
         if output_file_path:
             dot.render(output_file_path, **kwargs)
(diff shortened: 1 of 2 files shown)


hamilton/graph.py
```diff
@@ -787,8 +788,14 @@ def display(
             deduplicate_inputs,
             display_fields=display_fields,
         )
-        kwargs = {"view": True}
-        if render_kwargs and isinstance(render_kwargs, dict):
+        kwargs = {"view": False, "format": "png"}  # default format = png
+        if output_file_path:  # infer format from path
+            if suffix := pathlib.Path(output_file_path).suffix:
```
elijahbenizzy, line 793: IMO you should use `is not None` for clarity (though I have a thing against truthiness in python...)

elijahbenizzy (commented):
Think this is good but I haven't thought through the edge cases. Is this backwards compatible? 

Also, add to docstring -- will help clarify as well.

hamilton/graph.py
zilto: The changes in behavior:
1. PNG is the default instead of PDF
2. a non-default format (previously PDF) can be produced without specifying it in `render_kwargs`
3. the format is inferred from "the rightmost `.{string}`

Regarding backwards compatibility and #3, the main foreseeable issue is people who were specifying `dr.display_all_functions("/path/to/file.dot")`, which would generate `file` and `file.dot` that both contain the Graphviz specs.

elijahbenizzy (approved):
This looks good to me, I think its backwards compatible enough, and a bit QOL improvement.

PR #654, opened 2024-01-22
Update decorators.md

elijahbenizzy:
Fixes some docs

## Checklist


skrawcz (commented):
we should really remove decorators.md...

Otherwise the code snippet is fine - why are you removing it?
I would show both ways.

PR #658, opened 2024-01-24
Experiment manager

zilto:
This introduces a small utility server to track local Hamilton runs and launch a UI to view runs metadata and local artifacts.

# ⛔IMPORTANT: Refer to the README for up-to-date instructions to run the hook and experiment server
## How to use it
To track your run metadata, add to your driver the `ExperimentTracker()` adapter and specify an absolute path to the cache `cache_path` and an absolute or relative path to the runs base directory `run_dir`. You will be able to view the artifacts that you materialize

```python
dr = (
  driver.Builder()
  .with_modules(...)
  .with_adapter(
    ExperimentTracker(
      base_directory="./experiments",  # you need to have created this base directory yourself
    ),
  )
  .build()
)

materializers = [...]
dr.visualize_materialization(*materializers, output_file_path="dag")
dr.materialize(*materializers)
```

Then, `uvicorn server:app --reload` will start a FastAPI server. It mounts the `run_dir` that was specified in the `ExperimentTracker` instantiation and serves these local files for DAG viz and metadata. The repository cache is read in-memory to gather metadata.

## Challenges
- the hooks cannot produce the DAG viz
- arbitrary inputs, overrides, and config types are not JSON serializable
- manage nested objects in-process, as JSON, as Pydantic models

## What doesn't work or needs to be polished
- Inspired by [hydra](https://hydra.cc/docs/intro/), I wanted to have each call to the driver to generate new base directory (see hook `__init__()`). Way harder to implement in practice. If I move the current directory, I break all relative paths such as `pd.read_csv(...)` inside a node. I believe Hydra makes this possible by wrapping the `main()` entrypoint into a `Task` abstraction.
- FastUI Table component is unlikely to scale for large dataframe because you need each row to be instantiated as a Pydantic model. Probably need to do manual partitioning and paging of the Table component
- Serialize datetime objects to JSON (didn't bother yet, Pydantic could make it easy)
- Would be better to deduplicate Pydantic and dataclass models, but didn't want the hook to depend on Pydantic

## How I tested this
- no tests

## Notes
- edit the values for `CACHE_PATH` in `runner.py` and `server.py`


skrawcz:
Thoughts:

1. Experiment should map to a logical concept of what a user is doing. i.e. hyper parameter tuning, or logistic regression model. 
2. Under an experiment then there should be runs. Runs could correspond to just a single run, or multiple runs. 

So directory structure should be: `base/{user_provide-name or UUID()}/{run UUID created from hash of graph & inputs or random UUID}`.

The experiment UUID could be configurable:
 - user provided
 - random UUID
 - based on the graph hash

The run UUID could be configurable for the experiment:
 - e.g. based on graph  & inputs, or just inputs, or random each time.

That way a user could run things like in MLFLow, or use the graph structure, and therefore decide to turn on logic to skip execution (e.g. throw an error) or overwrite - or re-run...

elijahbenizzy:
Some other notes:
1. I'm going to merge [this](https://github.com/DAGWorks-Inc/hamilton/pull/662) -- it'll make it so that the `file` works all the time
2. Something funky around json/dict -- have you updated pandas recently?


Re what @skrawcz said, I think we need pages for:

1. Experiments
2. Runs (for an experiment)
3. Data (for a run)

Also, we should have some features:
1. Truncate dataframe
2. Display experiment name in header
3. Get date to work (you'll need a custom json serializer probably, pydantic should help here?)
4. If you can, copy the code/environment variables for a driver instantiation (you can just get the script that's running) -- this would be nice to display.
5.  The inputs/configs/overrides are displayed as a list -- can we display it as a dict? Or is pydantic not happy?

zilto:
Please see the README for relevant information. The README is not final, but wanted to share to have feedback

## Notes
- This is packaged as an installable Python package shipped under hamitlon/examples/experiment_management.
- It's inconvenient to pass variable to a FastAPI server via command line, so argparse in `__main__` is used to set the environment variable "HAMILTON_EXPERIMENTS_PATH"

zilto:
TODO:
- add support for loading figures from to.plt and to.plotly

examples/experiment_management/experiments/.gitinclude
skrawcz, line 1: what is this for?

examples/experiment_management/pyproject.toml
```diff
@@ -15,5 +15,8 @@ dependencies = [
     "uvicorn",
 ]
 
+[project.scripts]
+hexperiments = "hamilton_experiments.__main__:main"
```
skrawcz, line 19: this is fine. If we create a fully fledged hamilton CLI then we can always migrate this to it.

skrawcz (commented):
so move the plugins, and then add the script to hamilton setup.py.

We want people to import something like:
```python
from hamilton.plugins import h_experiments

t = h_experiments.ExperimentTracker(...)
```

zilto:
Sorry commit history is messy because of `git rebase main` :/

skrawcz:
> Sorry commit history is messy because of `git rebase main` :/ All of them between [f19fa0f](https://github.com/DAGWorks-Inc/hamilton/pull/658/commits/f19fa0fb144211dc6f1941a373b3b7cf4aab8988) and [182f8e2](https://github.com/DAGWorks-Inc/hamilton/pull/658/commits/182f8e25b778d3226f312c6b205a9a5d8b640c6a) are irrelevant/duplicates

yeah I think next time I would:
 - squash to one commit
 - rebase
 - then redo the commits

Not sure why they're duplicates 🤔

skrawcz:
I think we want to merge https://github.com/DAGWorks-Inc/hamilton/pull/680 instead of this one? We should port the PR message, etc.

PR #659, opened 2024-01-24
Docs: updated node page

zilto:
I updated the node page

## Changes
- added 2 viz
- incorporated info from a "best practices" page
- detailed the benefits of Hamilton vs. other options

## How I tested this
- built the docs locally

## Notes
- Next steps: Driver/Execution Basics, Node Advanced (decorator, parallel, config, tag, etc.)


docs/concepts/node.rst
```diff
@@ -0,0 +1,90 @@
+====
+Node
```
skrawcz, line 2: Functions and Nodes

docs/concepts/node.rst
```diff
@@ -0,0 +1,90 @@
...
+====
+
+The Node is the basic unit of Hamilton. Each node contains code to execute, and they can be assembled to form a dataflow.
+
+Nodes are simply Python functions
+---------------------------------
+Nodes are defined using the native Python function definition API. The figure below shows how the functions ``A()``, ``B()``, and ``C()`` are automatically assembled into a dataflow. Notice that function definitions don't have any special "Hamilton code". Also, the function name and return type ``A() -> int`` match the parameter ``A: int`` found in functions ``B()`` and ``C()``.
```
skrawcz, line 9: I don't think this is the way to introduce "node". We start on all the other pages talking about functions, so we should start with that here and transition to node -- or do we even need to talk about nodes?

docs/concepts/node.rst
```diff
@@ -0,0 +1,90 @@
...
+        """Divide A by 3"""
+        b = A / 3
+        return _round_three_decimals(b)
+
+
+Function naming
+---------------
+Hamilton strongly agrees with the `Zen of Python <https://peps.python.org/pep-0020/>`_ #2: "Explicit is better than implicit". Meaningful function names help document what functions do, so don't shy away from longer names. If you were to come across a function named ``life_time_value`` versus ``ltv`` versus ``l_t_v``, which one is most obvious? Remember your code usually lives a lot longer that you ever think it will.
+
+Generally, it is good practice to include meaningful verbs in function names (e.g., ``get_credentials()``, ``statistical_test()``). With Hamilton, the function name determines the node name and how data will be queried. Therefore, names that describe the node result rather than its action may be more readable (e.g., ``credentials()``, ``statistical_results()``).
```
skrawcz, line 75:
```suggestion
Unlike the common practice to include meaningful verbs in function names (e.g., ``get_credentials()``, ``statistical_test()``), with Hamilton, the function name should more closely align with nouns. That's because the function name, determines the node name and how data will be queried. Therefore, names that describe the node result rather than its action may be more readable (e.g., ``credentials()``, ``statistical_results()``).
```

docs/concepts/best-practices/function-naming.rst
```diff
@@ -1,51 +0,0 @@
-===============
```
skrawcz, line 1: I feel like a bunch of this content was not incorporated in the new page?

zilto: I should actually revert deletion. 

The goal would be to remove the `best-practices` section and have all the info under either `Concepts` or `How to`.

What wasn't included in the `Node` page would probably fit best under a `Hamilton for teams` page under `How to`

zilto:
I wonder if the `Function naming` and the `Benefits of Hamilton` best fit here. Following the diataxis [Tutorial](https://diataxis.fr/tutorials/) approach, explanations should be minimized and these two sections don't help a user get started. 

I think the sections are well-written and relevant, but maybe best suited elsewhere

docs/concepts/node.rst
zilto: it's true that the welcome page should be adjusted accordingly

hamilton/version.py
```diff
@@ -1 +1 @@
-VERSION = (1, 46, 0)
+VERSION = (1, 45, 0)
```
skrawcz, line 1: remove this change.

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
+====
+Node
+====
+
+The Node is the basic unit of Hamilton. Each node contains code to execute, and they can be assembled to form a dataflow.
```
skrawcz, line 5:
```suggestion
You as a user write functions, that Hamilton ingests to create nodes in a dataflow, a.k.a. directed acyclic graph. Nodes are the unit of work that Hamilton thinks and operates in.  So it is common to for users to interchangeably use "function" and "node" as synonyms for each other, since they have a duality. However, you as a user never directly create "nodes" in Hamilton. In this section we'll describe how functions become nodes, that then form a dataflow. 
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Nodes are simply Python functions
```
skrawcz, line 7:
```suggestion
How functions become nodes
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+  However, Hamilton allows to create multiple nodes with a single function to keep your code `DRY <https://en.wikipedia.org/wiki/Don't_repeat_yourself>`_. We'll discuss these features later in this guide.
+
+Function naming
+---------------
+Hamilton strongly agrees with the `Zen of Python <https://peps.python.org/pep-0020/>`_ #2: "Explicit is better than implicit". Meaningful function names help document what functions do, so don't shy away from longer names. If you were to come across a function named ``life_time_value`` versus ``ltv`` versus ``l_t_v``, which one is most obvious? Remember your code usually lives a lot longer that you ever think it will.
+
+Unlike the common practice of including meaningful verbs in function names (e.g., ``get_credentials()``, ``statistical_test()``), with Hamilton, the function name should more closely align with nouns. That's because the function name determines the node name and how data will be queried. Therefore, names that describe the node result rather than its action may be more readable (e.g., ``credentials()``, ``statistical_results()``).
+
+
+Node definitions -> Dataflow definition
```
skrawcz, line 56:
```suggestion
Function definitions -> Node definitions -> Dataflow definition
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Nodes are simply Python functions
+---------------------------------
+Nodes are defined using the native Python function definition API. The figure below shows how the functions ``A()``, ``B()``, and ``C()`` are automatically assembled into a dataflow. Notice that function definitions don't have any special "Hamilton code". Also, the function name and return type ``A() -> int`` match the parameter ``A: int`` found in functions ``B()`` and ``C()``.
+
+.. image:: ../_static/abc.png
+    :scale: 50
+    :align: center
+
+Anatomy of a node
```
skrawcz, line 15:
```suggestion
Anatomy of a function/node
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+     - Node dependencies
+   * - 3
+     - Docstring
+     - Description of the node return value
+   * - 4
+     - Function body
+     - Implementation of the node
+
+.. note::
+  A `node` is an object Hamilton creates and uses as part of a `dataflow`. A `function` is a piece of Python code wrote by someone. A `function` almost always map to a single `node` so the two terms are often used interchangeably.
```
skrawcz, line 45:
```suggestion
Given Hamilton `functions`, and you'll get `nodes`, that form a `dataflow`. A `function` almost always map to a single `node` so the two terms are often used interchangeably, however there are ways to augment `functions` to create multiple `nodes` in dataflow -- usually via decorators (TODO link). 
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+
+  However, Hamilton allows to create multiple nodes with a single function to keep your code `DRY <https://en.wikipedia.org/wiki/Don't_repeat_yourself>`_. We'll discuss these features later in this guide.
```
skrawcz, line 47:
```suggestion
Note: often the reason for having a function create multiple nodes is to keep your code `DRY <https://en.wikipedia.org/wiki/Don't_repeat_yourself>`_. We'll discuss these features later in this guide.
```

docs/concepts/node.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Unlike the common practice of including meaningful verbs in function names (e.g., ``get_credentials()``, ``statistical_test()``), with Hamilton, the function name should more closely align with nouns. That's because the function name determines the node name and how data will be queried. Therefore, names that describe the node result rather than its action may be more readable (e.g., ``credentials()``, ``statistical_results()``).
+
+
+Node definitions -> Dataflow definition
+---------------------------------------
+Defining an Hamilton node requires specifying it's dependencies as function parameters. By walking through dependencies, Hamilton can automatically build a dataflow from a collection of node definitions. In practice, Hamilton loads all functions of a Python modules, so you don't have to import them one by one.
```
skrawcz, line 58:
```suggestion
A function's parameters declare dependencies, which in turn form edges between nodes, to link a dataflow together. So the steps to create a dataflow are:

1. Provide modules that contain functions for Hamilton to parse.
2. For each function, Hamilton creates a node(s).
3. Based on each function's parameters, edges are created between Hamilton nodes.
4. At the end of this process we have a dataflow defined.

Note: Hamilton automatically crawls modules provided for functions so you don't have to import them one by one or explicitly define what the dataflow is, because you've already defined it via function naming and parameters.
```

hamilton/version.py
zilto: aren't we on 1.46.0 ?

skrawcz: yep we're on 1.46.0 this was reverting that.

skrawcz: so remove this too.

skrawcz (approved):
Cool. The dynamic DAGs section I assume will be resurrected in the next part of this?

PR #663, opened 2024-01-26
Custom visualization styling

zilto:
Added a user-facing interface to customize visualization styling

## Changes
- changes to visualization functions on the `Driver` class
- changes to the visualization algorithm in the `graph.py` module
- added an example under `examples/styling-visualization`

## How I tested this
- added 3 tests
  - custom style correctly applies only to tagged nodes
  - custom style appears in node
  - custom style appears in legend
- all tests ran successfully

### Style your dataflow visualizations
A core feature of Hamilton is the ability to generate visualizations directly from your code. By default, each nodes is labelled with its name and type and is stylized (shape, color, outline, etc.).

Now, it is possible to customize the visualization style based on node attributes. To do so, you need to define a function that will be applied to each node of the graph.

```python
def custom_style(*, node, node_class):
    """Custom style function for the visualization."""
    if node.tags.get("PII"):
        style = ({"fillcolor": "aquamarine"}, node_class, "PII")

    elif node.tags.get("owner") == "data-science":
        style = ({"fillcolor": "olive"}, node_class, "data-science")

    elif node.tags.get("owner") == "data-engineering":
        style = ({"fillcolor": "lightsalmon"}, node_class, "data-engineering")

    else:
        style = ({}, node_class, None)

    return style
``` 

### Steps to define your custom style

1. The function must used only keyword arguments, taking in `node` and `node_class`.
2. It needs to return a tuple `(style, node_class, legend_name)` where:
    - `style`: dictionary of valid [graphviz node style attributes](https://graphviz.org/doc/info/attrs.html).
    - `node_class`: class used to style the default visualization - we recommend return the input `node_class`
    - `legend_name`: text to display in the legend. Return `None` for no legend entry.
3. For the execution-focused visualizations, your custom styles is applied before the modifiers for outputs and overrides are applied.

If you need more customization, we suggest getting the graphviz object back, and then modifying it yourself.

This [online graphviz editor](https://edotor.net/) can help you get started!

dag

## Checklist


examples/styling_visualization/lineage_commands.py
```diff
@@ -0,0 +1,253 @@
+"""
```
skrawcz, line 1: did you mean to include this file?

hamilton/driver.py
```diff
@@ -755,6 +756,7 @@ def display_all_functions(
             Can improve readability depending on the specifics of the DAG.
         :param show_schema: If True, display the schema of the DAG if
             the nodes have schema data provided
+        :param custom_style_function: Optional. Custom style function.
```
skrawcz, line 759:
```suggestion
        :param custom_style_function: Optional. Custom style function. See example in repository for example use.
```

skrawcz (commented):
Looking good.

Just that one file.

Otherwise just need to put the content of this PR somewhere in the docs too!

examples/styling_visualization/lineage_commands.py
zilto: nope! Just copied the full lineage example. Will remove

zilto:
Will add to the docs as I rework them. I believe the feature is niche enough to live under the examples directory for now!

skrawcz (approved):
cool -- can you then delete the lineage_commands.py and this is good to go? The example shouldn't have files that aren't used.

PR #677, opened 2024-02-02
Truncates inputs logged when a node errors

skrawcz:
This was requested by a user, because we could overwhelm a log system. This should suffice for now. We can always modify tweak this later.

## Changes
- graph_functions.py error handling code

## How I tested this
 - locally

```
********************************************************************************
> spend_per_signup [my_functions.spend_per_signup()] encountered an error      <
> Node inputs:
{'signups': '0      1\n1     10\n2     50\n3    100\n4    200\n5    ...',
 'spend': '0    10\n1    10\n2    20\n3    40\n4    40\n5    50\ndt...'}
********************************************************************************
```

## Checklist

### DIFF
--- a/hamilton/execution/graph_functions.py
+++ b/hamilton/execution/graph_functions.py
@@ -1,4 +1,5 @@
 import logging
+import pprint
 from typing import Any, Collection, Dict, List, Optional, Set, Tuple
 
 from hamilton import node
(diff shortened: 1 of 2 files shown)


hamilton/execution/graph_functions.py
```diff
@@ -95,6 +96,29 @@ def combine_config_and_inputs(config: Dict[str, Any], inputs: Dict[str, Any]) ->
...
+    for k, v in kwargs.items():
+        item_repr = repr(v)
+        if len(item_repr) > 50:
+            item_repr = item_repr[:50] + "..."
+        else:
+            item_repr = v
+        inputs[k] = item_repr
+    input_string = pp.pformat(inputs)
+    if len(input_string) > 1000:
+        input_string = input_string[:80] + "..."
```
elijahbenizzy, line 118: Why 80 versus 1000? Shouldn't we truncate it to 1000?

hamilton/execution/graph_functions.py
```diff
@@ -95,6 +96,29 @@ def combine_config_and_inputs(config: Dict[str, Any], inputs: Dict[str, Any]) ->
     return {**config, **inputs}
 
 
+def create_input_string(kwargs: dict) -> str:
+    """THis is a utility function to create a string representation of the inputs to a function.
```
elijahbenizzy, line 100: This

elijahbenizzy (commented):
Looks good -- we don't have a unit tests for the overall, non-item-level truncation, but we should probably just fix that (and maybe add a test?)

hamilton/execution/graph_functions.py
skrawcz: oops, yeah missed that.

skrawcz:
> Looks good -- we don't have a unit tests for the overall, non-item-level truncation, but we should probably just fix that (and maybe add a test?)

? yes we do.

elijahbenizzy (approved):

PR #680, opened 2024-02-06
Add experiment UI and plugin

skrawcz:
This introduces a small utility server to track local Hamilton runs and launch a UI to view runs metadata and local artifacts.

# ⛔IMPORTANT: Refer to the README for up-to-date instructions to run the hook and experiment server
## How to use it
To track your run metadata, add to your driver the `ExperimentTracker()` adapter and specify an absolute path to the cache `cache_path` and an absolute or relative path to the runs base directory `run_dir`. You will be able to view the artifacts that you materialize

```python
dr = (
  driver.Builder()
  .with_modules(...)
  .with_adapter(
    ExperimentTracker(
      base_directory="./experiments",  # you need to have created this base directory yourself
    ),
  )
  .build()
)

materializers = [...]
dr.visualize_materialization(*materializers, output_file_path="dag")
dr.materialize(*materializers)
```

Then, `uvicorn server:app --reload` will start a FastAPI server. It mounts the `run_dir` that was specified in the `ExperimentTracker` instantiation and serves these local files for DAG viz and metadata. The repository cache is read in-memory to gather metadata.

## Challenges
- the hooks cannot produce the DAG viz
- arbitrary inputs, overrides, and config types are not JSON serializable
- manage nested objects in-process, as JSON, as Pydantic models

## What doesn't work or needs to be polished
- Inspired by [hydra](https://hydra.cc/docs/intro/), I wanted to have each call to the driver to generate new base directory (see hook `__init__()`). Way harder to implement in practice. If I move the current directory, I break all relative paths such as `pd.read_csv(...)` inside a node. I believe Hydra makes this possible by wrapping the `main()` entrypoint into a `Task` abstraction.
- FastUI Table component is unlikely to scale for large dataframe because you need each row to be instantiated as a Pydantic model. Probably need to do manual partitioning and paging of the Table component
- Serialize datetime objects to JSON (didn't bother yet, Pydantic could make it easy)
- Would be better to deduplicate Pydantic and dataclass models, but didn't want the hook to depend on Pydantic

## How I tested this
- no tests

## Notes
- edit the values for `CACHE_PATH` in `runner.py` and `server.py`


elijahbenizzy (approved):
Looks good once https://github.com/DAGWorks-Inc/hamilton/pull/681 is merged

PR #686, opened 2024-02-08
Fix path metadata

elijahbenizzy:
Fixes path metadata for s3/other URLs.

## Checklist

### DIFF
--- a/hamilton/io/utils.py
+++ b/hamilton/io/utils.py
@@ -1,6 +1,9 @@
 import os
+import time
 from datetime import datetime
+from pathlib import Path
 from typing import Any, Dict, Union
+from urllib import parse
 
 import pandas as pd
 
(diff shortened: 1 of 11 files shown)


hamilton/io/utils.py
```diff
@@ -21,12 +24,28 @@ def get_file_metadata(path: str) -> Dict[str, Any]:
     - the last modified time
     - the current time
     """
+    if isinstance(path, Path):
+        path = str(path)
+    parsed = parse.urlparse(path)
+    size = 0
```
skrawcz, line 30: this should be None or -1. 0 could be a real value.

tests/io/test_utils.py
```diff
@@ -16,3 +18,19 @@ def test_get_sql_metadata():
...
+    assert metadata["file_metadata"]["path"] == str(file_path)
+    assert metadata["file_metadata"]["size"] > 0
+    assert metadata["file_metadata"]["last_modified"] == file_path.stat().st_mtime
+    assert metadata["file_metadata"]["timestamp"] is not None
+
+
+def test_get_file_metadata_url_schema():
+    url = "s3://bucket/key"
+    metadata = get_file_metadata(url)
+    assert metadata["file_metadata"]["path"] == url
```
skrawcz, line 36: shouldn't this check scheme?

skrawcz (approved):
can add test that checks scheme & notes

tests/io/test_utils.py
elijahbenizzy: schema yes, not notes cause that's not really part of the API

skrawcz (approved):

PR #689, opened 2024-02-09
Adds first pass at jupyter magic

skrawcz:
To use:

```python
%load_ext hamilton.plugins.jupyter_magic

%%cell_to_module -m MODULE_NAME
def my_hamilton_funcs(): ...

```

The above directive does three things:

1. it creates a module with the contents of this cell
2. it imports the module under the name `MODULE_NAME`.
3. if changes are made to the module, it will be reloaded automatically. If you constructed a driver with this module, you won't need to re-create it (in most cases). Note, it skips the case where adapters are passed because I couldn't get it to work. We should think about what to change to make the internal APIs less dependend on here. I don't think this will support multi-processing drivers either.

## Changes
- adds to plugins

## How I tested this
- ran it locally

## Checklist


zilto:
My current API

```python
# %%cell 1
%load_ext hamilton_magic  # add Stefan's plugin import structure

# %%cell 2
%%cell_to_module -m my_cell -v True

def hello() -> str:
    return "hello"

def world(hello: str) -> str:
    return hello + " " + "world"
```

The option `-m` specifies a module name that will exist in the notebook space and `-v` is for "view" and any value will automatically call `display_all_functions()`

Code snippets:
```python
from hamilton import ad_hoc_utils, driver

def create_module(source: str) -> ModuleType:
    """Create a temporary module from source code"""
    module_name = ad_hoc_utils._generate_unique_temp_module_name()
    module_object = ModuleType(module_name, "")
    sys.modules[module_name] = module_object
    exec(source, module_object.__dict__)
    return module_object

@magics_class
class HamiltonMagics(Magics):
    """Magics to facilitate Hamilton development in Jupyter notebooks"""
    
    @magic_arguments()  # needed on top to enable parsing
    @argument("-m", "--module_name", help="Optional module name.")  # keyword / optional arg
    @argument("-v", "--view", help="Flag to visualize dataflow.")  # keyword / optional arg
    @cell_magic
    def cell_to_module(self, line, cell):
        """Compile the cell as a module and make it available to notebook"""
        self.shell.ex(cell)
        
        args = parse_argstring(self.cell_to_module, line)  # specify how to parse by passing method
        module_name = "jupyter" if args.module_name is None else args.module_name
        module_object = create_module(cell)
        self.shell.push({module_name: module_object})
        
        if args.view:
            dr = driver.Builder().with_modules(module_object).build()
            return display(dr.display_all_functions(None))
```

hamilton/plugins/jupyter_magic.py
```diff
@@ -0,0 +1,259 @@
...
+            try:
+                if args.config.startswith("'") or args.config.startswith('"'):
+                    # strip quotes if present
+                    args.config = args.config[1:-1]
+                display_config = json.loads(args.config)
+            except json.JSONDecodeError:
+                print("Failed to parse config as JSON. Please ensure it's a valid JSON string.")
+                print(args.config)
+
+        # should_rebuild_drivers = True if args.rebuild_drivers is not None else False
```
elijahbenizzy, line 209: Why is this commented out?

elijahbenizzy (approved):
looks good, let's ship it

hamilton/plugins/jupyter_magic.py
skrawcz: ah I need to remove.

zilto:
@elijahbenizzy should allow by default `.enable_dynamic_execution(allow_experimental_mode=True)` in the Builder we use behind the scene? are there downsides to having the Parallel mode on?

PR #690, opened 2024-02-11
fix: JSON materializer don't support `list[dict]`

zilto:
At the top level, JSON files are either a mapping `{ }` or a list `[ ]`. Currently, the default JSON DataSaver and DataLoader only map to a Python `dict` type and don't support a `list[dict]` annotation.

For example:
```python
JSON_FILE_CONTENT = [
    {"name": "A"},
    {"name": "B"}
]

# this works
def node(json_data: dict) -> dict:
    return json_data

# this should work but doesn't
def from_to_bug(json_data: list[dict]) -> list[dict]:
    return json_data

# this doesn't work because of `from_`
def from_bug(json_data: list[dict]) -> dict:
    return json_data

# this doesn't work because of `to`
def to_bug(json_data: dict) -> list[dict]:
    return json_data


if __name__ == "__main__":
    from hamilton import driver
    from hamilton.io.materialization import to, from_
    import __main__
    
    # comment out functions to avoid name/type conflicts
    dr.materialize(
        from_.json(target="json_data", path="data.json"),
        to.json(path="./out.json", id="out__json", dependencies=["from_to_bug"]),
        additional_vars=["from_to_bug"],
    )
```

## Changes
- Changed the type annotations for JSON `DataSaver` and `DataLoader`

## How I tested this
- the test suite runs successfully
- the above example now works

## Notes
- ❗ one issue with supporting both `dict` and `list[dict]` is that the object might be miss-typed and still load/save. For example, 

```python
JSON_FILE_CONTENT = [
    {"name": "A"},
    {"name": "B"}
]

# with the fix, the next 2 fucnctions would work
# however, `json_data` can't be `list[dict]` and `dict`

# for the above data,
# this one fails because isinstance(json_data, dict) is False
def from_bug_alert(json_data: list[dict]) -> dict:
    return json_data

# this one works / fails silently because type(json_data) in (dict, list[dict])
def from_bug_silent(json_data: dict) -> list[dict]:
    return json_data
    
# similar logic applies with `to`, but it matters less because nothing
# depends on materializer nodes
```

## Checklist

### DIFF
--- a/hamilton/io/default_data_loaders.py
+++ b/hamilton/io/default_data_loaders.py
@@ -4,7 +4,7 @@ import json
 import os
 import pathlib
 import pickle
-from typing import Any, Collection, Dict, Tuple, Type, Union
+from typing import Any, Collection, Dict, List, Tuple, Type, Union
 
 from hamilton.io.data_adapters import DataLoader, DataSaver
 from hamilton.io.utils import get_file_metadata
@@ -16,7 +16,7 @@ class JSONDataLoader(DataLoader):
 
     @classmethod
     def applicable_types(cls) -> Collection[Type]:
-        return [dict]
+        return [dict, List[dict]]
 
     def load_data(self, type_: Type) -> Tuple[dict, Dict[str, Any]]:
         with open(self.path, "r") as f:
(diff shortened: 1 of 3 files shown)


elijahbenizzy (commented):
Hmm -- maybe this should be a `.jsonl` format? Its not to spec -- while python json is happy with this, these are usually marked differently.

Fine either way, but we should add a unit test as well for this

elijahbenizzy (approved):
lgtm! There are parameterizations without more than 1, but I think that's a pattern we should just kill in the rest of it (or add more parameterizations).

PR #707, opened 2024-02-21
CLI added to enable CI of Hamilton dataflows

zilto:
The CLI is implemented using the Typer library which is based on Click and have similar structure to FastAPI.

For design:
- hamilton.cli.__main__ is the CLI app. It should only rely on hamilton.cli.commands and typer
- hamilton.cli.commands isn't concerned with I/O and primarily relies on hamilton.cli.logic
- hamilton.cli.logic implements the core of the CLI in testable chunks.

The entrypoint is `hamilton` and was added to setup.py There are currently 4 commands:
- build: build a dataflow from Python modules
- view: dr.display_all_functions() from Python modules
- version: get nodes and dataflow hashes
- diff: compare version of current dataflow vs git reference Use `hamilton --help` for help.

## How I tested this
- added some tests to core logic and cli
- multiple relevant tests are missing

## Notes
- currently bad at logging exceptions to JSON
- will add an example to `examples/`
- could add more docstrings to internal CLI logic
- write docs page

## Checklist

- [] Project documentation has been updated if adding/changing functionality.


hamilton/cli/__main__.py
```diff
@@ -0,0 +1,188 @@
...
+
+# entrypoint for `hamilton` without command
+@cli.callback()
+def main(
+    verbose: Annotated[bool, typer.Option(help="Output all intermediary commands")] = False,
+    json_out: Annotated[
+        bool, typer.Option(help="Output JSON for programmatic use (e.g., CI)")
+    ] = False,
+):
+    """Hamilton CLI"""
```
elijahbenizzy, line 40: Why the capital letters everywhere?

hamilton/cli/__main__.py
```diff
@@ -0,0 +1,188 @@
...
+@cli.command()
+def build(
+    ctx: typer.Context,
+    modules: Annotated[
+        list[Path], typer.Argument(exists=True, dir_okay=False, readable=True, resolve_path=True)
+    ],
+):
+    """Build dataflow with MODULES"""
+    try:
+        config = dict()
```
elijahbenizzy, line 54: Add a TODO for the config loading -- good to pass in a json file but not needed now

hamilton/cli/__main__.py
```diff
@@ -0,0 +1,188 @@
...
+            command="build", success=False, message={"error": str(type(e)), "details": str(e)}
+        )
+        print(response.model_dump_json())
+        raise typer.Exit(code=1)
+
+    response = Response(
+        command="build", success=True, message={"modules": [p.stem for p in modules]}
+    )
+
+    if (ctx.info_name == "build") or state.verbose:
```
elijahbenizzy, line 67: You should consider using `pprint` -- it'll format json nicely.

hamilton/cli/__main__.py
```diff
@@ -0,0 +1,188 @@
...
+    dr: Optional[driver.Driver] = None
+    dataflow_version: Optional[dict] = None
+
+
+cli = typer.Typer()
+state = CliState()
+
+
+# entrypoint for `hamilton` without command
+@cli.callback()
```
elijahbenizzy, line 33: Logging statements of what's happening might be nice -- E.G> `attempting to build`, `attempting to download prior git revision`, etc..

hamilton/cli/logic.py
```diff
@@ -0,0 +1,232 @@
...
+
+    return origins_version
+
+
+def hash_hamilton_nodes(dr: driver.Driver) -> Dict[str, str]:
+    """Hash the source code of Hamilton functions from nodes in a Driver"""
+    graph = graph_types.HamiltonGraph.from_graph(dr.graph)
+
+    nodes_version = dict()
+    for n in graph.nodes:
```
elijahbenizzy, line 67: Is this order going to be deterministic? I think dicts are, but might as well sort it as hashing is sensitive to order.

tests/cli/test_cli.py
```diff
@@ -0,0 +1,65 @@
...
+    result = runner.invoke(cli, inputs)
+    message = json.loads(result.stdout)
+
+    assert result.exit_code == 0
+    assert message["success"] is True
+    # TODO difficult to assert the output file path from test
+    # assert output_file_path.exists()
+
+
+# def test_cli_view_output_arg(tmp_path: pathlib.Path):
```
elijahbenizzy, line 39: Commented out?

tests/cli/test_cli.py
```diff
@@ -0,0 +1,65 @@
...
+    module_path = inspect.getfile(example_module)
+    inputs = ["--json-out", "view", module_path]
+    # output_file_path = tmp_path / "dag.png"  # default file name
+
+    result = runner.invoke(cli, inputs)
+    message = json.loads(result.stdout)
+
+    assert result.exit_code == 0
+    assert message["success"] is True
+    # TODO difficult to assert the output file path from test
```
elijahbenizzy, line 35: Why is this?

elijahbenizzy (commented):
Code is looking pretty clean, a few comments. Let's add a `README` on this as well, and maybe add to the docs.

hamilton/cli/__main__.py
skrawcz: CLI is an acronym :)

zilto: Command Line Interface (as opposed to GUI, UI, Terminal User Interface, etc.)

hamilton/cli/__main__.py
zilto: I was afraid that pretty printing the JSON output would create a multiline mess for those trying to capture the `stdout`. In verbose mode, you get one JSON object per line, one for each intermediary command executed

tests/cli/test_cli.py
zilto: My manual verification of the `view --output` option works. 

I believe the `typer.testing.CliRunner` is instantiated at the pytest top-level and isn't instantiated in the relative `tmp_path` that is scoped to a single test function.

Could be fixed by passing fixtures, but probably requires more state management across tests...

tests/cli/test_cli.py
zilto: see thread for Line 35

hamilton/cli/logic.py
zilto: This function is responsible for creating one hash per node, which should be independent. 

Next, `hash_dataflow` sorts the hashes and creates a single string from the list and encodes it to bytes. Not the most elegant, but it should work.

Did I missing something?

zilto:
> Let's add a `README` on this as well, and maybe add to the docs.

I was able to easily generate [docs via their typer utils](https://typer.tiangolo.com/tutorial/package/#generate-docs-with-typer-cli-optional). Will add manually to API reference in the docs, but it could be part of CI

hamilton/cli/logic.py
elijahbenizzy: Ahh got it. I'm wondering if there's a good reason to split it out? I imagine you're using both in different places. Makes sense.

skrawcz:
@zilto broken unit tests for 3.11 and 3.12 ?

hamilton/cli/logic.py
```diff
@@ -152,46 +154,46 @@ def diff_nodes_against_functions(
...
-        v1 = mapping_v1.get(node_name)
+    for node_name, v2 in reference_map.items():
+        v1 = current_map.get(node_name)
         if v1 is None:
-            v2_only.append(node_name)
+            reference_only.append(node_name)
             continue
 
-    return dict(v1_only=v1_only, v2_only=v2_only, edit=edit)
+    return dict(current_only=current_only, reference_only=reference_only, edit=edit)
```
skrawcz, line 180: do we need the `only` suffix?

skrawcz: since that's what is output to the command line right?

zilto: I think so for clarity. Thinking in terms of set, "current only" is the difference of "current" and "reference".

skrawcz:
The docs here should also have a presence in the main docs.

You could include them doing something like I did here - https://raw.githubusercontent.com/DAGWorks-Inc/hamilton/main/docs/how-tos/use-hamilton-for-lineage.md.

skrawcz (approved):

PR #709, opened 2024-02-22
Updates jupyter magic to display in databricks

skrawcz:
Databricks doesn’t automatically render. So we
have to render it ourselves. I decided on HTML
over Image for no good reason really.

This should also handle errors if graphviz isn’t installed and the visualization extra isn’t installed.

## Changes
 - jupyter_magic.py
 
## How I tested this
 - via databricks notebook
 - locally via regular notebook

## Checklist

### DIFF
--- a/hamilton/plugins/jupyter_magic.py
+++ b/hamilton/plugins/jupyter_magic.py
@@ -15,6 +15,7 @@ If you are developing on this module you'll then want to use:
 """
 
 import json
+import os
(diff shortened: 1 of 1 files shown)


elijahbenizzy (commented):
Looks good. Will it work without jupyter_magic as well? E.G. can we get the return of a graphviz object to display in a databricks notebook?

PR #729, opened 2024-03-01
`cli` added config support and validate command

zilto:
Now, all available commands accept `--config` which receives a `Path`. The path can be a `.json`, `.py`, or `.ini` file and the content is returned as a dictionary.

## Changes
- added logic to handle config path
- modified logic that iterates over nodes to handle `node.originating_functions is None` for config nodes
- updated `commands.py` to load config before building the driver
- updated commands from `__main__.py` for the CLI to handle a new option

## How I tested this
- added config of each supported format to the `examples/cli` folder and manually tried them.

## Notes
❗ I will add a `hamilton validate` command then update docs at once
- we might want to add support for JSON config from the command line, but Typer doesn't support `Union` of types for arguments. Would need to add an additional `--config-json` `-cjson` to each command and prevent conflicts with `--config`. Maybe things could be renamed to `--config-path` and `--config` when received as JSON string
- we could accept argument overload to CLI commands  such as `hamilton build key1="true" key2=0 MODULES`, but it would be messier to parse and maintain

## Checklist

----

| <a href="https://ellipsis.dev" target="_blank">Ellipsis</a> | :rocket: This PR description was created by [Ellipsis](https://www.ellipsis.dev) for commit 692fb0d2981c9de976c379ad3a37ceed0109f39d.  | 
|--------|--------|

### Summary:
This PR enhances the CLI by adding support for configuration files, handling a new case for config nodes, updating the driver building process, handling a new CLI option, improving error handling, adding a `validate` command, and providing a demonstration of the new configuration options.

**Key points**:
- Added support for configuration files in the CLI
- Handled a new case where `node.originating_functions` is `None` for config nodes
- Updated `commands.py` to load the config before building the driver
- Updated commands from `__main__.py` for the CLI to handle a new option
- Improved error handling
- Added a `validate` command to the CLI
- Demonstrated the new configuration options in `module_v1.py` in the `examples/cli` directory

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)


hamilton/cli/logic.py
```diff
@@ -1,15 +1,15 @@
...
-import subprocess
 from pathlib import Path
 from types import ModuleType
 from typing import Dict, List, Union
 
-from hamilton import ad_hoc_utils, driver, graph, graph_types, graph_utils
+from hamilton import driver
 
 
+V = "HENLO"
```
skrawcz: ?

hamilton/cli/logic.py
```diff
@@ -22,12 +22,14 @@ def get_git_base_directory() -> str:
             return result.stdout.strip()
         else:
             print("Error:", result.stderr.strip())
+            raise Exception
```
skrawcz, line 29: Usually you raise an error with a message.

E.g. `ValueError(...)` etc

hamilton/cli/logic.py
```diff
@@ -50,6 +52,8 @@ def get_git_reference(git_relative_path: Union[str, Path], git_reference: str) -
 
 def version_hamilton_functions(module: ModuleType) -> Dict[str, str]:
     """Hash the source code of Hamilton functions from a module"""
+    from hamilton import graph_utils
```
skrawcz, line 60: why the inline import?

hamilton/cli/logic.py
```diff
@@ -1,15 +1,15 @@
-import functools
-import hashlib
-import subprocess
 from pathlib import Path
 from types import ModuleType
 from typing import Dict, List, Union
 
-from hamilton import ad_hoc_utils, driver, graph, graph_types, graph_utils
```
skrawcz, line 8: what was wrong with importing it all here?

hamilton/cli/logic.py
zilto: it reduces the CLI startup time. For one-of commands it doesn't matter as much, but when using the Watcher, it can be pretty annoying.

For the back log, I want to profile the CLI and see how I can make things faster

elijahbenizzy (commented):
Inline imports are generally bad practice -- let's remove them?

Also, can we add config stuff to the README?

zilto:
> Also, can we add config stuff to the README?

I think the CLI is still super experimental. As I work with pre-commit and GitHub actions, new API needs appear and things change frequently. 

We should strongly emphasize that user rely on `hamilton --help` (after all that's why we picked Typer).

Generally speaking, we should probably start adding dates to `examples/` to communicate potential deprecation until we automate CI for them. (hopefully the CLI will help)

elijahbenizzy (approved):
Looks good

Issue #728, opened 2024-03-01
Standardize API for dataflow and node versioning


zilto:
Currently, there are at least 3 sets of features that rely on dataflow and node versioning:
- the CLI `version` command and its underlying logic ([see here](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/cli/logic.py#L62)) which outputs a dictionary / JSON of nodes and dataflow versions. Node versioning is based on the hash of the originating function using `graph_utils.hash_source_code()`. The dataflow version is derived from the dictionary of `{node_name: node_version}`.
- the `h_diskcache.DiskCacheAdapter` also uses `graph_utils.hash_source_code()` to version the node's originating function. The cache key is derived from `(node_hash, input_1, input_2, ..., input_n)`
- the `h_experiments.hook.ExperimentTracker` preceeded the two other features. Therefore, it has its own `h_experiments.hook.graph_hash()` function. It creates a hash from the dictionary of node originating functions. It should leverage the same logic as the CLI.

# Solution design
- does it make sense to have `hash_source_code()` under `graph_utils`? If yes, we could add `hash_node()` and `hash_dataflow()`
- it could make sense as properties for `graph_types.HamiltonNode` and `graph_types.HamiltonGraph`
- disambigue terms around "hash", "version", "node", "function", "originating_function"

elijahbenizzy:
I don't think people should be using `graph_utils` as part of the API -- I like having properties on the `graph_types` to compute these in a "blessed" way.

Issue #175, opened 2023-05-23
Versioning of features / nodes


tim-habitat:
**Is your feature request related to a problem? Please describe.**
I'd love to know what version a feature was created with. I could just use a git commit hash, but it may be that between 2 commits, the code though which the feature flows hasn't changed so the version should be the same. 
This would be ideal in situation where we would need to pre-process a large amount of feature to be used later, than some process could just check whether the features were created with the same "version". If not, recompute only those ones.

**Describe the solution you'd like**
Each features / last node of a graph has a version attached to it. The version depends on the code within the function used to create the feature but also of all the parents. 
I feel in some ways i'd like the solution to be clear / easy. Like a hash of the code of the node.

i feel it might be tricky to think about for nodes that interact with `io` but perhaps that should be dealt by the user. 

**Describe alternatives you've considered**
I considered using git commit hash, but features code changes much less frequently than the overall repo. I considered implement something hacky myself but haven't got to it.

**Additional context**
![Screenshot 2023-05-23 at 15 03 52](https://github.com/DAGWorks-Inc/hamilton/assets/86600518/d73f649e-5aec-4f8e-8c2c-9ae1e3e6091b)

skrawcz:
@tim-habitat thanks for the idea! @elijahbenizzy and I have talked about this, as this is very close in functionality to "caching"/"check pointing". Would love to chat more about your use case. Would you be up for a call? Feel free to [jump into slack](https://join.slack.com/t/hamilton-opensource/shared_invite/zt-1bjs72asx-wcUTgH7q7QX1igiQ5bbdcg) and we can schedule a time there.

zilto:
some features are now available under the `h_diskcache` plugin and the `hamilton` CLI. Continuing discussion in #728

PR #734, opened 2024-03-04
Unify the node and dataflow versioning API

zilto:
Motivation described in #728 and relates to #175.  

## Changes
- moved the `hash_source_code()` and the utility `remove_docs_and_comments()` from `grap_utils` to `graph_types`
- added to `graph_types.HamiltonNode` and `graph_types.HamiltonGraph` a `.version` property
- updated code that depended on these functions and the related tests
- updated the `hamilton.plugins.h_experiments.hook.ExperimentTracker` to use the `HamiltonGraph.version` implementation to align with CLI and DiskCache 

## How I tested this
- ran the full test suite successfully after edits

## Notes
- some code path use the `hash_source_code()` directly on functions/Callables or on `hamilton.node.Node`

## Checklist


hamilton/graph_types.py
```diff
@@ -73,6 +140,15 @@ def from_node(n: node.Node) -> "HamiltonNode":
...
         )
 
+    @property
+    def version(self) -> str:
+        """Generate a hash of the node originating function source code.
+
+        The option `strip=True` means docstring and comments are ignored
+        when hashing the function.
+        """
+        return hash_source_code(self.originating_functions[0], strip=True)
```
elijahbenizzy, line 150: Ditto with the comment above

tests/test_driver_export.py
```diff
@@ -18,6 +19,7 @@
                 "    return sum_step_squared_plus_step_cubed\n"
             ),
             "documentation": "",
+            "version": None,
         },
```
elijahbenizzy, line 23: Why is `version` `None` here? Add a comment.

tests/test_graph_types.py
```diff
@@ -1,11 +1,80 @@
...
 
 from tests import nodes as test_nodes
 
 
+@pytest.fixture()
+def func_a():
+    """Default function implementation"""
+
+    return textwrap.dedent(
+        """
```
elijahbenizzy, line 18: Nit -- should be able to have the `def` on the same line as `"""`

tests/test_graph_types.py
```diff
@@ -1,11 +1,80 @@
...
+    """
+    )
+
+
+@pytest.fixture()
+def func_a_docstring():
+    """The function A() has a docstring"""
+
+    return textwrap.dedent(
+        '''
```
elijahbenizzy, line 42: nit -- quote styles should match the above

tests/test_graph_types.py
```diff
@@ -17,6 +86,7 @@ def node_to_create(required_dep: int, optional_dep: int = 1) -> str:
         tags={"tag_key": "tag_value"}, originating_functions=(node_to_create,)
     )
     hamilton_node = graph_types.HamiltonNode.from_node(n)
+    print(hamilton_node.version)
```
elijahbenizzy, line 89: `println`

hamilton/graph_types.py
```diff
@@ -16,6 +18,70 @@
...
+    The use of `ast.unparse()` requires Python 3.9
+
+    1. Parsing then unparsing the AST of the source code will
+    create a code object and convert it back to a string. In the
+    process, comments are stripped.
+
+    2. walk the AST to check if first element after `def` is a
+    docstring. If so, edit AST to skip the docstring
+
+    NOTE. The ast parsing will fail if `source` has syntax errors. For the
```
elijahbenizzy, line 33: Agreed, this is on them

elijahbenizzy (approved):
A few thoughts, but overall good

tests/test_graph_types.py
zilto: Triple quotes occupy 3 spaces while indentation is 4 spaces. Having `def` and `"""` hinders readability IMO even though `textwrap.dedent` makes everything safer.

I adjusted the alignment to avoid having the string under the `return` statement though

tests/test_graph_types.py
zilto: Ohh, it's because I have the docstring within the triple-quoted string. I used `'''` for the docstring instead.

I was wondering why

tests/test_graph_types.py
```diff
@@ -38,14 +109,76 @@ def node_to_create(required_dep: int, optional_dep: int = 1) -> str:
...
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires Python 3.9 or higher")
+def test_remove_comment(func_a: str, func_a_comment: str):
+    func_a_no_whitespace = func_a.strip()
+    stripped = graph_types._remove_docs_and_comments(func_a_comment)
+    assert func_a_no_whitespace == stripped
+
+
+@pytest.mark.skipif(sys.version_info < (3, 9), reason="requires Python 3.9 or higher")
+@pytest.mark.parametrize("strip", [True, False])
+def test_different_hash_function_body(func_a: str, func_a_body: str, strip: bool):
```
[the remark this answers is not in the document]
zilto: Doesn't it make sense to verify the intended behavior, i.e., "different func lead to different hashes", whether docstrings/comments are stripped or not. Then, other tests check conditions under which matching or non-matching hashes are expected depending on `strip=True`.

Maybe I'm missing the most basic case `hash(A) = hash(A)`, which I took for granted? Maybe it's the name of the tests? I could rename `test_different_hash_docstring` to `test_matching_hash_docstring` etc. ?

PR #744, opened 2024-03-06
Adds target_ parameter to save_to

elijahbenizzy:
This allows us to layer materialization decorators. Previously it would use the output of the other node, and thus try to save the result of the other saver. This allows them both to specify.

## Checklist


skrawcz:
unit test?

elijahbenizzy:
> unit test?

Updated, had forgot to push

skrawcz (approved):
Looks good. But seems like docs are failing.

PR #757, opened 2024-03-12
Adds FunctionInputOutputTypeChecker

skrawcz:
This is a lifecycle adapter that one can add to strictly enforce the types of things at runtime. It will check the input to and the output from a function and check that it matches the expected type annotation.

It modifies/fixes the pandas split-apply-combine example to conform to the expected types; it actually found mismatches between what was typed and what was passed in.

I had to modify graph_functions.py so that the error message from the pre-node-execute hook would show up in the right place and the post-node-execute hook would not run.

Note: this does not do anything smart with Pandera or Annotated types just yet, but it could in future.

Test generated with the help of co-pilot. A little ugly, but they do the job.

## Changes
- split-apply-combine example
- graph_functions.py -- to handle error case of pre-node-execute and post-node-execute and creating an error message
- adds FunctionInputOutputTypeChecker to lifecycle.default  & check type function

## How I tested this
- ran it locally

## Checklist


hamilton/execution/graph_functions.py
```diff
@@ -200,41 +232,29 @@ def dfs_traverse(
...
-                        "post_node_execute",
-                        run_id=run_id,
-                        node_=node_,
-                        kwargs=kwargs,
-                        success=success,
-                        error=error,
-                        result=result,
-                        task_id=task_id,
-                    )
+                if not pre_node_execute_errored and adapter.does_hook(
```
elijahbenizzy, line 240: Hmm we should really have a context manager that does this... But this works for now.

hamilton/lifecycle/api.py
```diff
@@ -206,6 +208,7 @@ def pre_node_execute(
             node_return_type=node_.type,
             task_id=task_id,
             run_id=run_id,
+            node_input_types={k: v[0] for k, v in node_.input_types.items()},
```
elijahbenizzy, line 211: Add TODO for integrating the `HamiltonNode` class

hamilton/lifecycle/default.py
```diff
@@ -4,9 +4,15 @@
 import pdb
 import pprint
 import random
+import sys
 import time
 from typing import Any, Callable, Dict, List, Optional, Union
 
+if sys.version_info >= (3, 9):
+    from typing import Literal
```
elijahbenizzy, line 12: Think literal is `3.8+`?

hamilton/lifecycle/default.py
```diff
@@ -340,3 +346,118 @@ def run_before_node_execution(self, **future_kwargs: Any):
     def run_after_node_execution(self, **future_kwargs: Any):
         """Does nothing"""
         pass
+
+
+def check_instance(obj: Any, type_: Any) -> bool:
```
elijahbenizzy, line 351: Why are we not using this? https://github.com/DAGWorks-Inc/hamilton/blob/6bb3c6955ed35735cde4e2c9c67ef4019070151d/hamilton/htypes.py#L286

elijahbenizzy (approved):
As discussed, looks good when you make the changes

PR #765, opened 2024-03-18
`fix` make visualization nodes `filled` by default

zilto:
Previously, certain nodes types didn't have a `style="filled"`. 

Setting the visualization background color to transparent with `graphviz_kwargs=dict(graph_attr=dict(bgcolor="transparent"))` would also make nodes transparent. It is very hard to fix by user in an ad hoc way because of graphviz's quirks.

before
image

after
image

## Changes
- change default attributes
- update tests accordingly

## How I tested this
- successfully ran the test suite locally

## Notes
- I am making this change to help with the VSCode Hamilton extension.
- edges can be set to white or black for visibility depending on the background

## Checklist

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -299,19 +299,20 @@ def create_graphviz_graph(
         if node_type == "config":
             node_style = dict(
                 shape="note",
-                style="",
+                style="filled",
                 fontname=fontname,
             )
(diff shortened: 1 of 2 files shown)


skrawcz (approved):
I don't think this will break anyone's stuff. So looks good.

PR #766, opened 2024-03-18
added default `CacheAdapter` lifecycle hook

zilto:
This follows a similar logic to `hamilton.plugins.h_diskcache`, but uses the standard `shelve` library. It has the same API as the diskcache, but provides less utilities to manage cache size.

## Changes
- added a `CacheAdapter` to `hamilton.lifecycle.default`
- added relevant tests, derived from `tests/plugins/h_diskcache`

## How I tested this
- ran tests successfully locally

## Notes
- one edge case regarding cache history is mentioned in TODOs, but it shouldn't affect usability.

## Checklist

### DIFF
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -1,15 +1,18 @@
 """A selection of default lifeycle hooks/methods that come with Hamilton. These carry no additional requirements"""
 
+import hashlib
 import logging
 import pdb
+import pickle
 import pprint
 import random
+import shelve
 import time
 from typing import Any, Callable, Dict, List, Optional, Union
 
-from hamilton import htypes
-from hamilton.lifecycle import NodeExecutionHook
-from hamilton.lifecycle.api import NodeExecutionMethod
+from hamilton import graph_utils
+from hamilton.graph_types import HamiltonGraph
+from hamilton.lifecycle import GraphExecutionHook, NodeExecutionHook, NodeExecutionMethod
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


skrawcz:
@zilto this should have a companion example, and also docs entry.

hamilton/lifecycle/default.py
```diff
@@ -313,6 +316,109 @@ def run_after_node_execution(
             pdb.set_trace()
 
 
+class CacheAdapter(NodeExecutionHook, NodeExecutionMethod, GraphExecutionHook):
+    """Class to cache node results on disk with a key baed on the node code implementation and inputs.
```
elijahbenizzy, line 320:
```suggestion
    """Class to cache node results on disk with a key based on the node code implementation and inputs.
```

hamilton/lifecycle/default.py
```diff
@@ -313,6 +316,109 @@ def run_after_node_execution(
...
+    ):
+        """Create cache key based on node callable hash (equiv. to HamiltonNode.version) and
+        the node inputs (`node_kwargs`).If key in cache (cache hit), load result; else (cache miss),
+        compute the node and append node name to `used_nodes_hash`.
+
+        Note:
+            - the callable hash is stored  in `used_nodes_hash` because it's required to create the
+            key in `run_after_node_execution` and the callable won't be accessible to recompute it
+        """
+        if node_name not in self.cache_vars:
```
elijahbenizzy, line 368: For afterwards -- we may want to make this a set? Odds are high this is not a slow operation, but it could get `O(n**2)` which we don't want.

elijahbenizzy (approved):
Looks good! A few comments to fix after, merging.

PR #768, opened 2024-03-18
Hamilton + AWS

tyapochkin:
Draft of PR with additional examples/features related to AWS.

Current roadmap (provisional):
1. [x] Example of deployment Hamilton functions in AWS Lambda.
2. [x] Example of deployment jobs with Hamilton functions in AWS Glue.
3. [x] Example of running a processing job with Hamilton functions in AWS SageMaker.
4. [ ] Plugin(?) for easy deployment jobs with Hamilton functions in AWS Glue.

## Changes
TODO

## How I tested this
TODO

## Notes
TODO

## Checklist


examples/aws/lambda/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM public.ecr.aws/lambda/python:3.11
+
+COPY requirements.txt ./
+RUN pip install -r requirements.txt
+
```
zilto, line 5: You could install `graphviz` to be able to generate visualization. Then, in your JSON reponse, you could include an SVG string of the viz.

The code would look like:
```python
inputs = {"input_table": df}

dot = dr.visualize_execution(final_vars=output_cols, inputs=inputs)
svg_str = dot.pipe(format="svg")

df_result = dr.execute(output_columns, inputs=inputs )

return {"statusCode": 200, "body": dict(results=df_result.to_json(orient="split"), viz=svg_str)}
```

skrawcz: I'm not sure there's much value to creating an image on every lambda request. But yes something to document as a possibility.

tyapochkin: Yes, Lambda is usually used as part of a backend when we want to execute the same code repeatedly for various inputs. It's infrequently used for data analysis or tasks when visualization is helpful. So, adding visualization to its output seems, for me, not really necessary here, also because all visualizations will actually be the same for all lambda invocations.

However, I see a bit more opportunity here. What if we add to `Driver.export_execution` an additional bool argument like `add_outputs_to_json` that will turn on adding the output of each node in the JSON representation? Yes, JSONs will be larger, but we'll be able to save them to S3 as logs of executions (for debugging, monitoring, or stuff like that). What do you think about it?

skrawcz: > However, I see a bit more opportunity here. What if we add to Driver.export_execution an additional bool argument like add_outputs_to_json that will turn on adding the output of each node in the JSON representation? Yes, JSONs will be larger, but we'll be able to save them to S3 as logs of executions (for debugging, monitoring, or stuff like that). What do you think about it?

I think if people want to capture outputs, that's what the lifecycle adapters can easily enable. I'd be hesitant in including this functionality here because I could see it becoming difficult to manage, and in practice not that practical -- since the graph might not change, but the inputs/outputs would. If someone wants this, they will have to write custom code to parse the graph & the outputs, in which case coupling them into a single JSON to me doesn't really help all that much with what can be achieved already.

tyapochkin: @zilto, I added DAG visualization to the example for AWS SageMaker job processing. This feature is more often used for analytic purposes, so an example of visualization could be more useful here.

examples/aws/sagemaker/requirements.txt
```diff
@@ -0,0 +1 @@
+sagemaker
```
skrawcz, line 1: sf-hamilton too right?

skrawcz (approved):

PR #774, opened 2024-03-21
`fix` reduce type constraints on JSON materializer

zilto:
This issue is a follow-up on #690 

Before, the JSON Loader&Saver would only support `dict` type objects. However, the [JSON format specs](https://www.json.org/json-en.html) are more flexible. We then introduced support for `List[dict]`.

However, it seems reasonable to support more generally `list` (of `str`, `int`, `None`, etc.). By making it more flexible, we delegate to the underlying `json` Python library to surface "not json-serializable" errors to the user.

## Changes
- changed supported types from `List[dict]` to `list`
- adjusted tests accordingly

## How I tested this
- test succeed locally

## Checklist

### DIFF
--- a/hamilton/io/default_data_loaders.py
+++ b/hamilton/io/default_data_loaders.py
@@ -4,7 +4,7 @@ import json
 import os
 import pathlib
 import pickle
-from typing import Any, Collection, Dict, List, Tuple, Type, Union
+from typing import Any, Collection, Dict, Tuple, Type, Union
 
 from hamilton.io.data_adapters import DataLoader, DataSaver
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
OK, I think this should be 100% backwards compatible. Only issue is if the original json gets passed over to someone's custom json one, but that should work anyway (as we'll do the same thing). Might want to add a note on this.

Basically, we evaluate on type + order -- they're supposed to add their custom one in last, as it is done in reverse order. Given that, I don't think this will break anything.

PR #775, opened 2024-03-22
Polars Lazyframe Support

buggtb:
This PR is to aid for support of Polars LazyFrames in Hamilton.

## Changes

I've currently stubbed out the CSV Reader / Writer to work both on Eager and Lazy mode in Polars.

## How I tested this

There is an accompanying test that mimics the Eager test but using Lazyframes instead

## Notes

I've also had to update the get_dataframe_metadata in utils.py to allow it to work with Lazyframes that don't have a row count. I abstracted all the lookups so that if others passed/failed in the future for support of other read/writers they would return what they can.

## Checklist


skrawcz (changes requested):
Thanks @buggtb  for driving this!

I think we can simplify a few things actually.

1. `polars_extensions.py`: All existing polars `DataSaver`s can have the following `applicable_types()`:
```python
    @classmethod
    def applicable_types(cls) -> Collection[Type]:
        return [DATAFRAME_TYPE, pl.LazyFrame]
```
Then in the `save_data()` method, add union type annotation, and then in the body:
```python
if isinstance(data, pl.LazyFrame):
    data = data.collect()
# continue with rest of function
```

2. `h_polars.py` should be modified to:
(a) include a PolarsLazyDataFrameResult -- this won't call collect, it will just return a LazyFrame. (though not sure if it makes sense)
(b) the existing PolarsDataFrameResult should have isinstance checks for lazyframe, and if found, should collect and move on.
This means we can delete `h_polars_lazyframe.py`.

3. `polars_lazyframe_extensions.py`: I would use `spark_extensions.py` as the template. No need to copy all of what `polars_extesions.py` has.  Otherwise it should house the `scan_*` and `sink_*` DataLoaders and DataSavers, e.g. `scan_csv`. Let me know if you need more guidance here.

5. Remove `polars_shared.py`. 
I think this is a little over engineering. I don't see the value in the abstraction right now -- I think a little duplication is more straightforward right now.

6. `io/utils.py`. I think the changes here are okay. @elijahbenizzy and I can think about whether there's a nicer way to do this.

buggtb:
Thanks for the review @skrawcz 

A couple of follow up questions:

https://github.com/DAGWorks-Inc/hamilton/pull/775/files#diff-fbd89fcde3b2c949a6f81da8ecfd52a163490f1645f1eca8f9451047cf7785faR179

the save function is fair enough, the load is more problematic in terms of design, if I want to load into a lazy frame instead of a dataframe, there's nothing in there currently to allow that. I could extend that function to default to dataframe and let a user request a lazyframe I guess. I'm unsure how you'd want that pattern to look perhaps something like this, or something else:
```
    def load_data(self, type_: Type, frametype: Union[DATAFRAME_TYPE, pl.LazyFrame] = pl.DataFrame) -> Tuple[DATAFRAME_TYPE, Dict[str, Any]]:
        if isinstance(type_, pl.LazyFrame):
            df = pl.scan_csv(self.file, **self._get_loading_kwargs())
        else:
            df = pl.read_csv(self.file, **self._get_loading_kwargs())

        metadata = utils.get_file_and_dataframe_metadata(self.file, df)
        return df, metadata
```

also there are some minor differences in kwargs in the lazyframe when compared to the dataframe and so I'd have to add some conditional logic to the get_kwargs() functions to return the valid sets there.

Not entirely sure how the spark extensions example applies here as I suspect we'll need loaders and savers for each type, but I'm probably missing something.

skrawcz:
> Thanks for the review @skrawcz
> 
> A couple of follow up questions:
> 
> https://github.com/DAGWorks-Inc/hamilton/pull/775/files#diff-fbd89fcde3b2c949a6f81da8ecfd52a163490f1645f1eca8f9451047cf7785faR179
> 
> the save function is fair enough, the load is more problematic in terms of design, if I want to load into a lazy frame instead of a dataframe, there's nothing in there currently to allow that. I could extend that function to default to dataframe and let a user request a lazyframe I guess. I'm unsure how you'd want that pattern to look perhaps something like this, or something else:
> 
> ```
>     def load_data(self, type_: Type, frametype: Union[DATAFRAME_TYPE, pl.LazyFrame] = pl.DataFrame) -> Tuple[DATAFRAME_TYPE, Dict[str, Any]]:
>         if isinstance(type_, pl.LazyFrame):
>             df = pl.scan_csv(self.file, **self._get_loading_kwargs())
>         else:
>             df = pl.read_csv(self.file, **self._get_loading_kwargs())
> 
>         metadata = utils.get_file_and_dataframe_metadata(self.file, df)
>         return df, metadata
> ```
> 
> also there are some minor differences in kwargs in the lazyframe when compared to the dataframe and so I'd have to add some conditional logic to the get_kwargs() functions to return the valid sets there.
> 

Yep, so I'm thinking don't couple them at all precisely for the reasons you mentioned.  So `polars_lazyframe_extensions.py` would contain the "load" ones specific to creating lazyframes that pertain to `scan_csv`, `scan_parquet` etc.  Does that clarify it more? 

> Not entirely sure how the spark extensions example applies here as I suspect we'll need loaders and savers for each type, unless we can overload an existing class but that doesn't appear to be what the spark_extensions example does, but I'm probably missing something.

I just mean to use it in terms of structure. Since pyspark doesn't make use of the column stuff. Lazyframes wouldn't make use of it either.

buggtb:
Okay so I removed Polars shared, I then removed most from polars_lazyframe_exensions but kept the loader. Can you check to make sure it's as you imaged, now I see how the save_to and load_from annotations work it makes sense now, I wasn't sure how you could blend loaders and savers across classes, all good.

In h_polars.py I took the same logic I had in the lazyframe version and I do run collect and return a dataframe. When I was running it, to me at least it made sense to collect at the end and return a dataframe as the pipeline is done, but now I think about it there are times, perhaps if you're running chained pipelines, subdags or whatever that you may not want to, I don't know, the jury is out there!

hamilton/plugins/polars_extensions.py
```diff
@@ -108,7 +111,7 @@ class PolarsCSVReader(DataLoader):
 
     @classmethod
     def applicable_types(cls) -> Collection[Type]:
-        return [DATAFRAME_TYPE]
+        return [DATAFRAME_TYPE, pl.LazyFrame]
```
skrawcz, line 114: We don't want this here -- only for Loaders.
```suggestion
        return [DATAFRAME_TYPE]
```

hamilton/plugins/polars_lazyframe_extensions.py
```diff
@@ -0,0 +1,183 @@
...
+        scalar_value = [scalar_value]
+    return df.with_columns(pl.Series(name=column_name, values=scalar_value))
+
+
+def register_types():
+    """Function to register the types for this extension."""
+    registry.register_types("polars_lazyframe", DATAFRAME_TYPE, COLUMN_TYPE)
+
+
+register_types()
```
skrawcz, line 66: We don't need the column stuff. So it can become:
```suggestion
DATAFRAME_TYPE = pl.LazyFrame
COLUMN_TYPE = None
COLUMN_FRIENDLY_DF_TYPE = False

def register_types():
    """Function to register the types for this extension."""
    registry.register_types("polars_lazyframe", DATAFRAME_TYPE, COLUMN_TYPE)


register_types()
```

hamilton/plugins/polars_lazyframe_extensions.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+@dataclasses.dataclass
+class PolarsCSVReader(DataLoader):
```
skrawcz, line 70: Just to make it clear what it wraps:
```suggestion
class PolarsScanCSVReader(DataLoader):
```

skrawcz (commented):
Yep looking good. Added code for what I would change.

skrawcz:
> In h_polars.py I took the same logic I had in the lazyframe version and I do run collect and return a dataframe. When I was running it, to me at least it made sense to collect at the end and return a dataframe as the pipeline is done, but now I think about it there are times, perhaps if you're running chained pipelines, subdags or whatever that you may not want to, I don't know, the jury is out there!

Yep! So that's a vote for a second one specific to LazyFrames IMO.

hamilton/plugins/polars_extensions.py
```diff
@@ -200,7 +202,7 @@ class PolarsCSVWriter(DataSaver):
...
```
skrawcz, line 205: Can then add this for all the other polars loaders. A few were just merged, so you might want to rebase from main if you haven't already.

hamilton/plugins/polars_extensions.py
skrawcz: sorry I meant savers not loaders.

hamilton/plugins/h_polars_lazyframe.py
```diff
@@ -0,0 +1,46 @@
...
+    Use this when you want to create a polars dataframe from the outputs. Caveat: you need to ensure that the length
+    of the outputs is the same, otherwise you will get an error; mixed outputs aren't that well handled.
+
+    To use:
+
+    .. code-block:: python
+
+        from hamilton import base, driver
+        from hamilton.plugins import polars_extensions
+        polars_builder = polars_extensions.PolarsDataFrameResult()
```
skrawcz, line 20:
```suggestion
        polars_builder = polars_extensions.PolarsLazyFrameResult()
```

hamilton/plugins/polars_extensions.py
```diff
@@ -69,6 +70,8 @@ def fill_with_scalar_polars(df: pl.DataFrame, column_name: str, scalar_value: An
 
 
 def register_types():
+    logger = logging.getLogger(__name__)
+    logger.info("registering lazyframe")
```
skrawcz, line 74:
```suggestion
```

hamilton/plugins/polars_extensions.py
```diff
@@ -1,4 +1,5 @@
 import dataclasses
+import logging
```
skrawcz, line 2:
```suggestion
```

hamilton/plugins/polars_extensions.py
skrawcz: not needed.

hamilton/plugins/polars_extensions.py
skrawcz, line 1: There's a few more Savers to add the pl.LazyFrame as an applicable type to.

skrawcz (commented):
Looking great. A few minor things.

We should also have a lazyframe test for each of the savers. If code duplication is the simplest here, I think that's fine; not sure there's a DRY way that would be worth the time/effort since this is a one off.

hamilton/plugins/polars_extensions.py
buggtb: Faster reviews than I can push the code! Cleaned up the logging code and pushed additional tests for the savers.

tests/plugins/test_polars_lazyframe_extensions.py
```diff
@@ -0,0 +1,177 @@
...
+def test_lazy_polars_parquet(df: pl.LazyFrame, tmp_path: pathlib.Path) -> None:
+    file = tmp_path / "test.parquet"
+
+    writer = PolarsParquetWriter(file=file)
+    kwargs1 = writer._get_saving_kwargs()
+    writer.save_data(df)
+
+    reader = PolarsScanParquetReader(file=file, n_rows=2)
+    kwargs2 = reader._get_loading_kwargs()
+    df2, metadata = reader.load_data(pl.DataFrame)
```
skrawcz, line 60: I think this should be lazy frame right? (it's inconsequential but it's a little confusing to read)
```suggestion
    df2, metadata = reader.load_data(pl.LazyFrame)
```

tests/plugins/test_polars_lazyframe_extensions.py
```diff
@@ -0,0 +1,177 @@
...
+    assert kwargs1["compression"] == "zstd"
+    assert kwargs2["n_rows"] == 2
+    assert_frame_equal(df.collect(), df2.collect())
+
+
+def test_lazy_polars_feather(tmp_path: pathlib.Path) -> None:
+    test_data_file_path = "tests/resources/data/test_load_from_data.feather"
+    reader = PolarsScanFeatherReader(source=test_data_file_path)
+    read_kwargs = reader._get_loading_kwargs()
+    df, _ = reader.load_data(pl.DataFrame)
```
skrawcz, line 73:
```suggestion
    df, _ = reader.load_data(pl.LazyFrame)
```

skrawcz (commented):
Thanks @buggtb for the work! Looking good I think we just need an example to help people show how to use the new functionality you just added!

So two things:

1. Can you add an example exercising the scan variants? E.g. add something to `examples/polars/lazyframe`?  I think we should add this to complete this PR.
2. There are two minor things (see suggested commits), I can do the commit before merging, or you can. Let me know.

Otherwise question, should I create an issue for the `sink_*` variants for lazyframe writing to csv, etc. to track that? No need to add it in this PR, but I can see that be a natural progression if we have scan_* support.

buggtb:
Added a test which loads a CSV using the annotation, does a simple sum and dumps out the result as a LazyFrame, added a comment in my_script.py explaining how you could use either LazyFrame or DataFrame as a resultset depending on what you needed.

Added https://github.com/DAGWorks-Inc/hamilton/issues/792 and https://github.com/DAGWorks-Inc/hamilton/issues/791 on the issues lists for some missing sources and sinks

skrawcz (approved):

skrawcz:
Argh, I see there's a rebase that's needed. Mind doing that? Then we'll be able to squash merge.

buggtb:
Pulled the latest main branch into my fork, should be good, I think.

skrawcz:
hmm -- 🤔 
Screen Shot 2024-03-28 at 10 02 42 AM

buggtb:
Screenshot 2024-03-28 at 17 05 19

I'm not sure why I don't see the same

skrawcz:
Done. Thanks @buggtb 🍾 . This will go out next week.

PR #787, opened 2024-03-27
`fix` can't sort `None` in `HamiltonGraph.version`

zilto:
`HamiltonNode` can take value `None` when they have no `.originating_functions[0]`. This lead to an issue when `HamiltonGraph.version` derives its value from its constituent nodes.

## Changes
- sort nodes based on their `name` attribute; this change will break lineage with hashes from previous Hamilton versions
- replace `None` for `""` explicitly

## How I tested this
- the existing tests succeed locally
- ran the `examples/experiment_manager/run.py` successfully 
- manually checked that the experiment UI behaved properly

## Notes
- we should add tests for `.version` on more complex DAG that use function modifiers, materializers, config, etc.

## Checklist

### DIFF
--- a/hamilton/graph_types.py
+++ b/hamilton/graph_types.py
@@ -203,5 +203,6 @@ class HamiltonGraph:
         Node hashes are in a sorted list, then concatenated as a string before hashing.
         To find differences between dataflows, you need to inspect the node level.
         """
-        sorted_node_versions = sorted([n.version for n in self.nodes])
+        sorted_nodes = sorted(self.nodes, key=lambda n: n.name)
+        sorted_node_versions = [n.version if n.version else "" for n in sorted_nodes]
         return hashlib.sha256(str(sorted_node_versions).encode()).hexdigest()


hamilton/graph_types.py
```diff
@@ -203,5 +203,6 @@ def version(self) -> str:
         Node hashes are in a sorted list, then concatenated as a string before hashing.
         To find differences between dataflows, you need to inspect the node level.
         """
-        sorted_node_versions = sorted([n.version for n in self.nodes])
+        sorted_nodes = sorted(self.nodes, key=lambda n: n.name)
+        sorted_node_versions = [n.version if n.version else "" for n in sorted_nodes]
```
elijahbenizzy, line 207: I wonder if we want to fall back on the node name? E.G. if adding a materializer will make it the same hash?

hamilton/graph_types.py
```diff
@@ -203,5 +203,6 @@ def version(self) -> str:
         Node hashes are in a sorted list, then concatenated as a string before hashing.
         To find differences between dataflows, you need to inspect the node level.
         """
-        sorted_node_versions = sorted([n.version for n in self.nodes])
+        sorted_nodes = sorted(self.nodes, key=lambda n: n.name)
```
elijahbenizzy, line 206: Let's add a unit test -- you can just create a node with a null version or ensure it has a function that doesn't give one.

hamilton/graph_types.py
zilto: We'll that's the question. IMO, `HamiltonGraph.version` should return a hash equivalent to the state of the graph, so different materializers, different graph hash.

The idea is that by sorting by name, your `None` value are in the same position before being converted to `""`. Since we're concatenating everything to a string after, it should be equivalent to sorting node hashes and `""`.

elijahbenizzy: Yep, agreed. If you use a "", however, there are edge-cases in which it could be the same if it changes in some way? E.G. add a third matierializer that goes right after the second?

Also, the position of `""` doesn't matter, it won't effect it at all, right? E.G. no matter where the materializer are, an empty string has no impact. So won't this be the same regardless of where they end up?

zilto:
closing in favor of #789

PR #788, opened 2024-03-27
`feature` render Driver in notebooks

zilto:
Simple quality of life improvement to directly call `driver.display_all_functions()`. The change was inspired by the logic found in Jupyter Magic.

This simply propagates the "render attribute" of the `graphviz.Digraph` object returned by `driver.display_all_functions()` so it should be robust.

image

## Changes
- added `._repr_mimebundle_()` to `Driver` class

## How I tested this
- tests succeeded locally
- intended behavior works in Jupyter notebooks (browser interface) and VSCode notebooks

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -399,6 +399,16 @@ class Driver:
             # TODO -- update this to use the lifecycle methods
             self.capture_constructor_telemetry(error, modules, config, adapter)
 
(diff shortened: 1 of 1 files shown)


skrawcz (approved):
Looks good. Can you make the PR message the body of the commit for this please, since the PR message provides a good explanation of why that would be useful to have in the commit. Thanks!

PR #789, opened 2024-03-27
Adds guards around version logic

skrawcz:
HamiltonNode.version was returning None in certain cases. This broke graph.version. Culprits: config, and materializers.

This change:

1. makes the node version for config == name of the node.
2. ~makes materializers all have originating function values.~ Most do get values. So no need to modify them.
3. guards graph.version against computing functions with None as the version value.

So node.version should now hardly ever be None. But if it is for some strange reason, we should now handle it better.

Adds tests for it.
Also fixed a fixture that was failing for me for the shelve caching adapter.

## Changes
 - graph_types
 - tests

## How I tested this
- locally
- tested experiment example run.py works without issue

## Notes
- 

## Checklist

### DIFF
--- a/hamilton/function_modifiers/adapters.py
+++ b/hamilton/function_modifiers/adapters.py
@@ -233,6 +233,7 @@ class LoadFromDecorator(NodeInjector):
             namespace=(
                 (namespace, "load_data") if namespace else ("load_data",)
             ),  # We want no namespace in this case
+            originating_functions=(load_data,),
         )
 
         # the filter node is the node that takes the data from the data source, filters out
@@ -259,6 +260,7 @@ class LoadFromDecorator(NodeInjector):
             # to have this weird DAG shape. For now, this solves the problem, and this is an internal component of the API
             # so we're good to go
             namespace=((namespace, "select_data") if namespace else ()),  # We want no namespace
+            originating_functions=(filter_function,),
         )
         return [loader_node, filter_node]
 
@@ -576,6 +578,7 @@ class SaveToDecorator(SingleNodeNodeTransformer):
                 "hamilton.data_saver.sink": f"{saver_cls.name()}",
                 "hamilton.data_saver.classname": f"{saver_cls.__qualname__}",
             },
+            originating_functions=(save_data,),
         )
         return save_node
 
--- a/hamilton/function_modifiers/adapters.py
+++ b/hamilton/function_modifiers/adapters.py
@@ -233,7 +233,6 @@ class LoadFromDecorator(NodeInjector):
             namespace=(
                 (namespace, "load_data") if namespace else ("load_data",)
             ),  # We want no namespace in this case
-            originating_functions=(load_data,),
         )
 
         # the filter node is the node that takes the data from the data source, filters out
@@ -260,7 +259,6 @@ class LoadFromDecorator(NodeInjector):
             # to have this weird DAG shape. For now, this solves the problem, and this is an internal component of the API
             # so we're good to go
             namespace=((namespace, "select_data") if namespace else ()),  # We want no namespace
-            originating_functions=(filter_function,),
         )
         return [loader_node, filter_node]
 
@@ -566,7 +564,6 @@ class SaveToDecorator(SingleNodeNodeTransformer):
             key: value for key, value in input_types.items() if key not in resolved_kwargs
         }
         input_types[node_to_save] = (node_.type, DependencyType.REQUIRED)
-
         save_node = node.Node(
             name=artifact_name,
             callabl=save_data,
@@ -578,7 +575,6 @@ class SaveToDecorator(SingleNodeNodeTransformer):
                 "hamilton.data_saver.sink": f"{saver_cls.name()}",
                 "hamilton.data_saver.classname": f"{saver_cls.__qualname__}",
             },
-            originating_functions=(save_data,),
         )
         return save_node
 
--- a/hamilton/graph_types.py
+++ b/hamilton/graph_types.py
@@ -154,9 +154,13 @@ class HamiltonNode:
         The option `strip=True` means docstring and comments are ignored
         when hashing the function.
         """
-        if not self.originating_functions:
-            return None
+        if self.originating_functions is None or len(self.originating_functions) == 0:
+            if self.is_external_input:
+                # return the name of the config node. (we could add type but skipping for now)
+                return self.name
+            return None  # this shouldn't happen often.
         try:
+            # return hash of first function. It could be that others are Hamilton framework code.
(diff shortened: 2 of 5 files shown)


hamilton/graph_types.py
```diff
@@ -154,9 +154,13 @@ def version(self) -> typing.Optional[str]:
         The option `strip=True` means docstring and comments are ignored
         when hashing the function.
         """
-        if not self.originating_functions:
-            return None
+        if self.originating_functions is None or len(self.originating_functions) == 0:
```
zilto, line 157: If `self.originating_functions == None`, `len(self.originating_functions)` should throw an exception "can't call len() on NoneType", no?

tests/test_graph_types.py
```diff
@@ -113,14 +114,42 @@ def node_to_create(required_dep: int, optional_dep: int = 1) -> str:
     }
 
 
-def test_create_hamilton_node_missing_version():
+def test_create_hamilton_config_node_verssion():
```
zilto, line 117: typo in func name: `_verssion()`-> `_version()`

tests/test_graph_types.py
```diff
@@ -113,14 +114,42 @@ def node_to_create(required_dep: int, optional_dep: int = 1) -> str:
...
+    hamilton_node = graph_types.HamiltonNode.from_node(n)
+    # the above will have no specified versions
     assert hamilton_node.version is None
     assert hamilton_node.as_dict()["version"] is None
 
 
+def test_hamilton_graph_version_normal():
+    dr = driver.Builder().with_modules(no_parallel).build()
+    graph = graph_types.HamiltonGraph.from_graph(dr.graph)
+    assert graph.version == "3b3487599ccc4fc56995989c6d32b58a90c0b91b8c16b3f453a2793f47436831"
```
zilto, line 142: these version values can't be hardcoded because the hashing is dependent on the Python version. see `test_create_hamilton_node()` from this file

hamilton/graph_types.py
skrawcz: Because it's an `OR` statement, it won't evaluate the second clause unless the first one is False.

tests/test_graph_types.py
skrawcz: sure - but it seems like only python 3.8 failed.

hamilton/graph_types.py
skrawcz:
```python
if True or obj.asdfs.asf():
    print("passed")
# passed

if False or obj.asdfs.asf():
    print("not hit")
    
  File ... in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-7945ebca460e>", line 1, in <module>
    if False or obj.asdfs.asf():
NameError: name 'obj' is not defined
```

hamilton/function_modifiers/adapters.py
```diff
@@ -233,6 +233,7 @@ def get_input_type_key(key: str) -> str:
             namespace=(
                 (namespace, "load_data") if namespace else ("load_data",)
             ),  # We want no namespace in this case
+            originating_functions=(load_data,),
```
elijahbenizzy, line 236: OK, this `load_data` function will always be a local one. So its not particularly interesting. The spec for `originating_functions` is that its the function's code that generated it. So this will load up a local lambda, and will actually already be assigned (it should be). 

In the case of materializers this is not assigned, so that's why it errors out. So this here will provide really weird behavior in the decorating case.

hamilton/function_modifiers/adapters.py
```diff
@@ -259,6 +260,7 @@ def filter_function(_inject_parameter=inject_parameter, **kwargs):
             # to have this weird DAG shape. For now, this solves the problem, and this is an internal component of the API
             # so we're good to go
             namespace=((namespace, "select_data") if namespace else ()),  # We want no namespace
+            originating_functions=(filter_function,),
```
elijahbenizzy, line 263: Same as above, this is a local lambda, and won't tell us anything interesting. This should already be added.

hamilton/io/materialization.py
```diff
@@ -247,7 +247,9 @@ def join_function(**kwargs):
                 callabl=join_function,
                 input_types={dep.name: dep.type for dep in node_dependencies},
                 originating_functions=(
-                    None if self.result_builder is None else [self.result_builder.build_result]
```
elijahbenizzy, line 250: This will be a locally defined function and will has the same for every single case in which it is not `None`. Is that intended?

elijahbenizzy: I guess in this case that makes sense -- it should just be the same?

tests/test_graph_types.py
```diff
@@ -113,14 +114,52 @@ def node_to_create(required_dep: int, optional_dep: int = 1) -> str:
...
     assert hamilton_node.version is None
     assert hamilton_node.as_dict()["version"] is None
 
 
+def test_hamilton_graph_version_normal():
+    dr = driver.Builder().with_modules(no_parallel).build()
+    graph = graph_types.HamiltonGraph.from_graph(dr.graph)
+    # assumption is for python 3
+    if sys.version_info.minor == 8:
+        hash_value = "0a375f3366590453dea8927d4c02c15dc090f8be42e9129d9a1139284eac920c"
```
elijahbenizzy, line 144: Huh? This is a bit confusing... Why is it these hashes/why is it different between versions? Best to just test its a certain amount of characters IMO

skrawcz: yeah hashing can change between python versions. Only found it to be different between 3.8 and the rest. Since I've already gone and done it, no need to shorten.

hamilton/function_modifiers/adapters.py
skrawcz: removing.

hamilton/io/materialization.py
skrawcz: reverting

hamilton/function_modifiers/adapters.py
```diff
@@ -576,6 +578,7 @@ def get_input_type_key(key: str) -> str:
                 "hamilton.data_saver.sink": f"{saver_cls.name()}",
                 "hamilton.data_saver.classname": f"{saver_cls.__qualname__}",
             },
+            originating_functions=(save_data,),
```
skrawcz, line 581: Reverting -- but something like this actually works:

```python
originating_functions = (saver_cls.save_data,) if hasattr(saver_cls, "save_data") else None
```

hamilton/function_modifiers/adapters.py
skrawcz:
```python
(loader_cls.load_data,)
``` 
would be a better fit

Issue #685, opened 2024-02-08
Easy way to scrape local functions


elijahbenizzy:
Can we get all fns that are defined in the local module and insert them into a driver?

Two workarounds for now:
1. `create_temporary_module`
2. `__main__` hack

```python
def foo() -> int:
    return 1


from hamilton import driver

if __name__ == '__main__':
    import __main__
    dr = driver.Builder().with_modules(__main__).build()
    print(dr.execute(["foo"]))
```

Thinking:
```python 
driver.Builder().with_local_modules().build()
```

```python
driver.Builder().with_modules(hamilton.find_local()).build()

zilto:
A generic utility function to gather function definitions from local scope is planned.

PR #793, opened 2024-03-28
Build with Local Function Directly

swapdewalkar:
--- PR TEMPLATE INSTRUCTIONS (1) ---

Address this [Issue #685 ](https://github.com/DAGWorks-Inc/hamilton/issues/685)

## Changes
- Added a local builder helper.
- Used python inspect to find out parent module and included with in sys modules.
## How I tested this
Unit Tests

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1669,6 +1669,18 @@ class Builder:
         self.config.update(config)
         return self
 
+    def with_local_modules(self) -> "Builder":
+        """Adds the specified modules to the modules list.
+        This can be called multiple times -- later calls will take precedence.
+
+        :param file_path: Path to the file to import.
+        :return: self
+        """
+        import inspect
+        module = inspect.getmodule(inspect.stack()[1][0])
+        self.modules.append(module)
+        return self
+
     def with_modules(self, *modules: ModuleType) -> "Builder":
         """Adds the specified modules to the modules list.
         This can be called multiple times -- later calls will take precedence.
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1675,6 +1675,7 @@ class Builder:
         :return: self
         """
         import inspect
+
         module = inspect.getmodule(inspect.stack()[1][0])
         self.modules.append(module)
         return self
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1670,10 +1670,8 @@ class Builder:
         return self
 
     def with_local_modules(self) -> "Builder":
-        """Adds the specified modules to the modules list.
-        This can be called multiple times -- later calls will take precedence.
+        """Adds the local modules to the modules list.
 
-        :param file_path: Path to the file to import.
         :return: self
         """
         import inspect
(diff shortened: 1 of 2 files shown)


swapdewalkar:
@skrawcz I saw the requirement [here ](https://github.com/DAGWorks-Inc/hamilton/issues/685) to get all fns that are defined in the local module and insert them into a driver.

```
dr = driver.Builder()
      .with_local_modules()
      .build()
```

instead of 
```
if __name__ == '__main__':
    import __main__
    dr = driver.Builder()
                     .with_modules(__main__)
                     .build()
    print(dr.execute(["foo"]))
````

Let me know if I understood something different from ticket.

File: tests/test_driver_local_modules.py [hunk 1]
```diff
@@ -0,0 +1,7 @@
+from hamilton.driver import Builder, DefaultGraphExecutor
+
+
+def test_driver_with_local_modules() -> None:
+    dr = Builder().with_local_modules().build()
+    assert isinstance(dr.graph_executor, DefaultGraphExecutor)
+    assert __name__ == list(dr.graph_modules)[0].__name__
```
skrawcz, line 7: I think this test probably needs to pull in a function and exercise the driver.

e.g. define 
```python
def a(b:int) -> int:
   return b * 2
```

Then we should execute the driver to get "a". 

That should work right?

skrawcz:
@swapdewalkar cool makes sense -- want to comment on that ticket so I can assign it to you? 

Otherwise I think we just need a test to exercise this properly.

zilto:
@swapdewalkar thanks for the contribution!

The Python import system behaves in strange ways at times, so it would be beneficial if you could add to the docstring of `.with_local_modules()` a snippet of what Python code should be in your `.py` file and how to execute the code.

For instance, the current approach with `inspect.stack()` seems to work for `python -m run` but will fail for `python run.py` of the same file. In that case, it can lead to hard to debug exceptions if uncaught

```
  File "~/run.py", line 37, in <module>
    print(s[1])
IndexError: list index out of range
```

hamilton/driver.py
```diff
@@ -1669,6 +1669,17 @@ def with_config(self, config: Dict[str, Any]) -> "Builder":
...
         return self
 
+    def with_local_modules(self) -> "Builder":
+        """Adds the local modules to the modules list.
+
+        :return: self
+        """
+        import inspect
+
+        module = inspect.getmodule(inspect.stack()[1][0])
```
skrawcz, line 1679: See Thierry's comment. I think we can make this more robust.

hamilton/driver.py
```diff
@@ -1669,6 +1669,17 @@ def with_config(self, config: Dict[str, Any]) -> "Builder":
         self.config.update(config)
         return self
 
+    def with_local_modules(self) -> "Builder":
+        """Adds the local modules to the modules list.
```
skrawcz, line 1673: We should add more caveats here. 
E.g. If people use this, they could run into problems when using some of the adapters because this code will be under `__main__`.

swapdewalkar:
I will add more tests!

skrawcz:
Just waiting on more tests for this one :)

zilto:
A more generic "gather functions from local scope" is planned

Issue #794, opened 2024-03-28
Slack notifier. Have one that can notify a slack channel on error.


skrawcz:
Using the [Lifecycle API](https://blog.dagworks.io/p/customizing-hamiltons-execution-with) add one that can notify a slack channel on error.

This should go under `hamilton/plugins/h_slack.py`.

PR #797, opened 2024-03-29
WIP: Slack Notifier

swapdewalkar:
--- PR TEMPLATE INSTRUCTIONS (1) ---

Address [#794 ](https://github.com/DAGWorks-Inc/hamilton/issues/794
)
## Changes
Added adapter for slack
## How I tested this
WIP
## Checklist

### DIFF
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -13,7 +13,6 @@ from typing import Any, Callable, Dict, List, Optional, Union
 from hamilton import graph_types, htypes
 from hamilton.graph_types import HamiltonGraph
 from hamilton.lifecycle import GraphExecutionHook, NodeExecutionHook, NodeExecutionMethod
-from slack_sdk import WebClient
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


skrawcz:
@swapdewalkar could you add the stacktrace to the error message? I assume people will want that. What do you think?

Otherwise this looks like a good first pass. I could see a few extensions but we can add them as needed (e.g. filtering to only nodes that people want to monitor, etc).

docs/reference/lifecycle-hooks/SlackNotifierHook.rst
```diff
@@ -0,0 +1,13 @@
+==========================
+plugins.h_slack.SlackNotifier
+==========================
```
skrawcz, line 3: I think this needs to be longer and it'll work
```suggestion
================================
plugins.h_slack.SlackNotifier
================================

```

docs/reference/lifecycle-hooks/SlackNotifierHook.rst
```diff
@@ -0,0 +1,13 @@
...
```
skrawcz, line 1: and then this file needs to be referenced in an index.rst/md file for the docs to build.

docs/reference/lifecycle-hooks/index.rst
```diff
@@ -53,3 +53,4 @@ Recall to add lifecycle adapters, you just need to call the ``with_adapters`` me
     ProgressBar
     DDOGTracer
     FunctionInputOutputTypeChecker
+    SlackNotifier
```
skrawcz, line 56:
```suggestion
    SlackNotifierHook
```

skrawcz: You reference the file name here.

This will then fix the document build.

skrawcz (approved):
Thanks @swapdewalkar ! 🚀 !

PR #799, opened 2024-04-01
examples: Jaffle shop example added

zilto:
This repository is a reimplementation of the canonical [dbt example jaffle_shop](https://github.com/dbt-labs/jaffle_shop). Here, Hamilton + Ibis provides a Python-centric alternative to dbt ([Learn more](https://hamilton.dagworks.io/en/latest/integrations/ibis/)).  It illustrates data transformations for an ecommerce store. Since it is used by many different frameworks, it can ground comparisons between tools. 
 
## Changes
- added `examples/jaffle-shop`
- included data files

## How I tested this
- ran `run.py` locally

## Notes
- it servers as a good reference for tabular data workflows, but Hamilton supports a much wider set of use cases than dbt

## Checklist


examples/jaffleshop/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Content
+The content and structure aims to match the [original dbt `jaffle_shop`](https://github.com/dbt-labs/jaffle_shop/tree/main) example. On the other hand, Hamilton is just a Python library and is flexible regarding project structure.
+
+- `data/`: samples data; equivalent to `seeds/` in the dbt repo.
+- `dataflows/staging` load raw data and rename columns to avoid naming conflicts; equivalent to `models/staging/` in the dbt repo.
+- `dataflows/customer_flow.py` and `dataflows/order_flow.py` define data transformations; equivalent to `models/customers.sql` and `models/orders.sql` in the dbt repo.
+- `run.py` specify where to load data from and how to execute dataflows.
+
+ Genrally, you'll notice Hamilton aims to reduce the sprawl of configurations (`.yaml`) and documentation (`.md`). Instead, it uses docstrings, type hints, or Python object to couple them to with your code (`.py`).
```
skrawcz, line 15:
```suggestion
 Generally, you'll notice Hamilton aims to reduce the sprawl of configurations (`.yaml`) and documentation (`.md`). Instead, it uses docstrings, type hints, or Python object to couple them to with your code (`.py`).
```

skrawcz (approved):
Looks good.

But question on location. Should this really be under Ibis? Versus standalone jaffle-shop?

zilto:
> 1. But question on location. Should this instead be under Ibis? Versus standalone jaffle-shop?
> 2. Where would a link in the main documentation help?
> 3. Would a vanilla pandas version be helpful?

1. top-level `jaffle-shop` is a bit easier to find for those that know the name, but I agree to move it to Ibis given it's so tightly coupled.
2. I would include it in an upcoming `docs/code-comparisons/dbt` page. Hamilton and dbt might not be direct competitor, but it would give an idea to new users how concepts from translate to the other
3. it would be more work with little payoff IMHO given the goal is to showcase the "declarative orchestration" rather than "the data transforms". The code seems intelligible even if you don't know Ibis.

Other thoughts:
It would be cool to have a canonical Hamilton project (e.g., Ibis uses penguins dataset, dbt jaffle shop). It should serve the "Getting started" step-by-step tutorial, and be extensible to be able to showcase Hamilton features (add data validation, change execution engine, etc.). Will open an issue to spec this

PR #800, opened 2024-04-01
Ray Remote Options

flavour:
## Changes

Support to pass options to ray.remote() via @ray_remote_options() decorator
- uses Tags internally, so ray_remote. namespace is now reserved.

## How I tested this

I setup 4 different Ray nodes with different resources available & then decorated 4 functions to target these nodes and one with no decorator.
I confirmed that the functions ran on the correct nodes when using RayGraphAdapter, just as if I wasn't using Hamilton and that the undecorated function also ran

RayWorkflowGraphAdapter also tested.

## Checklist

- [N/A] Placeholder code is flagged / future TODOs are captured in comments

### DIFF
--- a/hamilton/function_modifiers/metadata.py
+++ b/hamilton/function_modifiers/metadata.py
@@ -55,7 +55,7 @@ class tag(base.NodeDecorator):
         "ccpa",
         "dag",
         "module",
-        #RAY_REMOTE_TAG_NAMESPACE,
+        RAY_REMOTE_TAG_NAMESPACE,
     ]  # Anything that starts with any of these is banned, the framework reserves the right to manage it
 
     def __init__(self, *, target_: base.TargetType = None, **tags: Union[str, List[str]]):
@@ -85,8 +85,11 @@ class tag(base.NodeDecorator):
         node_tags.update(self.tags)
         return node_.copy_with(tags=node_tags)
 
+    def _key_allowed_wrapper(self, key: str) -> bool:
+        return self._key_allowed(key)
+
     @staticmethod
-    def _key_allowed(key: str) -> bool:
+    def _key_allowed(key: str, reserved_forbidden: bool = True) -> bool:
         """Validates that a tag key is allowed. Rules are:
         1. It must not be empty
         2. It can have dots, which specify a hierarchy of order
@@ -100,7 +103,8 @@ class tag(base.NodeDecorator):
         if len(key_components) == 0:
             # empty string...
             return False
-        if key_components[0] in tag.RESERVED_TAG_NAMESPACES:
+        if reserved_forbidden and \
+           key_components[0] in tag.RESERVED_TAG_NAMESPACES:
             # Reserved prefixes
             return False
         for key in key_components:
@@ -134,7 +138,7 @@ class tag(base.NodeDecorator):
         """
         bad_tags = set()
         for key, value in self.tags.items():
-            if (not tag._key_allowed(key)) or (not tag._value_allowed(value)):
+            if (not self._key_allowed_wrapper(key)) or (not tag._value_allowed(value)):
                 if isinstance(value, list):
                     value = str(value)
                 bad_tags.add((key, value))
@@ -285,6 +289,9 @@ class RayRemote(tag):
         super(RayRemote, self).__init__(
             **ray_tags
         )
+
+    def _key_allowed_wrapper(self, key: str) -> bool:
+        return self._key_allowed(key, reserved_forbidden = False)
 
 
 def ray_remote(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
--- a/hamilton/function_modifiers/metadata.py
+++ b/hamilton/function_modifiers/metadata.py
@@ -309,7 +309,7 @@ def ray_remote_options(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
 
     .. code-block:: python
 
-        @ray_remote(
+        @ray_remote_options(
             num_gpus=1,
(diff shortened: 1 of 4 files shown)


hamilton/function_modifiers/metadata.py
```diff
@@ -271,3 +277,43 @@ def example_schema() -> pd.DataFrame:
...
+
+        super(RayRemote, self).__init__(
+            **ray_tags
+        )
+
+    def _key_allowed_wrapper(self, key: str) -> bool:
+        return self._key_allowed(key, reserved_forbidden = False)
+
+
+def ray_remote(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
```
skrawcz, line 297: should this be `ray_resources` instead? Don't want to confuse the case when a function doesn't have this that people will think it's not being run via ray remote when it very well could be. 🤔

skrawcz (commented):
looking good. @elijahbenizzy might have more thoughts on the decorator than me.

hamilton/function_modifiers/metadata.py
elijahbenizzy: +1 for ray_resources.

elijahbenizzy: Also we're using lowercase classnames here -- breaking standard procedure but it'll be more consistent with the rest and remove a layer of indirection.

elijahbenizzy (commented):
Nice work! I'm in favor of putting this in `h_ray`, @skrawcz thoughts?

hamilton/function_modifiers/metadata.py
flavour: ray.remote options include things beyond resources.
A current list of possible options is here:
`
['max_calls', 'max_retries', 'num_cpus', 'num_returns', 'object_store_memory', 'retry_exceptions', '_generator_backpressure_num_objects', 'concurrency_groups', 'lifetime', 'max_concurrency', 'max_restarts', 'max_task_retries', 'max_pending_calls', 'namespace', 'get_if_exists', 'accelerator_type', 'memory', 'name', 'num_gpus', 'placement_group', 'placement_group_bundle_index', 'placement_group_capture_child_tasks', 'resources', 'runtime_env', 'scheduling_strategy', '_metadata']
`

flavour: RayRemote matches SchemaOutput, which is also standard CamelCase...this is what I was suggested to base upon.

elijahbenizzy: Ahh yep, my bad 😆 

And 👍 on `ray_remote_options`

hamilton/plugins/h_ray.py
```diff
@@ -184,7 +193,20 @@ def check_node_type_equivalence(node_type: typing.Type, input_type: typing.Type)
...
+        :param kwargs: the arguments that should be passed to it.
+        :return: returns a ray object reference.
+        """
+        # Tags are added to nodes via the @ray_remote_options decorator
+        ray_tags = {
+            tag_name: tag_value
+            for tag_name, tag_value in node.tags.items()
+            if tag_name.startswith(f"{RAY_REMOTE_TAG_NAMESPACE}.")
+        }
+        ray_options = {name.split(".", 1)[1]: json.loads(value) for name, value in ray_tags.items()}
```
skrawcz, line 208: can we pull this out into a function to unit test it?

Otherwise I don't see an easy integration test (that would live here https://github.com/DAGWorks-Inc/hamilton/tree/main/plugin_tests/h_ray) for this?

hamilton/plugins/h_ray.py
```diff
@@ -97,7 +99,14 @@ def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) ->
...
         :return: returns a ray object reference.
         """
-        return ray.remote(raify(node.callable)).remote(**kwargs)
+        # Tags are added to nodes via the @ray_remote_options decorator
+        ray_tags = {
+            tag_name: tag_value
+            for tag_name, tag_value in node.tags.items()
+            if tag_name.startswith(f"{RAY_REMOTE_TAG_NAMESPACE}.")
+        }
+        ray_options = {name.split(".", 1)[1]: json.loads(value) for name, value in ray_tags.items()}
```
skrawcz, line 108: ditto to use a common function that we can unit test.

hamilton/function_modifiers/metadata.py
```diff
@@ -271,3 +282,38 @@ def example_schema() -> pd.DataFrame:
...
+
+    def __init__(self, **options: Union[int, Dict[str, int]]):
+        """Initializes RayRemote. See docs for `@ray_remote_options` for more details."""
+
+        ray_tags = {f"ray_remote.{option}": json.dumps(value) for option, value in options.items()}
+
+        super(RayRemote, self).__init__(bypass_reserved_namespaces_=True, **ray_tags)
+
+
+def ray_remote_options(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
```
skrawcz, line 297: can we also get a unit test on this to check that the tags are created as expected please.

You can do something similar to `test_decorate_node_with_schema_output` in `test_metadata.py`.

skrawcz (commented):
Looks good. I think it'd be good to add some unit tests please and then this is ready to ship!

flavour:
> Looks good. I think it'd be good to add some unit tests please and then this is ready to ship!

Tests added & now passing :)

skrawcz (approved):

skrawcz:
Thanks @flavour 🚀  this will go out this coming week.

Issue #796, opened 2024-03-29
Error cleanly when no output requested using .materialize


skrawcz:
If no output is requested using .materialize() we don't error at all.

# Current behavior
It gets stuck in an infinite loop.

## Steps to replicate behavior
This recreates the behavior -- there is no output requested.
```python
# test.py
from hamilton import driver
from hamilton.plugins import h_experiments
from hamilton.io.materialization import to, from_

def connect(a:dict) -> str:
    print('connect')
    print(a)
    return 'connect'

if __name__ == "__main__":
    import test

    tracker_hook = h_experiments.ExperimentTracker(
        experiment_name='exp',
        base_directory='.',
    )


    dr = (
        driver.Builder()
        .with_modules(test)
        .enable_dynamic_execution(allow_experimental_mode=True)
        # .with_execution_manager(execution_manager)
        .with_adapters(tracker_hook)
        .build()
    )
    materilizers = [
        from_.pickle(path='./data.pickle', target='a'),
    ]
    dr.materialize(*materilizers)
```

## Library & System Information
Latest hamilton, mac.

# Expected behavior
It errors out.

# Additional context

PR #810, opened 2024-04-08
No Output Requested while materializer should lead to error

swapdewalkar:
Address [#796 ](https://github.com/DAGWorks-Inc/hamilton/issues/796) 

## Changes
Added a check to to throw Error if materializer are 0.
Added Test case for the same

## How I tested this
Unit Tested
Run Example on local as well
test_no_materializer

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1454,6 +1454,8 @@ class Driver:
         materializer_vars = []
         try:
             materializer_factories, extractor_factories = self._process_materializers(materializers)
+            if len(materializer_factories)==0:
+                raise ValueError("No output requested or materializer factories were provided.")
             function_graph = materialization.modify_graph(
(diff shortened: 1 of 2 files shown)


hamilton/driver.py
```diff
@@ -1454,6 +1454,8 @@ def materialize(
         materializer_vars = []
         try:
             materializer_factories, extractor_factories = self._process_materializers(materializers)
+            if len(materializer_factories) == len(final_vars) == 0:
+                raise ValueError("No output requested or materializer factories were provided.")
```
skrawcz, line 1458:
```suggestion
                raise ValueError("No output requested. Please either pass in materializers that will save data, or pass in `additional_vars` to compute.")
```

skrawcz (approved):
LGTM

PR #819, opened 2024-04-11
[feature] YAMLDataLoader and YAMLDataSaver as default data adapters

wmoreiraa:
Adding the option to do I/O using YAML files, similar as what we have with json.

## Changes

Adding YAML as default data adapter

## How I tested this
Same test module as others default data adapters, 3 unit tests:
1. Isolated load
2. Isolated save
3. Using both loader and saver.

## Notes
Followed pattern of json, but expanded the options of applicable_types

## Checklist

### DIFF
--- a/hamilton/io/default_data_loaders.py
+++ b/hamilton/io/default_data_loaders.py
@@ -6,8 +6,47 @@ import pathlib
 import pickle
 from typing import Any, Collection, Dict, Tuple, Type, Union
 
+import yaml
+
 from hamilton.io.data_adapters import DataLoader, DataSaver
 from hamilton.io.utils import get_file_metadata
+
+PrimitiveType = Union[str, int, bool, dict, list]
+
+
+@dataclasses.dataclass
+class YAMLDataLoader(DataLoader):
+    path: str
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return [str, int, bool, dict, list]
+
+    @classmethod
+    def name(cls) -> str:
+        return "yaml"
+
+    def load_data(self, type_: Type) -> Tuple[PrimitiveType, Dict[str, Any]]:
+        with pathlib.Path(self.path).open(mode="r") as f:
+            return yaml.safe_load(f), get_file_metadata(self.path)
+
+
+@dataclasses.dataclass
+class YAMLDataSaver(DataSaver):
+    path: str
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return [str, int, bool, dict, list]
+
+    @classmethod
+    def name(cls) -> str:
+        return "yaml"
+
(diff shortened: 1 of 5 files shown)


hamilton/io/default_data_loaders.py
```diff
@@ -6,9 +6,55 @@
...
+PrimitiveType = Union[str, int, bool, dict, list]
+
+
+@dataclasses.dataclass
+class YAMLDataLoader(DataLoader):
+    path: Union[str, pathlib.Path]
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return [str, int, bool, dict, list]
```
skrawcz, line 23: can replace with PrimititeType?

hamilton/io/default_data_loaders.py
```diff
@@ -6,9 +6,55 @@
...
+            return yaml.safe_load(f), get_file_metadata(path)
+
+
+@dataclasses.dataclass
+class YAMLDataSaver(DataSaver):
+    path: Union[str, pathlib.Path]
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return [str, int, bool, dict, list]
```
skrawcz, line 44: same here

skrawcz (changes requested):
Looks good!

One request: we should move this to the plugins folder. We don't have `yaml` as a core dependency, and I don't want to add it.

So we should move this to a file like:

`hamilton/plugins/yaml_extensions.py`

At the top it should have some code like:
```
try:
    import yaml
except ImportError:
    raise NotImplementedError("yaml is not installed.")
```

Then at the bottom it should have some additional code like:

```python
from hamilton import registry

def register_data_adapters():
    """Function to register the data loaders for this extension."""
    for loader in [
        YAMLDataLoader,
        YAMLDataSaver
    ]:
        registry.register_adapter(loader)


register_data_adapters()
```

and then in `hamilton/function_modifiers/base.py` add `yaml` to `plugins_modules` list so that we try to import it.

hamilton/io/default_data_loaders.py
wmoreiraa: Idk why I thought this would fail lol, thanks for the review.

wmoreiraa:
thanks for the review @skrawcz and totally agree! Nice to learn how the plugins works too :)

hamilton/io/default_data_loaders.py
skrawcz: Actually good point -- something to validate with an example. I could be wrong here.

skrawcz: actually I think I'm wrong here. What you had was correct.

wmoreiraa:
Fixed back the PrimitiveType issue. @skrawcz  lemme know if something still missing

skrawcz (approved):

skrawcz:
Thanks @wmoreiraa !

PR #820, opened 2024-04-12
dlt plugin

zilto:
The dlt library provides many `Sources` and `Destinations` to extract (E) and load (L) data.

With Hamilton being a transform (T) tool the goal is to build constructs to:
1. *ET*: Create Hamilton nodes from dlt Sources, providing a structured way for them to be upstream dependencies
2. *TL*: Store/materialize Hamilton node results into dlt Destinations.
These 2 features alone allow for the full flexibility of (ETL, ELTL, etc.)

For a good user experience, and more easily integrating "Hamilton within dlt", it is valuable to have:
3. *ETL*: essentially a combination of 1. and 2.
4. *ELT*: no special integration required here, (dlt does EL, then Hamilton does T), but the transition could be streamlined. We can showcase that via the `pipeline.sql_client()`. This is where dlt + Ibis integration would shine.

## Changes
- Added `DltResourceLoader` (1.)
- Added `DltDestinationSaver` materializer (2.)
- Scenario (3.) is covered by adding both Loaders and Savers.

## How I tested this
- DataSaver is tested for the 3 basic types accepted by dlt: iterables of records, pandas DataFrame, and pyarrow Table
- DataLoader tested for a mock Resource
- Given dlt pipelines are stateful, there's a possibility errors appear when using many Loader/Saver in a single project.

## Notes
- there's no easy mapping for dlt Source to a Hamilton construct. The `Source` has many `Resource` that would map to Hamilton nodes. Then `Resource` is closer to a  `DataLoader`, but the `Source` construct remains needed because it is responsible for authentication and more.

## Checklist


hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,74 @@
...
+        if isinstance(data, dict) is False:  # only 1 node, no `combine=base.DictResult()`
+            raise NotImplementedError(
+                f"DltSaver needs to a receive a `Dict[str, {SAVER_TYPES}]`"
+                f"When using `to.dlt()`, make sure to specify `combine=base.DictResult()`"
+            )
+
+        rename = {} if self.rename is None else self.rename
+
+        resources = []
+        for node_name, result in data.items():
```
elijahbenizzy, line 54: Hmm -- reason for having multiple per node? As opposed to one per node? More verbose but gets aroudn this problem.

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,74 @@
...
+        return SAVER_TYPES
+
+    def save_data(self, data) -> Dict[str, Any]:
+        """
+        ref:
+            tables: https://dlthub.com/docs/dlt-ecosystem/verified-sources/arrow-pandas
+            standalone resources: https://dlthub.com/docs/general-usage/resource#load-resources
+            dynamic resources: https://dlthub.com/docs/general-usage/source#create-resources-dynamically
+        """
+        if isinstance(data, dict) is False:  # only 1 node, no `combine=base.DictResult()`
```
elijahbenizzy, line 45: So this `data` type maps to the `applicable_types`. As in, this is the type that we can save.

The `combiner` will create a node that will have an output type of the corresponding result builder. E.G. the `DataFrameResult()`. So, this check shoudln't be necessary if `dict` is the only one. Also it means that none of the other ones should work.

That said, the ergonomics feel off... What are the use-cases behind the other types?

elijahbenizzy, line 45: nit -- `if not isinstance` is cleaner IMO

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,74 @@
...
+
+        rename = {} if self.rename is None else self.rename
+
+        resources = []
+        for node_name, result in data.items():
+            name = rename.get(node_name, node_name)
+            resources.append(dlt.resource(result, name=name))
+
+        load_info = self.pipeline.run(resources, write_disposition=self.write_disposition)
+        return load_info.asdict()
```
elijahbenizzy, line 59: Let's put this behind `dlt_data` or some field so we can add our own metadata as well

hamilton/plugins/h_dlt.py
```diff
@@ -0,0 +1,61 @@
...
+
+from hamilton import htypes
+from hamilton.graph_types import HamiltonGraph
+from hamilton.lifecycle import GraphExecutionHook
+from hamilton.plugins.dlt_extensions import DltSaver
+
+logger = logging.getLogger(__name__)
+
+
+class DltAdapter(GraphExecutionHook):
```
elijahbenizzy, line 14: I'm not seeing the value of this -- we really should have it part of the graph. If not, they should be totally separate, but this is burying critical side-effects pretty deep down...

skrawcz: I think it's an interesting alternative -- definitely implicit. But something to get feedback on.

🤔  we could also model this as a result builder.

hamilton/plugins/dlt_extensions.py
zilto: what metadata should we return given we don't have a file path to check for

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,128 @@
...
+from hamilton.io.data_adapters import DataLoader, DataSaver
+
+DATAFRAME_TYPES = [Iterable, pd.DataFrame]
+
+# TODO add types for other Dataframe libraries
+try:
+    import pyarrow as pa
+
+    DATAFRAME_TYPES.extend([pa.Table, pa.RecordBatch])
+except ModuleNotFoundError:
```
elijahbenizzy, line 23: So this works, the problem is that the error message will be confusing... For now, however, I think we can just document it well.

hamilton/plugins/dlt_extensions.py
elijahbenizzy: Meh no idea, but it fits in with the rest of the pattern if we have it namespaced. E.G. this: https://github.com/DAGWorks-Inc/hamilton/blob/5dd6ab4bf8933aeb71b78e1c28045afc7e46623c/hamilton/io/utils.py#L42

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,128 @@
...
+        normalize_info = pipeline.normalize(loader_file_format="parquet")
+
+        partition_file_paths = []
+        for package in normalize_info.load_packages:
+            load_info = package.jobs["new_jobs"][0]
+            partition_file_paths.append(load_info.file_path)
+
+        # TODO use pyarrow directly to support different dataframe libraries
+        # ref: https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetDataset.html#pyarrow.parquet.ParquetDataset
+        df = pd.concat([pd.read_parquet(f) for f in partition_file_paths], ignore_index=True)
```
elijahbenizzy, line 58: Parquet locally versus duckdb? Will the next `pipeline.drop()` erase it?

elijahbenizzy (commented):
Looking good! Most curious about the parquet thing -- wonder if we can use duckdb:memory...

hamilton/plugins/dlt_extensions.py
skrawcz: We should return metadata that could help us identify/understand this artifact better. It might be the same that's in the dict? or some subset of that?

hamilton/plugins/dlt_extensions.py
zilto: Why would we want to throw an error? You mean if a node has annotions for `pa.Table`, but pyarrow is not installed and therefore finds no registered materializers?

hamilton/plugins/dlt_extensions.py
zilto: So the dlt implementation for in-memory duckdb is currently bugged (they're fixing it), but I found a work around.

Using duckdb in-memory doesn't really provide additional value and only adds a dependency:
```
source -> extract (parquet) -> normalize (parquet) -> load (parquet) -> duckdb (memory) -> query db (memory) -> to pandas (memory)
```
At the end of the process, memory (duckdb, pandas) is freed and dlt pipeline is cleaned (extract, normalize, load)

My current implementation skips the duckdb steps
```
source -> extract (parquet) -> normalize (parquet) -> read parquet partitions (memory) -> pandas (memory)
```

zilto: The related design decision is "how can I selectively reload dlt Sources". For example, only load Slack messages once then run Hamilton dataflows many times over that same data. 

This would be an ELT use case (dlt -> Hamilton) where you want to refresh each independently. It probably makes more sense to have `run_dlt.py` and `run_hamilton.py`

The current "Source" materializer with everything in-memory aims to enable ET and the user is responsible for loading the data, potentially with the Destination materializer which does TL

docs/integrations/dlt/index.md
```diff
@@ -289,6 +294,114 @@ results = dr.execute(
...
+        target="general_messages",  # node name assigned to the data
+        resource=source.resources["general_messages"]
+    ),
+    from_.dlt(
+        target="general_replies_message",
+        resource=source.resources["general_replies_message"]
+    ),
+]
+
+dr.materialize(*materializers, ...)
```
skrawcz, line 331: need additional_vars here since no `to.*` materialize is provided.
```suggestion
dr.materialize(*materializers, additional_vars=[...], ...)
```

docs/integrations/dlt/index.md
```diff
@@ -289,6 +294,114 @@ results = dr.execute(
...
+    ),
+    to.dlt(
+        id="threads__dlt",
+        dependencies=["threads"],
+        table_name="slack_threads",
+        pipeline=pipeline,
+    )
+]
+
+dr.materialize(*materializers)
```
skrawcz, line 401: show a picture of this!

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,133 @@
...
+try:
+    import pyarrow as pa
+
+    DATAFRAME_TYPES.extend([pa.Table, pa.RecordBatch])
+except ModuleNotFoundError:
+    pass
+
+# convert to tuple to dynamically define type `Union[DATAFRAME_TYPES]`
+DATAFRAME_TYPES = tuple(DATAFRAME_TYPES)
+
```
skrawcz, line 28:
```suggestion
COLUMN_FRIENDLY_DF_TYPE = False
```

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,133 @@
...
+# TODO add types for other Dataframe libraries
+try:
+    import pyarrow as pa
+
+    DATAFRAME_TYPES.extend([pa.Table, pa.RecordBatch])
+except ModuleNotFoundError:
+    pass
+
+# convert to tuple to dynamically define type `Union[DATAFRAME_TYPES]`
+DATAFRAME_TYPES = tuple(DATAFRAME_TYPES)
```
skrawcz, line 27: where is this used?

hamilton/plugins/dlt_extensions.py
skrawcz: can just cache the result of this function if we don't want it to run more than once?

hamilton/plugins/dlt_extensions.py
```diff
@@ -0,0 +1,133 @@
...
+def register_data_loaders():
+    """Function to register the data loaders for this extension."""
+    for loader in [
+        DltDestinationSaver,
+        DltResourceLoader,
+    ]:
+        registry.register_adapter(loader)
+
+
+register_data_loaders()
```
skrawcz, line 133: this needs to be added to the function_modifiers/base.py bit to try to register this plug in automatically.

hamilton/plugins/dlt_extensions.py
elijahbenizzy: So yeah, the worry is that it fills up disk space. But I think we can switch this up as needed. Seems reasonable.

hamilton/plugins/dlt_extensions.py
elijahbenizzy: Yeah, exactly. Its just a confusing case, but let's not worry about it (maybe add to docs).

elijahbenizzy (approved):
Let's ship, see where it goes from here!

PR #826, opened 2024-04-16
docs: "user guide" section cleanup

zilto:
## Changes
- Renamed the section from "How To Guides" to "User Guide".
- Changed the title of "User Guide" pages to be easier to visually scan. Essentially, this consists of shorter titles with the most important word at the beginning
- changed the order of user guide items
- updated the list of integrations

Before:
image

After:
image

## How I tested this
- build docs locally

## Checklist


docs/how-tos/custom-driver.rst
skrawcz, line 1: why the deletion?

docs/how-tos/async.rst
```diff
@@ -1,5 +1,5 @@
 ==============================
-Run Hamilton in a Microservice
+Online transformations
```
skrawcz, line 2: I think micro-service is better here than online transformation. The main point to take away is that you can run Hamilton in a web-service.

skrawcz, line 2:
```suggestion
Micro-service
```

docs/how-tos/custom-driver.rst
skrawcz: This could be renamed to "Custom Driver"

docs/how-tos/load-data.rst
```diff
@@ -1,12 +0,0 @@
-==================
```
skrawcz, line 1: even though this section isn't great, I think a `loading data` guide makes sense that points to examples/concepts (and even the recording from today).

docs/how-tos/use-without-pandas.rst
```diff
@@ -1,14 +0,0 @@
-===========================
```
skrawcz, line 1: fine to delete.

skrawcz:
To me the purpose of the `how-to` section is for goal oriented people - "I want to do X with Hamilton"  where X is a broad concept -- for integrations X is a specific narrower technology/topic. 

So I think the changes mostly look good. Left comments where appropriate.

zilto:
> To me the purpose of the `how-to` section is for goal oriented people - "I want to do X with Hamilton" where X is a broad concept -- while the integrations section is when X is a specific narrower technology/topic.

I agree on the purpose of the "how-to" / "user guide" section, but I think a lot of is too introductory / generic. In that case, it should reframed to be part of the "Concepts" section. The few good pages include "feature engineering", "pre-commit", "jupyter notebooks".

For instance, I believe "User guide" pages should mirror frequent Q/A on Slack

skrawcz (approved):

PR #828, opened 2024-04-17
add a "resolve from config" wrapper around the resolve decorator

janhurst:
Add a small wrapper around `resolve` function modifier to "resolve from config"

## Changes

Add a `hamilton.function_modifiers.resolve_from_config` hook that just passes `ResolveAt.CONFIG_AVAILABLE` flag to the original `resolve` class.

## How I tested this

Have been using this as a small custom snippet (albeit not as a wrapper around the class) in some internal projects for quite a while. Its really just a convenience wrapper to make my pipeline code a little cleaner.

I've made a very small attempt to duplicate a relevant unit test.

## Checklist

### DIFF
--- a/hamilton/function_modifiers/delayed.py
+++ b/hamilton/function_modifiers/delayed.py
@@ -151,5 +151,15 @@ class resolve(DynamicResolver):
 
 
 class resolve_from_config(resolve):
+    """Decorator class to delay evaluation of decorators until after the configuration is available.
+    Note: this is a power-user feature, and you have to enable power-user mode! To do so, you have
+    to add the configuration hamilton.enable_power_user_mode=True to the config you pass into the
+    driver.
+
+    This is a convenience decorator that is a subclass of `resolve` and passes
+    `ResolveAt.CONFIG_AVAILABLE` to the `when` argument such that the decorator is resoled at
+    compile time, E.G. when the driver is instantiated.
+    """
+
     def __init__(self, *, decorate_with: Callable[..., NodeTransformLifecycle]):
         super().__init__(when=ResolveAt.CONFIG_AVAILABLE, decorate_with=decorate_with)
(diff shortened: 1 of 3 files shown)


hamilton/function_modifiers/delayed.py
```diff
@@ -148,3 +148,8 @@ def resolve(self, config: Dict[str, Any], fn: Callable) -> NodeTransformLifecycl
             if key in config:
                 kwargs[key] = config[key]
         return self.decorate_with(**kwargs)
+
+
+class resolve_from_config(resolve):
```
elijahbenizzy, line 153: Thought from @zilto  -- would `@resolve_at_config` be cleaner?

Also, let's add a docstring then stick it in the decorators documentation.

See https://github.com/DAGWorks-Inc/hamilton/blob/main/docs/reference/decorators/resolve.rst -- should be easy!

elijahbenizzy (commented):
Looking good! Minor change for docstring (and maybe name) then let's merge.

hamilton/function_modifiers/delayed.py
zilto: I think `@resolve_at_config` better capture that it's a subset of `@resolve` where we set the value of `ResolveAt` to `ConfigAvailable`.

There's no other `ResolveAt` value, but I think the semantic make sense. I believe the name `@resolve_at_config` is an improvement over `@resolve` as it better communicate it's role during the Driver building.

skrawcz: 🤔  just to voice my opinion that `resolve_from_config` doesn't need you to understand what the driver does, but it does help more clearly tie the connection that it comes from configuration passed in. Which I think for a new person to Hamilton that might be more intuitive, than `resolve_at_config` since now they need to understand what "at" might be referring to.

@janhurst thoughts on the name?

janhurst:
> Looking good! Minor change for docstring (and maybe name) then let's merge.

added a little bit of a redirect to the `@resolve` docs... feedback greatly appreciated

skrawcz (approved):
This works for me.

Issue #829, opened 2024-04-17
Adapter mapping types to pyarrow


zilto:
## Problem
The goal is to facilitate the materialization of Hamilton nodes. Currently, we need to develop and maintain a materializer for each library (pandas, polars, dask, vaex, ibis, etc.). 

More importantly, writing materializers like `dlt`, which supports pyarrow but not polars, for example, is challenging. It requires many `try/except` to import packages and duplicating DataSaver code to register the proper implementation based on successfully imported modules.

From the user perspective, it would require converting polars objects to pyarrow within the Hamilton function to be compatible with the materializer, which defeats the purpose of materializers (decoupling dataflows from I/O)

## Solution
By relying on pyarrow and the [Dataframe Interchange Protocol](https://arrow.apache.org/docs/python/interchange_protocol.html), we could have a step after the execution, but before materialization that convert polars et al. to pyarrow. Then, these converted nodes would be compatible with pyarrow materializers.

It enables annotating functions with polars, pandas, etc., but keeping the same pyarrow materializer. Users can use their preferred library while the platform team can maintain and track a single set of materializers

PR #830, opened 2024-04-17
added `PyarrowTableResult` 

zilto:
You can pass `to.SAVER(dependencies=["NODE_NAME"], combine=PyarrowTableResult())` to convert the specified node to the `pyarrow.Table` before materialization. The first motivation was to support more than `pd.DataFrame` and `pyarrow.Table` with the dlt `DataSaver` plugin. More generally, it can be useful for platform teams that want to have a "single way to store parquet files" that is independent of the specific API of a library (e.g., pandas, polars)

see #829 for more details

## Changes
- added `h_pyarrow` and tests
- updated the dlt plugin example notebook 

## How I tested this
- added 2 tests

## Checklist


hamilton/plugins/h_pyarrow.py
```diff
@@ -0,0 +1,39 @@
...
+
+    It implicitly support input_type == Any, but it expects dataframe objects implementing
+    the dataframe interchange protocol: ref: https://arrow.apache.org/docs/python/interchange_protocol.html
+    for example:
+    - pandas
+    - polars
+    - dask
+    - vaex
+    - ibis
+    - duckdb results
```
skrawcz, line 21: it would be nice to be stricter on types -- but I suppose this is fine.

skrawcz:
```python
def input_types(self) -> List[Type[Type]]:
    """Gives the applicable types to this result builder.
    This is optional for backwards compatibility, but is recommended.

    :return: A list of types that this can apply to.
    """
    _types = []
    try:
       import ...
   except ...
    return _types
```

skrawcz (approved):

hamilton/plugins/h_pyarrow.py
zilto: I couldn't figure a way to annotate type without a series of `try/except` statements. After feedback on the dlt extension, I thought this pattern undesirable, so I opted to not include `input_types()` if it was to return `Any`

PR #846, opened 2024-04-25
Jupyter magics v2

zilto:
Please read the tutorial notebook for all details.

## TODO
- do some light refactoring to allow a barebone `%%execute` cell magic
- convert the jupyter notebook to markdown and add it to docs

## Changes
- this currently overrides the previous jupyter magics and isn't backward compatible
- To make things backward compatible, we can use a different filename `jupyter_extensions` to be consistent with other extensions? Nothing stops this file from containing other things than "magics". This matters because it's what users need to import via `%load_ext hamilton.plugins.jupyter_extensions` whereas "magic" is probably an odd term for most people

## How I tested this
- it's hard to test given the "user interface" nature of the tool. Most of it is parsing code cells + dynamic module importing + standard Hamilton
- the main new feature is `rebuild_driver()`

## Notes
- I see to get performance hit after 50+ cell execution (maybe my computer is behaving oddly). This maybe because each time we create an ad hoc module, we add it to `sys.modules`? Maybe it's just VSCode + Pylance LSP disliking magic code cells. Don't know

## Checklist


hamilton/plugins/jupyter_magic.py
```diff
@@ -14,26 +14,118 @@
...
+    """Execute source code from a cell in the user namespace and collect
+    the values of all assigned variables into a dictionary.
+    """
+    shell.ex(cell)
+    expr = shell.input_transformer_manager.transform_cell(cell)
+    expr_ast = shell.compile.ast_parse(expr)
+    return {name: shell.user_ns[name] for name in get_assigned_variables(expr_ast)}
+
+
+def rebuild_driver(
```
elijahbenizzy, line 78: Add docstring + notes that this is using internal APIs/we'll want to replace with a more official pattern

hamilton/plugins/jupyter_magic.py
```diff
@@ -14,26 +14,118 @@
...
+    _config = config if config else _driver.graph.config
+    _modules = modules if modules else _driver.graph_modules
+    _adapter = adapters if adapters else _driver.adapter
+    _graph_executor = graph_executor if graph_executor else _driver.graph_executor
+
+    if reload_modules:
+        new_modules = []
+        for m in _modules:
+            try:
+                new_module = importlib.reload(m)
```
elijahbenizzy, line 98: Does this work with the dynamic modules?

elijahbenizzy: My guess is it does but worth some comments as to how it interacts.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +187,184 @@ def determine_notebook_type() -> str:
...
+        """Execute the cell and store all assigned variables as overrides"""
+        self.overrides = execute_and_get_assigned_values(self.shell, cell)
+
+    @cell_magic
+    def set_display_config(self, line: str, cell: str):
+        self.display_config = execute_and_get_assigned_values(self.shell, cell)
 
     @line_magic
-    def insert_module(self, line):
+    def module_to_cell(self, line):
```
elijahbenizzy, line 367: `module_to_cell` versus `cell_to_module` in docs

elijahbenizzy (commented):
Not 100% sure the compatibility case. It's backwards-incompatible, right? If so, are we OK with that?

examples/jupyter_notebook_magic/README.md
```diff
@@ -1,25 +1,27 @@
-# This example shows a notebook using the Hamilton Jupyter magic
+# Hamilton notebook extension
+
+One of the best part about notebooks is the ability to execute and immediately inspect results. They provide a "read-eval-print" loop (REPL) coding experience. However, the way Hamilton separates dataflow definition (functions in a module) from execution (building and executing a driver) creates an extra step that can slowdown this loop.
```
skrawcz, line 3:
```suggestion
One of the best part about notebooks is the ability to execute and immediately inspect results. They provide a "read-eval-print" loop (REPL) coding experience. However, the way Hamilton separates dataflow definition (functions in a module) from execution (building and executing a driver) creates an extra step that creates a little friction with this loop.
```

hamilton/plugins/jupyter_magic.py
```diff
@@ -87,74 +179,6 @@ def insert_cell_with_content():
...
-    impacted_drivers = []
-    for var_name, dr in driver_instances.items():
-        for driver_module in dr.graph_modules:
-            if driver_module.__name__ == module_name:
-                impacted_drivers.append((var_name, dr))
-                break
-    return impacted_drivers
-
-
-def rebuild_drivers(shell, module_name: str, module_object: ModuleType, verbosity: int = 1) -> dict:
```
skrawcz, line 111: what was the issue with this approach?

skrawcz (commented):
yeah so I like some of the changes, but not others.

For example the departure from having to provide valid driver builder code is a concern because I think that makes the experience confusing from regular Hamilton. It's also not clear where the modules comes from that go into the driver. E.g. What if I have multiple modules and want several different drivers?

Let's chat about it some more.

examples/jupyter_notebook_magic/example.ipynb
skrawcz, line 1: Do we link to this file from the blogs?

hamilton/plugins/jupyter_magic.py
zilto: Currently an unused feature, but added it since it came up to mind. The use case:

- User is writing dataflow in notebook
- User imports `module_x` and dynamics modules
- User updates `module_x` and wants it updated in the notebook without restarting the kernel.

User could manually `importlib.reload()` and rebuild the Driver, but we could automate it

hamilton/plugins/jupyter_magic.py
zilto: I want to have the Jupyter magic owning the `Driver`

examples/jupyter_notebook_magic/example.ipynb
zilto: Will create a new `example.ipynb` based on the recent RAG PR once we settle on the Jupyter V2 API

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
-            print("Help for %%cell_to_module magic:")
-            print("  -m, --module_name: Module name to provide. Default is jupyter_module.")
-            print("  -c, --config: JSON config string, or variable name containing config to use.")
-            print("  -r, --rebuild-drivers: Flag to rebuild drivers.")
-            print("  -d, --display: Flag to visualize dataflow.")
-            print("  -v, --verbosity: of standard output. 0 to hide. 1 is normal, default.")
-            return  # Exit early
-        if not hasattr(self, "notebook_env"):
-            # doing this so I don't have to deal with the constructor
-            self.notebook_env = determine_notebook_type()
```
skrawcz, line 201: why remove this?

hamilton/plugins/jupyter_magic.py
```diff
@@ -14,26 +14,125 @@
...
+    graph_executor: Optional[driver.GraphExecutor] = None,
+    reload_modules: Literal[True, False, "strict"] = False,
+) -> driver.Driver:
+    _driver = dr if dr else driver.Builder().build()
+    _config = config if config else _driver.graph.config
+    _modules = modules if modules else _driver.graph_modules
+    _adapter = adapters if adapters else _driver.adapter
+    _graph_executor = graph_executor if graph_executor else _driver.graph_executor
+
+    if reload_modules:
```
skrawcz, line 101: this should really be a separate function.

hamilton/plugins/jupyter_magic.py
skrawcz: I don't think we need this function given we're passed a builder.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
 
 @magics_class
 class HamiltonMagics(Magics):
-    """Magics to facilitate Hamilton development in Jupyter notebooks"""
+    """Magics to facilitate interactive Hamilton development in notebooks."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.driver = None
+        self.driver_name = None
```
skrawcz, line 202: can remove these two, right?

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
-        module_object = create_module(cell, module_name)
-
-        # shell.push() assign a variable in the notebook. The dictionary keys are variable name
+                file_path = Path(f"{module_name}.py")
+            file_path.write_text(cell)
+
+        # create modules and build driver
+        module_object = ad_hoc_utils.module_from_source(cell, module_name)
         self.shell.push({module_name: module_object})
+        base_dr = builder.build()  # easier to rebuild a Driver than messing with Builder
```
skrawcz, line 293: 🤔  what's the difficulty? You do need to copy it -- which is something that could be easily added.

```python
base = builder.copy()
base.with_modules(module_object) # this will add to whatever modules were there
# reload any other modules here.
dr = base.build()
```

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
+        dr = rebuild_driver(dr=base_dr, modules=[*base_dr.graph_modules, module_object])
 
-        # shell.user_ns is a dictionary of all variables in the notebook
-        # rebuild drivers that use this module
-        if args.rebuild_drivers:
```
skrawcz, line 231: I think this should be left in to keep things backwards compatible.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
+        self.external_modules = list()
+        self.final_vars = list()  # empty list will be converted to "all variables"
+
+        if not hasattr(self, "notebook_env"):
+            self.notebook_env = determine_notebook_type()
 
     @magic_arguments()  # needed on top to enable parsing
+    @argument("module_name", help="Name for the module defined in this cell.")
     @argument(
-        "-m", "--module_name", help="Module name to provide. Default is jupyter_module."
```
skrawcz, line 170: we need `-m` support for backwards compatibility, right?

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
+    )
     @argument(
-        "-v", "--verbosity", type=int, default=1, help="0 to hide. 1 is normal, default"
-    )  # keyword / optional arg
+        "-o",
+        "--overrides",
+        help="Execution overrides. The argument is the variable name of a dict of overrides; else {}.",
+    )
+    @argument(
+        "-h",
```
skrawcz, line 242: I think most people will think `-h` is help?

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +194,160 @@ def determine_notebook_type() -> str:
...
-    """Magics to facilitate Hamilton development in Jupyter notebooks"""
+    """Magics to facilitate interactive Hamilton development in notebooks."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.driver = None
+        self.driver_name = None
+        self.builder = None
+        self.external_modules = list()
+        self.final_vars = list()  # empty list will be converted to "all variables"
```
skrawcz, line 205: and remove these?

hamilton/plugins/jupyter_magic.py
zilto: We do. I applied the changes you requested for the user-facing API to pass a Builder instead of a Driver. Then, behind the scene, I do
```python
temp_module = module_from_source(cell)
dr = builder.build()
dr = rebuild_driver(dr, temp_module)   # exact API doesn't matter
```
 Otherwise, working with Builder directly is a mess:
 ```python
 temp_module = module_from_source(cell)
 dr = builder.with_modules(temp_module).build()
 ```
 It leads to a ton of hard to track bug because `builder` is stateful and comes from the user namespace. Therefore, it keeps reference to old copies of `temp_module`, which need to be manually removed and result in many `function foo already defined` errors

hamilton/plugins/jupyter_magic.py
zilto: As far as I can tell `Builder.copy()` doesn't exist? Am I missing something

hamilton/plugins/jupyter_magic.py
zilto: I'll remove the short form `-h` and keep `--hide_result`. I wanted a different word family from "display, view, visualize" since we use those for viewing the DAG in other contexts.

hamilton/plugins/jupyter_magic.py
zilto: There's a bit of a conflict here. I believe `module_name` should be a required argument, e.g. `%%cell_to_module joke`. Setting a name requires close to 0 efforts and improves readability.

Then, keeping `--module_name / -m` raises issues:
- your cell will fail anyway because it's missing the required arg. `%%cell_to_module -m joke` won't work.
- having both the required arg and `--module_name` is confusing. `%%cell_to_module something -m joke` is ambiguous

hamilton/plugins/jupyter_magic.py
zilto: Wouldn't it be confusing to preserve the arg but changing it's behavior?

Regarding migration (this and `--module_name`), a user of the old magic would get:
```python
%%cell_to_module -m joke -r
def reply(joke_prompt: str) -> str:
    _, _, right = joke_prompt.partition("? ")
    return f"{right} who?"
```
Error
```console
UsageError: unrecognized arguments: -m -r
```
This would prompt them to do `%%cell_to_module?` and get the updated docs.

As a change, I could allow the old args and do:
```python
if any(args.rebuild_driver, args.module_name, ...):
   raise ValueError("Arguments ... are deprecated. Use `%%cell_to_module?` for updated documentation")
```

hamilton/plugins/jupyter_magic.py
skrawcz: we can't push a backwards incompatible API change.

skrawcz (commented):
yeah so out of principal we don't push backwards incompatible changes.

Options:

- rename the magic.
- live with `-m` ... 
- have a really good error message and suggest the fix and apologize and ask for forgiveness... (is this possible to do well?)

skrawcz:
From chatting:

1. restart from main.
2. we want to make sure things are backwards compatible -- add deprecation warnings/etc.

examples/jupyter_notebook_magic/README.md
```diff
@@ -1,25 +1,27 @@
...
 ```python
-# load some extensions / magic...
 %load_ext hamilton.plugins.jupyter_magic
 ```
 
-Then to use it:
+For example, this would allow you to define the module `joke` from your notebook
 
 ```python
-%%cell_to_module -m MODULE_NAME # more args
```
skrawcz, line 12: should have output of `?cell_to_module` instead.

hamilton/ad_hoc_utils.py
```diff
@@ -60,9 +60,9 @@ def create_temporary_module(*functions: Callable, module_name: str = None) -> Mo
     return module
 
 
-def module_from_source(source: str) -> ModuleType:
+def module_from_source(source: str, module_name: str = None) -> ModuleType:
```
skrawcz, line 63: need to rebase I guess?

hamilton/driver.py
```diff
@@ -1810,6 +1810,20 @@ def build(self) -> Driver:
             _use_legacy_adapter=False,
         )
 
+    def copy(self) -> "Builder":
```
skrawcz, line 1813: need to rebase I guess?

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +154,154 @@ def determine_notebook_type() -> str:
...
-    """Magics to facilitate Hamilton development in Jupyter notebooks"""
+    """Magics to facilitate interactive Hamilton development in notebooks."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.builder = None
+        self.notebook_env = determine_notebook_type()
 
     @magic_arguments()  # needed on top to enable parsing
+    @argument("module_name", help="Name for the module defined in this cell.")
```
skrawcz, line 165: I think we agreed on this being okay if we can set a default & then having `-m` and `--module` back.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +154,154 @@ def determine_notebook_type() -> str:
...
-        "-m", "--module_name", help="Module name to provide. Default is jupyter_module."
-    )  # keyword / optional arg
+        "-x",
+        "--execute",
+        nargs="?",
+        const=True,
+        help="Execute the dataflow. The argument is the variable name of a list of nodes; else execute all nodes.",
+    )
     @argument(
-        "-c", "--config", help="JSON config, or variable name containing config to use."
```
skrawcz, line 173: we should keep this for backwards compatibility. 
If someone provides a builder we should error, or replace the builder's config with this.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +154,154 @@ def determine_notebook_type() -> str:
...
-    )  # keyword / optional arg
+        "-b",
+        "--builder",
+        help="Builder to which the module will be added and used for execution. Allows to pass Config and Adapters",
+    )
     @argument(
-        "-r", "--rebuild-drivers", action="store_true", help="Flag to rebuild drivers"
```
skrawcz, line 176: I think this is okay to remove -- we assume no one really is using this.

hamilton/plugins/jupyter_magic.py
```diff
@@ -163,95 +154,154 @@ def determine_notebook_type() -> str:
...
+    )
     @argument(
-        "-d", "--display", action="store_true", help="Flag to visualize dataflow."
-    )  # Flag / optional arg
+        "-o",
+        "--overrides",
+        help="Execution overrides. The argument is the variable name of a dict of overrides; else {}.",
+    )
     @argument(
-        "-v", "--verbosity", type=int, default=1, help="0 to hide. 1 is normal, default"
```
skrawcz, line 182: we might want to keep this for backwards compatibility?

zilto:
Closing in favor of #917

PR #847, opened 2024-04-25
`fix` closed `CacheAdapter`

zilto:
The `CacheAdapter()` would successfully run for the first execution, and fail on subsequent ones. This is because the cache was opened on `__init__()` (needs to read some attributes at instantiation) and closed "after graph execution".

## Changes
cache is now opened and closed at `__init__()`
cache is opened "before graph execution" and closed "after graph execution" for satefy

### DIFF
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -348,9 +348,11 @@ class CacheAdapter(NodeExecutionHook, NodeExecutionMethod, GraphExecutionHook):
             key=CacheAdapter.nodes_history_key, default=dict()
         )
         self.used_nodes_hash: Dict[str, str] = dict()
+        self.cache.close()
 
     def run_before_graph_execution(self, *, graph: HamiltonGraph, **kwargs):
         """Set `cache_vars` to all nodes if received None during `__init__`"""
+        self.cache = shelve.open(self.cache_path)
         if self.cache_vars == []:
             self.cache_vars = [n.name for n in graph.nodes]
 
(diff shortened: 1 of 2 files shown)


hamilton/lifecycle/default.py
```diff
@@ -348,9 +348,11 @@ def __init__(
             key=CacheAdapter.nodes_history_key, default=dict()
         )
         self.used_nodes_hash: Dict[str, str] = dict()
+        self.cache.close()
 
     def run_before_graph_execution(self, *, graph: HamiltonGraph, **kwargs):
         """Set `cache_vars` to all nodes if received None during `__init__`"""
+        self.cache = shelve.open(self.cache_path)
```
elijahbenizzy, line 355: Something feels wrong about this. Aren't we just opening it again? Feels like it should:
1. Open during `init`
2. Run during execution
3. Close at the end

elijahbenizzy (commented):
Not 100% sure this is the right solution...

hamilton/lifecycle/default.py
zilto: well, that was the previous behavior... I want to have `.close()` *somewhere* for safety, but putting it in `run_after_graph_execution` means you can only run it once

skrawcz: I think that's what Thierry just did here?

elijahbenizzy: Ahh yes, my bad

elijahbenizzy (approved):
Ship it! Sorry misunderstood a bit. This looks good.

PR #855, opened 2024-04-26
Changes jupyter magic to create temporary files

skrawcz:
Temporary files allow us to then hash the code properly for use with the Hamilton UI.

This does not change any functionality. If the interpreter is shutdown gracefully then the temporary files are deleted cleanly.

## Changes
 - adjust function that creates the module in the jupyter magic

## How I tested this
 - locally
 - on databricks

## Notes
 - this means magics in notebooks can be cleanly used with the Hamilton UI.

## Checklist

### DIFF
--- a/hamilton/ad_hoc_utils.py
+++ b/hamilton/ad_hoc_utils.py
@@ -1,7 +1,11 @@
 """A suite of tools for ad-hoc use"""
 
+import atexit
+import importlib.util
 import linecache
+import os
 import sys
+import tempfile
 import types
 import uuid
 from types import ModuleType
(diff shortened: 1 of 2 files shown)


elijahbenizzy (commented):
Move this to ad-hoc utils?

zilto:
I commented out one check from the sdk code, and the already implemented `module_from_source` properly behaves for logging in the UI.


Commenting out ([source](https://github.com/DAGWorks-Inc/hamilton/blob/5296232e8c8d6dac88fbe8ce01e89d850c8dd0be/ui/sdk/src/hamilton_sdk/driver.py#L71)
```python
    for name, value in inspect.getmembers(module):
        # Check if the attribute is a module
        if inspect.ismodule(value):
            if value.__package__ is None:
                logger.info(
                    f"Skipping hash for module {value.__name__} because it has no __package__ "
                    f"attribute or it is None. This happens with lazy loaders."
                )
                continue
            # Check if the module is in the same top level package
            # if value.__package__ != module.__package__ and not value.__package__.startswith(
            #     module.__package__
            # ):
            #     logger.debug(
            #         f"Skipping hash for module {value.__name__} because it is in a different "
            #         f"package {value.__package__} than {module.__package__}"
            #     )
            #     continue
            
            # Recursively hash the sub-module
            hash_object = _hash_module(value, hash_object, seen_modules)

    # Return the hash object
    return hash_object
```

PR #883, opened 2024-05-03
Adds incremental module creator to magic

skrawcz:
The idea is simple. Allow one to build up a module easily and quickly. This is useful when explaining a concept.

Example use:
```
%%incr_cell_to_module -m doc_pipeline -i 5 --display
```
The idea is that you can partition creating a module over multiple cells. The ONLY CAVEAT is that you maintain a unique identifier (int or string) for each cell so that they can are distinct; i.e. it's possible to override an earlier cell if you use the same identifier. Note: we use the sorted set of identifiers and build the module up in order of that. This is probably an unused feature, since we expect most people to logically increment the identifier and order things correctly top to bottom.

There are some utility methods to help reset state. Otherwise I duplicate some logic, since it wasn't easy to refactor. I centralized logic where it made sense.

## Changes
 -  adds new `incr_cell_to_module` cell magic.
 -  adds new `print_module` line magic to print what's created
 -  adds new `reset_module` line magic to reset an incremental module.

## How I tested this
- locally

## Notes
- the incremental path is 

## Checklist

### DIFF
--- a/hamilton/plugins/jupyter_magic.py
+++ b/hamilton/plugins/jupyter_magic.py
@@ -21,7 +21,7 @@ from types import ModuleType
 
 from IPython.core.magic import Magics, cell_magic, line_magic, magics_class
 from IPython.core.magic_arguments import argument, magic_arguments, parse_argstring
-from IPython.display import HTML, display
+from IPython.display import HTML, Code, display
 
 from hamilton import ad_hoc_utils, driver, lifecycle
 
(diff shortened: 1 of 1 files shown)


hamilton/plugins/jupyter_magic.py
```diff
@@ -260,6 +248,157 @@ def insert_module(self, line):
         # insert source code as text in the next cell
         self.shell.set_next_input(header + module_source, replace=False)
 
+    @magic_arguments()  # needed on top to enable parsing
+    @argument(
```
elijahbenizzy, line 252: Better to use a deafult value here? See docs: https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.magic_arguments.html.

elijahbenizzy (approved):
Looks good, one small comment but its not introduced here

hamilton/plugins/jupyter_magic.py
zilto: we'll refactor to match jupyter v2 once we're there

PR #884, opened 2024-05-03
Remove temporary DOT files

zilto:
Hamilton visualizations rely on the graphviz library.
It defines graphs using the DOT language, which
defines one statement per line using a string.

Previously, Hamilton used `graphviz.Digraph.render()`
to produce visualizations. This has the side-effect
of producing an intermediary DOT file on disk. This
is most often of no use and clutters the directory.

Now, we are switching to `graphviz.Digraph.pipe()`
to write bytes directly to an open file. Tests were
updated accordingly.

The keyword argument `keep_dot` was added to viz
functions in case users still want this DOT file
to be produced. It allows to rerender the viz with
a different style without re-executing the Hamilton
code. It could be useful when iterating over custom
styling.

## How I tested this
- updated tests and they succeed

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -9,6 +9,7 @@ Note: one should largely consider the code in this module to be "private".
 import inspect
 import logging
 import os.path
+import pathlib
 import uuid
 from enum import Enum
 from types import ModuleType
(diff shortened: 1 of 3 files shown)


hamilton/graph.py
```diff
@@ -894,7 +899,7 @@ def display(
             custom_style_function=custom_style_function,
             config=config,
         )
-        kwargs = {"view": False, "format": "png"}  # default format = png
+        kwargs = {"format": "png"}  # default format = png
```
skrawcz, line 902: why drop `"view": False`?

zilto: `graphviz.Digraph.pipe()` doesn't have a `view` kwarg. Forgot to add it back for `keep_dot=True` though.

Will try to investigate the failed CI for `build-py38`

skrawcz (approved):

PR #898, opened 2024-05-08
moved `hamilton-ui` docs to top-level

zilto:
Slight refactor of the Hamilton UI docs to have it appear at the top level of the docs

image

This will improve:
- page discovery
- maintain and extend the docs page 

## How I tested this
- built the docs locally

### DIFF
--- a/docs/concepts/index.rst
+++ b/docs/concepts/index.rst
@@ -17,4 +17,5 @@ concepts that makes Hamilton unique and powerful.
    builder
    function-modifiers-advanced
    parallel-task
+   ui
    best-practices/index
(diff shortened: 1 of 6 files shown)


docs/concepts/index.rst
```diff
@@ -17,5 +17,4 @@ concepts that makes Hamilton unique and powerful.
    builder
    function-modifiers-advanced
    parallel-task
-   ui
```
skrawcz, line 20: we'll just need to search for all the places we link to this.

E.g. blog, READMEs, Youtube I think.

skrawcz (approved):

PR #900, opened 2024-05-10
Hamilton restructure on FE code + removes enterprise from main repository 

elijahbenizzy:
Internal discussion led us to this conclusion.


skrawcz (approved):
As long as the build works!

PR #902, opened 2024-05-11
Updates JSON view to be cleaner + ensures table index column has fixed width

elijahbenizzy:


elijahbenizzy:
LOL whoops, didn't see that the other was merged: #901

Issue #877, opened 2024-05-02
Add `Builder.with_materializers()`


zilto:
Following the discussion from #816, there would be benefits to allow materializer nodes to be defined statically at the Driver level (both `DataLoader` and `DataSaver`). 

- The nodes can be called directly via `.execute()`
- Materializers appear in `HamiltonGraph` and visualizations even if they aren't executed.
- Validate the DAG, including the materializers before execution.

## Solution 1
```python
dr = (
  driver.Builder()
  .with_modules(...)
  .with_materializers(
    to.dlt(
      id="features_duckdb",
      dependencies=["features_df"]m
      destination=duckdb_dest(...),
    )
  )
  .build()
)
```

## Solution 2
An alternative, would be to allow materializers to be imported and added via `.with_modules()`. For example, `production_materializers.py` contains

```python
# production_materializers.py
from hamilton.io.materialization import to

to.dlt(
  id="features__duckdb",
  dependencies=["features_df"],
  destination=duckdb_dest(...),
)
```

```python
from hamilton import driver
import dataflow
import production_materializers

dr = driver.Builder().with_modules(dataflow, production_materializers).build()
```

For basic `to.parquet()` usage, it might be more efficient to store simple Python functions using `pd.to_parquet()` in a module to enable this patterns. More powerful materializers (e.g., dlt) would benefit from this approach though

Issue #713, opened 2024-02-24
Allow materializer targets to specify inputs


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
I want to be able to use materializers to create simple ETLs of just moving data from one place to another.

**Describe the solution you'd like**

A little odd but a good use-case. Specifically, I want something like this:

```python
dr = driver.Builder().build() # empty driver
dr.materialize(
    to.csv(
        id="save_some_input_to_csv",
        dependencies=["some_input"]),
    inputs={
        "some_input" : SOME_DATAFRAME
    }
),
```

This should form a two-node DAG.

**Describe alternatives you've considered**
Just writing a function to do this, or doing it outside of Hamilton.

skrawcz:
Doesn't this already work?

skrawcz:
Yeah it's because it's empty it doesn't work. With anything defined it seems to work.
![Screen Shot 2024-02-23 at 5 54 29 PM](https://github.com/DAGWorks-Inc/hamilton/assets/2328071/d5211d37-ccd3-45d8-8545-c20050dbf571)

elijahbenizzy:
> Yeah it's because it's empty it doesn't work. It needs a node with that input defined right now to work.
> 
> ![Screen Shot 2024-02-23 at 5 54 29 PM](https://github.com/DAGWorks-Inc/hamilton/assets/2328071/d5211d37-ccd3-45d8-8545-c20050dbf571)
> 
> 

Yep that's why I opened up this issue, cause I think it should work! It's a nice way to use materializers and track them. To avoid pains/encourage early failures we can add a validate_inputs param and default it to true.

skrawcz:
closing this since it's partially addressed with #911 .

PR #911, opened 2024-05-16
added Builder.with_materializers()

zilto:
Following #877 

Added the ability to add materializer nodes to the `FunctionGraph` at `Driver` build time. 

Use cases:
- Easier to view materializers as part of `.display_all_functions()`
- Allows to call materializers by name with `.execute()`. Would allow some users to completely ignore `.materialize()` 
- Possible to combine "static" materializers at build time and "dynamic" materializers at execution time
- Catch invalid dataflows earlier than `dr.materialize()`

## Changes
- Builder now accepts `.with_materializers(*materializers)`
- `Builder.build()` adds materializer nodes after creating the `Driver` and returns the updated Driver. The logic is derived from `Driver.materialize()`
- `Builder.copy()` copies the materializers

## How I tested this
- Added tests to check materializer nodes (savers and loaders) are properly added
- test that "static" and "dynamic" materializers can be used together

## Notes
- each time materializers are added `post_graph_construct` hooks are triggered
- Should be include materializers in the `version` hashing of the dataflow? should we treat "static" and "dynamic" differently. IMO, we might want two create two versions: one for "the dataflow transform, ignoring materializers" and one for "all nodes" since they answer different equality / diffing questions 
- there's duplicate logic between `Builder.build()` and `Driver.materialize()`. A better approach could be to have `Driver.add_materializers()` and `Driver.execute_materializers()`. Both `Builder.build()` and `Driver.materialize()` could call `Driver.add_materializers()`  
- Related to #713,  users need to be aware that materializers need a valid `target` or `dependencies` (type and name) to connect to.

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1657,6 +1657,7 @@ class Builder:
         # common fields
         self.config = {}
         self.modules = []
+        self.materializers = []
 
         self.legacy_graph_adapter = None
         # Standard execution fields
@@ -1734,6 +1735,16 @@ class Builder:
         self.adapters.extend(adapters)
         return self
 
+    def with_materializers(self, *materializers) -> "Builder":
+        """Add materializer nodes to the `Driver`
+        The generated nodes can be referenced by name in `.execute()`
+
+        :param materializers: materializers to add to the dataflow
+        :return: self
+        """
+        self.materializers.extend(materializers)
+        return self
+
     def with_execution_manager(self, execution_manager: executors.ExecutionManager) -> "Builder":
         """Sets the execution manager to use. Note that this cannot be used if local_executor
         or remote_executor are also set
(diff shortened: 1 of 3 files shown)


elijahbenizzy:
OK, I like the API. I think the implementation is overly built towards the current way it works. From what I understand:

1. This builds the standard driver
2. This then appends the materializers
3. This is due to the idea of keeping the graph version consistent (right?)
4. It ends up calling `post_graph_construct` twice (once in the construction of the driver, once otherwise)
5. Has a final graph version with materializers

IMO this should:
1. Build the standard driver with the materializers
2. Only call it once
3. Have only one version

So, how would this work? 

Idea: 
1. Take the code out of the driver [constructor](https://github.com/DAGWorks-Inc/hamilton/blob/818f2fca8b800b1a2b66fea31fdd73d5534beed8/hamilton/driver.py#L352) 
2. Put it in a static function
3. Make all driver parameters optional (E.G. have an internal one -- `_graph`)
4. Have there be two paths -- one if you want to construct the graph outside, and the other if you want to do it inside

Alternatively (and this is cleaner, but we have to be careful with some of the SDK stuff, or just say only use the builder/adapter API):

1. Create a superclass (`DriverBase`, that takes in graph/other info)
2. Move just the constructor (currently compatible) into a subclass `Driver`(user-facing for backwards compatibility with the non-builder API)
3. Have the `Builder` build it using the `DriverBase` API

Then we can put more logic in the builder. Perhaps the simplest solution might be:

1. Pass in materializers to the driver, default to empty (See the `_` parameters
2. Build it in the driver constructor

I think this is the easiest, and will save on complexity.

Regarding multiple versions that is, IMO, too complex now. We should add it later?

skrawcz:
I'm in favor of seeing if we can make it part of the constructor with this approach. That way post graph construct is only called once.

skrawcz (commented):
I think this looks good. Just need to fix the pre-commit hook.

skrawcz (approved):

PR #912, opened 2024-05-21
Add lancedb ner example

skrawcz:
See commits.

This is almost there expect for the example code -- need to also create a notebook for this to run properly.

## Changes
 - adds some materializers
 - adds union type check support 

## Checklist


examples/LLM_Workflows/NER_Example/lancedb_module.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _write_to_lancedb(
+    data: Union[list[dict], pa.Table], db: lancedb.DBConnection, table_name: str
+) -> int:
+    """Helper function to write to lancedb.
+
+    This can handle the case the table exists or it doesn't.
+    """
+    try:
```
elijahbenizzy, line 23: Is this the best way to do this with lance? No "does table exist"?

examples/LLM_Workflows/NER_Example/run.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def load_data(table_name: str, use_tracker: bool = False):
+    adapter_list = [lifecycle.PrintLn()]
+    if use_tracker:
+        from hamilton_sdk import adapters
+
+        tracker = adapters.HamiltonTracker(
+            project_id=41,  # modify this as needed
+            username="elijah@dagworks.io",
```
elijahbenizzy, line 28: lol

examples/LLM_Workflows/NER_Example/run.py
```diff
@@ -0,0 +1,89 @@
...
+
+from hamilton import driver, lifecycle
+
+
+def build_driver(adapter_list):
+    """Builds the driver with the necessary modules and adapters."""
+    dr = (
+        driver.Builder()
+        .with_config({})
+        .with_modules(ner_extraction, lancedb_module)
```
elijahbenizzy, line 14: I'm confused as to why these are lumped together. They share one node at the end but its confusing... We only call that node in one spot? 

Oh, it seems that they have shared resources (one way). 

IMO it is cleaner/easier to share if you break it up + it highlights the value of modularity + reuse for the blog post.

hamilton/plugins/huggingface_extensions.py
```diff
@@ -0,0 +1,206 @@
...
+from hamilton.io import utils
+from hamilton.io.data_adapters import DataLoader, DataSaver
+
+COLUMN_FRIENDLY_DF_TYPE = False
+
+HF_types = (DatasetDict, Dataset, IterableDatasetDict, IterableDataset)
+
+
+@dataclasses.dataclass
+class HuggingFaceDSLoader(DataLoader):
```
elijahbenizzy, line 38: Blech all the insane huggingface params + code 😆

hamilton/plugins/huggingface_extensions.py
```diff
@@ -0,0 +1,206 @@
...
+from hamilton import registry
+from hamilton.io import utils
+from hamilton.io.data_adapters import DataLoader, DataSaver
+
+COLUMN_FRIENDLY_DF_TYPE = False
+
+HF_types = (DatasetDict, Dataset, IterableDatasetDict, IterableDataset)
+
+
+@dataclasses.dataclass
```
elijahbenizzy, line 37: Comments on this -- why we have it, what it is (even though its obvious)

hamilton/plugins/huggingface_extensions.py
```diff
@@ -0,0 +1,206 @@
...
+        else:
+            f_meta = (utils.get_file_metadata(self.path_or_buf),)
+        return {"file_metadata": f_meta, "dataset_metadata": ds_meta}
+
+    @classmethod
+    def name(cls) -> str:
+        return "parquet"
+
+
+if lancedb is not None:
```
elijahbenizzy, line 142: We should come up with a better way to do this... I guess this is fine for now, but it seems like we include this in lanceDB and only have the `applicable_types` include Dataset if we have Huggingface installed...

elijahbenizzy (commented):
Looking good -- a few minor hygiene stuff

examples/LLM_Workflows/NER_Example/lancedb_module.py
skrawcz: yep.

examples/LLM_Workflows/NER_Example/run.py
skrawcz: not sure I follow?

hamilton/plugins/huggingface_extensions.py
skrawcz: that's the same though. So I thought this was more HF specific so should live here.

hamilton/plugins/huggingface_extensions.py
```diff
@@ -0,0 +1,206 @@
...
+    streaming: bool = False
+    num_proc: Optional[int] = None
+    storage_options: Optional[Dict] = None
+    config_kwargs: Optional[Dict] = None
+
+    @classmethod
+    def applicable_types(cls) -> Collection[Type]:
+        return list(HF_types)
+
+    def _get_loading_kwargs(self) -> dict:
```
zilto, line 67: Might want to look at this approach: https://github.com/DAGWorks-Inc/hamilton/blob/818f2fca8b800b1a2b66fea31fdd73d5534beed8/hamilton/plugins/dlt_extensions.py#L96

I think `fields_to_skip` is a clear approach. Could document further why to skip in the docstring

examples/LLM_Workflows/NER_Example/run.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+        adapter_list.append(tracker)
+
+    dr = build_driver(adapter_list)
+
+    r = dr.execute(["lancedb_result"], inputs={"query": query, "table_name": table_name})
+    print(r)
+
+
+def main():
```
zilto, line 64: I believe `main()` should contain the logic and not rely on user inputs while `argparse` code exist after `if __name__ == "__main__"`.

examples/LLM_Workflows/NER_Example/run.py
elijahbenizzy: Specifically there are two drivers. They do two different things. One has a dependency on the other. The graph is the same/there is one `build_driver` function.

Feels like we should break it into two?

hamilton/plugins/huggingface_extensions.py
skrawcz: I tried it -- and it's more lines of code in the other cases. So not going to use it here.

elijahbenizzy (approved):
Looks fine, I think we should probably break it up into more fns/modules, but see it going either way

PR #916, opened 2024-05-22
`plugins` Kedro

zilto:
Added a plugin to convert Kedro `Pipeline` objects to Hamilton `Driver`. This allows Kedro users to benefit from the ergonomics of Hamilton and it's powerful Hamilton UI observability features.

## Changes
- added `hamilton/plugins/h_kedro.py` with the Pipeline to Driver conversion
- added `hamilton/examples/kedro/kedro-plugin/` with a README and a tutorial notebook

## How I tested this
- the test coverage is low, but we manually made sure pipelines could run and successfully be captured by the Hamilton UI. Issues wouldn't be surprising as Kedro users try it with less common usage patterns.
- added `tests/plugins/test_h_kedro.py`. It focuses on the supporting Kedro nodes that return multiple values while Hamilton nodes return a single value (and may use `@extract_fields`).


zilto:
For Python <3.11, `@extract_fields` had a bug. This is fixed in [387cdc2](https://github.com/DAGWorks-Inc/hamilton/pull/916/commits/387cdc2a17b88516ab7cf6b57af8b8949a9e4716).

Essentially, the condition `isinstance(Any, type)` returns `False` in Python < 3.11. Here's a repro of the bug on Python 3.10

```python
from typing import Any
from hamilton.function_modifiers import extract_fields

@extract_fields(dict(B=Any, C=Any))
def A() -> dict:
    return dict(B=True, C=1)
    
if __name__ == "__main__":
    import __main__
    from hamilton import driver
    
    res = dr.execute(["B", "C"])
```

I previously discussed the issue with @elijahbenizzy in the context of this plugin, but I'm now fixing it given the bug it broader than the plugin itself

skrawcz (approved):
I think this looks good. I haven't tested it but from the demo you showed it works. I would squash commit and write up any design decisions in the commit.

PR #917, opened 2024-05-23
jupyter V2.1 with backwards compatible changes

zilto:
Reimplementation of #846 with backwards compatible changes with V1.

Compared to V2:
- support `module_name` as positional or named arg (`-m/--module_name`)
- support `-c/--config`
- support `-h/--help`
- ignores `-r/--rebuild-drivers`
- ignores `-v/--verbose`
- richer support for `-c/--config`
- alias `%%insert_module` to `%%module_to_cell`
- creates a temporary file for the temporary modules to play nicely with Hamilton UI

Currently ignores the `%%incr_cell_to_module`. We have to consider what it means to deduplicate these 100+ lines of logic. If this feature truly adds value, it should be an argument instead of a separate magic. I propose `-a/--append INDEX`. This means the magic would now require to be stateful though opening opportunities for more bugs. 

## How I tested this

The notebook `tutorial.ipynb` covers all implemented functionalities, so it's a good test for intended behavior.

## TODO
- Manually test all notebooks using `%%cell_to_module`
- Reimplement the 2 notebooks currently using `%%incr_cell_to_module`
- Update documentation


hamilton/plugins/jupyter_magic.py
```diff
@@ -77,159 +131,272 @@ def insert_cell_with_content():
...
+        )
+
+        # Decision: write to file before trying to build and execute Driver
+        # See argument `help` for behavior details
+        if args.write_to_file:
+            if isinstance(args.write_to_file, str):
+                file_path = Path(args.write_to_file)
+            else:
+                file_path = Path(f"{module_name}.py")
+            file_path.write_text(cell)
```
skrawcz, line 324: this should probably happen after creating the module to catch syntax/import errors, right? Else if it's here bad code could be saved when we should error instead?

hamilton/plugins/jupyter_magic.py
```diff
@@ -242,159 +409,13 @@ def insert_module(self, line):
...
         module_path = Path(line)
         # insert our custom %%with_functions magic at the top of the cell
-        header = f"%%cell_to_module -m {module_path.stem}\n\n"
+        header = f"%%cell_to_module {module_path.stem}\n"
         module_source = module_path.read_text()
-
         # insert source code as text in the next cell
         self.shell.set_next_input(header + module_source, replace=False)
 
-    @magic_arguments()  # needed on top to enable parsing
```
skrawcz, line 251: We can't delete it -- we can chat about whether subsuming the functionality makes sense or not.

hamilton/plugins/jupyter_magic.py
```diff
@@ -77,159 +131,272 @@ def insert_cell_with_content():
...
 
-        # create a driver to display things for every cell with %%with_functions
-        dr = driver.Builder().with_modules(module_object).with_config(display_config).build()
-        self.shell.push({f"{module_name}_dr": dr})
+        # determine final vars
+        if args.execute not in [True, None]:
+            final_vars = self.shell.user_ns[args.execute]
+        else:
+            nodes = [n for n in dr.list_available_variables() if not n.is_external_input]
+            final_vars = topological_sort(nodes)
```
skrawcz, line 344: why a topological sort? why not just sort it alphabetically?

skrawcz (changes requested):
looks good, but undelete `%incr_cell_to_module`

hamilton/plugins/jupyter_magic.py
zilto: The output display currently doesn't print the keys, so topological sorts makes the most sense to print values

image

hamilton/plugins/jupyter_magic.py
zilto: ## 1. Invalid Python
Invalid Python syntax won't get there because of the first line of the magic is `self.shell.ex(cell)`. This will catch any Python error

## Invalid Hamilton
The design decision here is to write the code cell after magic argument parsing, but before further magic logic.

The "contract" of `--write_to_file` is that the code cell is in sync with the module. The user is responsible for what they write to a file (they could write incorrect Python to a `.py` file too).

hamilton/plugins/jupyter_magic.py
```diff
@@ -382,6 +389,152 @@ def B(A: int) -> bool:
...
+        nargs="?",
+        const=True,
+        help="Write cell content to a file. The argument is the file path; else write to {module_name}.py",
+    )
+    @cell_magic
+    def incr_cell_to_module(self, line, cell):
+        """Incrementally build a module. This executes the cell and dynamically creates a Python module from its content.
+        A Hamilton Driver is automatically instantiated with that module for variable `{MODULE_NAME}_dr`.
+
+        > %%incr_cell_to_module -m MODULE_NAME -i IDENTIFIER --display
```
skrawcz, line 430:
```suggestion
        > %%incr_cell_to_module MODULE_NAME -i IDENTIFIER --display
```
can drop `-m`.

skrawcz (approved):
Looks good. I assume you've tested it, e.g. %incr_cell_to_module still works? :)

skrawcz:
Cool - I'd squash merge this PR when it's ready.

Issue #827, opened 2024-04-16
Lifecycle adapter for automatic schema evolution


zilto:
Inspired by [dlt schema features](https://dlthub.com/docs/general-usage/schema) (auto schema, schema evolution, data contracts), we could create lifecycle adapter to capture the schema of dataframe objects and store them to a file (pandas, polars, pyarrow, and Ibis to get started?). 

Then, the same adapter could do subsequent runs:
- strict check: fail if schema doesn't match
- warning: try to execute, but log schema mismatch
- evolution: update the stored schemas

We could possibly convert these schemas to dlt schemas for full visibility over ETL. Similarly, Ibis expression allows for granular lineage (e.g., "new_users" from HamiltonNode 1 was renamed to "users" in HamiltonNode 3)

**Describe the solution you'd like**
Use `lifecycle` adapters, in particular `run_after_node_execution` and `run_after_graph_execution`. The challenge is to find the right format to store this information and integrate nicely with other tools

**Describe alternatives you've considered**

**Additional context**

PR #925, opened 2024-05-28
added `SchemaValidator`and `databackend`

zilto:
Follows #827 

The `SchemaValidator()` is added to `hamilton.experimental.h_schema`. It allows to store dataframe schemas as `pyarrow.Schema` on disk and conduct validation against them on subsequent runs. I imagine that we would want to move this to the core library / outside experimental eventually

Another significant change is the introduction of `hamilton.experimental.h_databackends` to do type-checking against abstract class representations. For example, this allows to determine if an object is an ibis Table without having to import Ibis. More broadly, this would help simplify dependencies in our plugins. 

## TODO
- deal with columns
- create a diff between two pyarrow schema to surface to the use

## Checklist


examples/schema/README.md
```diff
@@ -0,0 +1,11 @@
+# Schema tracking and validation
+
+This examples showcases the `SchemaValidator()` adapter. At runtime, it will collect the schema of DataFrame-like objects and stored them as `pyarrow.Schema` objects. On subsequent runs, it will check if the schema of a noed output matches the stored reference schema.
+
+> NOTE­. `SchemaValidator` is currently under `hamilton.experimental.h_schema`. We're currently iterating over the API and are looking forward to integrate it in the main library.
+
+## Content
+- `dataflow.py` defines a dataflow with 3 nodes returning pyarrow, ibis, and pandas dataframes and use the `SchemaValidator()` to track their schema. After running it once, you can change the `Driver` config to use `version="2"` and see the schema validation fail.
```
elijahbenizzy, line 8: This is an interesting point -- changing the `config` might actually cause a desired change to the schema version... But for now I think this makes sense.

examples/schema/dataflow.py
```diff
@@ -0,0 +1,78 @@
+import duckdb
```
elijahbenizzy, line 1: This example is a little dense -- it feels more like a test than an example (enumerating a bunch of DF types). Maybe just a few more explanations + a DAG to help.

hamilton/experimental/h_databackends.py
```diff
@@ -0,0 +1,161 @@
+# Copyright (c) 2024 databackend contributors (MIT License)
```
elijahbenizzy, line 1: Licensing? We need to include this copyright (although @skrawcz is more of an expert here): https://github.com/machow/databackend/blob/main/LICENSE.)

hamilton/experimental/h_databackends.py
```diff
@@ -0,0 +1,161 @@
...
+        elif "Column" in name:
+            abstract_column_types.add(cls)
+
+    # overwrite sets with Union type objects
+    DATAFRAME_TYPES = Union[tuple(abstract_dataframe_types)]
+    COLUMN_TYPES = Union[tuple(abstract_column_types)]
+    return DATAFRAME_TYPES, COLUMN_TYPES
+
+
+DATAFRAME_TYPES, COLUMN_TYPES = register_backends()
```
elijahbenizzy, line 161: So, just to be clear, is this replacing our standard plugin methodology? Not 100% sure the value of this (although I did see your loom) -- need to look over it a bit closer.

hamilton/experimental/h_schema.py
```diff
@@ -0,0 +1,277 @@
...
+            nullable=field.nullable,
+            metadata=field.metadata,
+        )
+
+    return schema_dict
+
+
+# TODO use ENUMs for a safe format and create a human readable format
+# ENUM {0: "equal", 1: "added", 2: "removed", 3: "edited"}
+def diff_schemas(
```
elijahbenizzy, line 35: More comments on this page for functions/whatnot

hamilton/experimental/h_schema.py
```diff
@@ -0,0 +1,277 @@
...
+        df = from_dataframe(df, allow_copy=False)
+    except RuntimeError as e:
+        if allow_copy is False:
+            raise e
+        df = from_dataframe(df, allow_copy=True)
+
+    return df.schema
+
+
+@_get_arrow_schema.register
```
elijahbenizzy, line 135: So yeah, this uses something similar to single dispatch (which we use elsewhere). We have a few too many ways of building polymorphic things, but again, need to internalize what this is getting (seems pretty cool here).

hamilton/experimental/h_schema.py
```diff
@@ -0,0 +1,277 @@
...
+
+class SchemaValidator(NodeExecutionHook, GraphExecutionHook):
+    """Collect dataframe and columns schemas at runtime. Can also conduct runtime checks against schemas"""
+
+    def __init__(
+        self,
+        schema_dir: str = "./schema",
+        check: bool = True,
+        must_exist: bool = False,
+        importance: Literal["warn", "fail"] = "warn",
```
elijahbenizzy, line 160: May want to consider a filterlist of sorts at some point, or use tags.

elijahbenizzy (commented):
Looks

examples/schema/README.md
zilto: It's very use case dependent. Maybe `@config.when()` changes imputation methods, but both should ensure a schema with no nullable type.

I'm currently storing the `HamiltonNode` metadata, including the `.version` field on the `pyarrow.Schema` object. You can enable metadata comparison as part of the schema diff / equality check  via `check_schema_metadata=True`. 

Still have to figure out the best user-facing API though

hamilton/experimental/h_databackends.py
zilto: yeah, copy-pasted it here for the purpose the Draft PR. I'm guessing it's fine to keep the differently licensed (MIT) code in a separate file

hamilton/experimental/h_databackends.py
zilto: This could effectively replace the need to [dynamically register distributed plugin files](https://github.com/DAGWorks-Inc/hamilton/blob/770050751840318aa2ac923c0f3137c4412dfa2f/hamilton/registry.py#L76)

examples/schema/README.md
elijahbenizzy: Yeah, it's interesting.  Tangential though.

hamilton/experimental/h_databackends.py
elijahbenizzy: I think so? Not super worried but I'll delegate to @skrawcz. Probably best to add that in, I think the license is compatible.

skrawcz: Is this code taken from that repository?

zilto: The package is a single file. The original repo is from someone at Posit/RStudio who works on great_tables. They integrated it to great_tables without any special license since the two repos are MIT in that case

https://github.com/posit-dev/great-tables/blob/main/great_tables/_databackend.py

skrawcz: Yep cool. so this notice makes sense. I'd just mention which functions /parts of this file this relates to if we're being pedantic :)

hamilton/lifecycle/schema.py
skrawcz, line 1: This file needs unit tests.

skrawcz (approved):

PR #926, opened 2024-05-29
Graceful errors

elijahbenizzy:
Response to slack question: https://hamilton-opensource.slack.com/archives/C03M33QB4M8/p1716954444075279

### DIFF
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -8,7 +8,7 @@ import pprint
 import random
 import shelve
 import time
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Type, Union
 
 from hamilton import graph_types, htypes
 from hamilton.graph_types import HamiltonGraph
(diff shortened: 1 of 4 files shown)


elijahbenizzy:
POC, this works!

```python
from hamilton import driver
from hamilton.lifecycle import default


class DoNotProceed(Exception):
    pass


def working_res_1() -> int:
    return 1


def working_res_2() -> int:
    return 2


def do_not_proceed_1(working_res_1: int, working_res_2: int) -> int:
    raise DoNotProceed()


def proceed_1(working_res_1: int, working_res_2: int) -> int:
    return working_res_1 + working_res_2


def proceed_2(working_res_1: int, working_res_2: int) -> int:
    return working_res_1 * working_res_2


def short_circuited_1(working_res_1: int, working_res_2: int, do_not_proceed_1: int) -> int:
    return 1  # this should not be reached


def proceed_3(working_res_1: int, working_res_2: int) -> int:
    return working_res_1 - working_res_2


def do_not_proceed_2(proceed_1: int, proceed_2: int, proceed_3: int) -> int:
    raise DoNotProceed()


def short_circuited_2(proceed_1: int, proceed_2: int, proceed_3: int, do_not_proceed_2: int) -> int:
    return 1  # this should not be reached


def short_circuited_3(short_circuited_1: int) -> int:
    return 1  # this should not be reached


def proceed_4(proceed_1: int, proceed_2: int, proceed_3: int) -> int:
    return proceed_1 + proceed_2 + proceed_3


if __name__ == "__main__":
    import __main__

    dr = (
        driver.Builder()
        .with_modules(__main__)
        .with_adapters(
            default.GracefulErrorAdapter(error_to_catch=DoNotProceed, sentinel_value=None)
        )
        .build()
    )
    dr.display_all_functions()
    vars = dr.list_available_variables()
    res = dr.execute(vars)
    print(res) # has nones instead of values we don't want to compute or we want to short-circuit
```

skrawcz:
This is a vote to allow multiple adapters to implement `run_to_execute_node` but we can cross that bridge when someone gets there.

hamilton/lifecycle/default.py
```diff
@@ -506,3 +506,43 @@ def run_after_node_execution(
...
+        """Initializes the adapter. Allows you to customize the error to catch (which exception
+        your graph will throw to indicate failure), as well as the sentinel value to use in place of
+        a node's result if it fails (this defaults to ``None``).
+
+        Note that this is currently only compatible with the dict-based result builder (use at your
+        own risk with pandas series, etc...).
+
+        Be careful using ``None`` as the default -- feel free to replace it with a sentinel value
+        of your choice (this could negatively impact your graph's execution if you actually *do* intend
+        to use ``None`` return values).
```
skrawcz, line 530: can you provide python example of usage, e.g. passing in the Exception type to catch, and what it would do.

GracefulErrorAdapter(ValueError) vs  GracefulErrorAdapter(Exception) etc

skrawcz (approved):
just add more to the doc so a user can cut and paste an example.

skrawcz (approved):

PR #933, opened 2024-05-31
`UI` remove mention of "email" and request a username

zilto:
## Checklist

### DIFF
--- a/ui/frontend/src/auth/LocalAccount.tsx
+++ b/ui/frontend/src/auth/LocalAccount.tsx
@@ -6,7 +6,7 @@ export const LocalAccount = () => {
   const localAccount = useContext(UserContext)?.username;
   return (
       <div className="mt-20">
-        <h1 className="text-3xl font-bold text-gray-700 pl-5 pb-4">Email: {localAccount}</h1>
+        <h1 className="text-3xl font-bold text-gray-700 pl-5 pb-4">Username: {localAccount}</h1>
         <ProjectLogInstructions
           projectId={undefined}
           username={localAccount || ""}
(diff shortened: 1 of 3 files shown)


skrawcz:
does the backend assume an email? 🤔

zilto:
> does the backend assume an email? 🤔 or the SDK?

With this change UI change, I can create a new account with `username=my_username` and I'm able to run the `examples/hamilton_ui/run.py` and track results + display in the UI.

The SDK had only one reference to `email` in a docstring. Otherwise, everything is under `username` which is a string type. 

For the server, the concept of `email` is more common and it's a field of the `User` django model ([ref](https://github.com/DAGWorks-Inc/hamilton/blob/ab1b8a2cc0719df1936be63b6b04e21b2dbdf33f/ui/backend/server/trackingserver_auth/models.py#L86)). It will check for the input to be a[ valid email using regex](https://docs.djangoproject.com/en/5.0/ref/forms/fields/#emailfield). In our code, the user authentication during tracking [assumes an email ](https://github.com/DAGWorks-Inc/hamilton/blob/ab1b8a2cc0719df1936be63b6b04e21b2dbdf33f/ui/backend/server/trackingserver_base/auth/local.py#L12) from the `username` passed via HTTP.

## Next steps
From my manual testing, this UI change doesn't introduce any bugs for local mode, but a valid email might be required for enterprise features.

Potential solutions:
1.  Generate a valid `email` from the `username` string on the server side when in local mode.  
2. Add a UI element saying "Emails are stored within your application for authentication and never communicated externally."

IMHO, it still relevant to remove the "enter email" constraint for a locally hosted application. On the other hand, extremely popular self-hosted tools still require an email + password to enforce good security practices (e.g., NGINX proxy manager)

elijahbenizzy (approved):
So this will work -- the full auth system integration will break (it requires emails), but given that this is separate that's fine.

PR #947, opened 2024-06-12
Enables inputs to be also outputs; Fixes #936

### LINKED ISSUE #936: Enable outputs to access all inputs
You cannot currently request inputs that are not referenced in the DAG.

Allow one to pass in arbitrary inputs not used in the DAG and then be able to request them as outputs.

**To recreate**
```python
# functions.py - declare and link your transformations as functions....
import pandas as pd

def a(input: pd.Series) -> pd.Series:
    return input % 7

def b(a: pd.Series) -> pd.Series:
    return a * 2

def c(a: pd.Series, b: pd.Series) -> pd.Series:
    return a * 3 + b * 2

def d(c: pd.Series) -> pd.Series:
    return c ** 3
```

```python
# And run them!
import functions
from hamilton import driver
dr = driver.Driver({}, functions)
result = dr.execute(
   ['c','d', 'foo'],  # want to request foo -- but it's not referenced anywhere
   inputs={'input': pd.Series([1, 2, 3, 4, 5]), 'foo': 1}
)
print(result)
dr.display_all_functions(
   "graph.dot", orient="TB", show_legend=False)
```
```
ValueError: Unknown nodes [foo] requested. Check for typos?
```

You write a dummy function to contain it as input.

skrawcz:
This enables one to pass in inputs and request them as outputs independent of the graph.

The use case here is that you want to join some data at the end that is extra and not in the DAG. E.g. extra pandas data.

I skip the visualization parts because this is an odd case.

This was easier than expected because the new executor already uses inputs as outputs... So really only had to modify the old code path. 

## Changes
 - modifies dfs function to check inputs if node isn't in graph or config.

## How I tested this
 - locally via unit tests

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -156,6 +156,8 @@ class DefaultGraphExecutor(GraphExecutor):
         nodes = [fg.nodes[node_name] for node_name in final_vars if node_name in fg.nodes]
         fg.execute(nodes, memoized_computation, overrides, inputs, run_id=run_id)
         outputs = {
+            # we do this here to enable inputs to also be used as outputs
+            # putting inputs into memoized before execution doesn't work due to some graphadapter assumptions.
             final_var: memoized_computation.get(final_var, inputs.get(final_var))
             for final_var in final_vars
         }  # only want request variables in df.
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -153,10 +153,11 @@ class DefaultGraphExecutor(GraphExecutor):
         """Basic executor for a function graph. Does no task-based execution, just does a DFS
         and executes the graph in order, in memory."""
         memoized_computation = dict()  # memoized storage
-        nodes = [fg.nodes[node_name] for node_name in final_vars]
+        nodes = [fg.nodes[node_name] for node_name in final_vars if node_name in fg.nodes]
         fg.execute(nodes, memoized_computation, overrides, inputs, run_id=run_id)
         outputs = {
-            final_var: memoized_computation[final_var] for final_var in final_vars
+            final_var: memoized_computation.get(final_var, inputs.get(final_var))
+            for final_var in final_vars
         }  # only want request variables in df.
         del memoized_computation  # trying to cleanup some memory
         return outputs
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -1035,6 +1035,8 @@ class FunctionGraph:
         missing_vars = []
         for var in starting_nodes:
             if var not in self.nodes and var not in self.config:
+                # checking for runtime_inputs because it's not in the graph isn't really a graph concern. So perhaps we
+                # should move this outside of the graph in the future. This will do fine for now.
                 if var not in runtime_inputs:
                     # if it's not in the runtime inputs, it's a properly missing variable
                     missing_vars.append(var)
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -981,7 +981,9 @@ class FunctionGraph:
                 deps.append(dep)
             return deps
 
-        return self.directional_dfs_traverse(next_nodes_function, starting_nodes=final_vars)
+        return self.directional_dfs_traverse(
+            next_nodes_function, starting_nodes=final_vars, runtime_inputs=runtime_inputs
+        )
 
     def nodes_between(self, start: str, end: str) -> Set[node.Node]:
         """Given our function graph, and a list of desired output variables, returns the subgraph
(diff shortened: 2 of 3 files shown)


tests/test_end_to_end.py
```diff
@@ -448,3 +451,93 @@ def test_driver_validate_with_overrides_2():
...
+    )
+    actual = dr.execute(
+        ["avg_3wk_spend", "e"],
+        inputs={"spend": pd.Series([1, 1, 1, 1, 1]), "e": pd.Series([10, 10, 10, 10, 10])},
+    )
+    pd.testing.assert_frame_equal(
+        actual,
+        pd.DataFrame(
+            {
+                "avg_3wk_spend": pd.Series([math.nan, math.nan, 1.0, 1.0, 1.0], dtype=float),
```
zilto, line 482: could be `pd.NaN` and avoid importing `math`

tests/test_end_to_end.py
```diff
@@ -448,3 +451,93 @@ def test_driver_validate_with_overrides_2():
...
+                "e": [10, 10, 10, 10, 10],
+            }
+        ),
+    )
+
+
+def test_driver_v2_extra_inputs_can_be_outputs():
```
zilto, line 489: We should add tests for `.validate_execution()`, `.validate_materialization()`, and `.validate_inputs()`

tests/test_end_to_end.py
skrawcz: No it doesn't work because I need to say it's a float type for the test comparison to work.

```python
pd.Series([pd.NA, pd.NA, 1.0, 1.0, 1.0], dtype=float)
  File "/Users/stefankrawczyk/.pyenv/versions/knowledge_retrieval-py39/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-baac06950501>", line 1, in <module>
    pd.Series([pd.NA, pd.NA, 1.0, 1.0, 1.0], dtype=float)
  File "/Users/stefankrawczyk/.pyenv/versions/3.9.13/envs/knowledge_retrieval-py39/lib/python3.9/site-packages/pandas/core/series.py", line 470, in __init__
    data = sanitize_array(data, index, dtype, copy)
  File "/Users/stefankrawczyk/.pyenv/versions/3.9.13/envs/knowledge_retrieval-py39/lib/python3.9/site-packages/pandas/core/construction.py", line 622, in sanitize_array
    subarr = _try_cast(data, dtype, copy, raise_cast_failure)
  File "/Users/stefankrawczyk/.pyenv/versions/3.9.13/envs/knowledge_retrieval-py39/lib/python3.9/site-packages/pandas/core/construction.py", line 840, in _try_cast
    subarr = np.array(arr, dtype=dtype, copy=copy)
TypeError: float() argument must be a string or a number, not 'NAType'
```
(there is no NaN)

hamilton/graph.py
```diff
@@ -981,7 +981,9 @@ def next_nodes_function(n: node.Node) -> List[node.Node]:
                 deps.append(dep)
             return deps
 
-        return self.directional_dfs_traverse(next_nodes_function, starting_nodes=final_vars)
+        return self.directional_dfs_traverse(
```
elijahbenizzy, line 984: I think this shouldn't change -- e.g. we should remove it from nodes if its just a runtime input in the level above. No reason for this to care about the strange 'inputs can be outputs too' requirement?

hamilton/driver.py
```diff
@@ -153,10 +153,11 @@ def execute(
         """Basic executor for a function graph. Does no task-based execution, just does a DFS
         and executes the graph in order, in memory."""
         memoized_computation = dict()  # memoized storage
-        nodes = [fg.nodes[node_name] for node_name in final_vars]
+        nodes = [fg.nodes[node_name] for node_name in final_vars if node_name in fg.nodes]
         fg.execute(nodes, memoized_computation, overrides, inputs, run_id=run_id)
         outputs = {
-            final_var: memoized_computation[final_var] for final_var in final_vars
+            final_var: memoized_computation.get(final_var, inputs.get(final_var))
+            for final_var in final_vars
```
elijahbenizzy, line 160: You should be passing in all but the inputs to upstream nodes, rather than relying on the graph to know about it IMO -- see comment below

elijahbenizzy (commented):
I think this is a little too deep -- we can do this at the driver level only, no?

hamilton/driver.py
skrawcz: oh sure I can just put the inputs into memoized dict in the first place..

hamilton/graph.py
skrawcz: I think the better way requires more changes and surgery. Will add a comment to say perhaps this shouldn't be here.

elijahbenizzy (approved):
This is fine, let's ship it. Still think it should be at the driver layer but this is OK for now.

PR #960, opened 2024-06-14
Adds ability to provide same validator N times

### LINKED ISSUE #950: `Value Error` When Applying Multiple Data Validation Decorators to the Same Function
## Description:
I encountered an issue while trying to apply multiple data validation decorators to a single function in the Hamilton DAG framework. Specifically, I am trying to validate different columns of a DataFrame using multiple instances of the @check_output_custom decorator. However, I receive a ValueError indicating that the function cannot be defined more than once.
## Steps to Reproduce:
1. Define a function to process a DataFrame.
2. Apply multiple @check_output_custom decorators to the function, each with different validation parameters.
3. Attempt to run the decorated function.

## 1st issue code snippet
```python
@check_output_custom(CompositePrimaryKeyValidatorPySparkDataFrame(columns=["OrderID", "ItemNumber"], importance="fail"))
@check_output_custom(CategoricalValuesValidatorPySparkDataFrame(column="CategoryID", allowed_values=[1, 2, 3], importance="fail"))
def process_order_data(order_data_config: dict, order_filter_template: List) -> DataFrame:
    # Function implementation
    pass
```

## This raises the error:
ValueError: Cannot define function process_order_data_raw more than once. Already defined by function <function process_order_data

## 2nd issue code snippet
```python
@check_output_custom(CategoricalValuesValidatorPySparkDataFrame(column="CategoryID", allowed_values=[1, 2, 3], importance="fail"))
@check_output_custom(CategoricalValuesValidatorPySparkDataFrame(column="ProductID", allowed_values=[10, 20, 30], importance="warn"))
def process_order_data(order_data_config: dict, order_filter_template: List) -> DataFrame:
    # Function implementation
    pass
```

## This raises the error:
ValueError: Cannot define function process_order_data_CategoricalValuesValidator more than once. Already defined by function <function process_order_data

## Expected Behavior
Applying multiple @check_output_custom decorators to a single function should allow for different validation checks on various columns of the DataFrame without raising a ValueError.

## Actual Behavior
A ValueError is raised, indicating that the function cannot be defined more than once by the same validator.

python version = 3.9.5
hamilton library version = 1.65.0 

## Additional Context:
This issue prevents the application of multiple validators to a single function, which is necessary for comprehensive data validation in our use case. It would be helpful if the framework could support multiple validators on the same function without raising errors.

Thank you for your attention to this issue.

skrawcz:
Fixes #950 by allowing multiple validators with the same name.

## Changes
 - validators

## How I tested this
 - locally via unit tests

## Checklist

### DIFF
--- a/hamilton/function_modifiers/validation.py
+++ b/hamilton/function_modifiers/validation.py
@@ -1,4 +1,5 @@
 import abc
+from collections import defaultdict
 from typing import Any, Callable, Collection, Dict, List, Type
 
 from hamilton import node
@@ -38,6 +39,7 @@ class BaseDataValidationDecorator(base.NodeTransformer):
         validators = self.get_validators(node_)
         validator_nodes = []
         validator_name_map = {}
+        validator_name_count = defaultdict(int)
         for validator in validators:
 
             def validation_function(validator_to_call: dq_base.DataValidator = validator, **kwargs):
@@ -45,6 +47,13 @@ class BaseDataValidationDecorator(base.NodeTransformer):
                 return validator_to_call.validate(result)
 
             validator_node_name = node_.name + "_" + validator.name()
+            validator_name_count[validator_node_name] = (
+                validator_name_count[validator_node_name] + 1
+            )
+            if validator_name_count[validator_node_name] > 1:
+                validator_node_name = (
+                    validator_node_name + "_" + str(validator_name_count[validator_node_name] - 1)
+                )
             validator_node = node.Node(
                 name=validator_node_name,  # TODO -- determine a good approach towards naming this
                 typ=dq_base.ValidationResult,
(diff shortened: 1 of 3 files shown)


hamilton/function_modifiers/validation.py
```diff
@@ -38,13 +39,21 @@ def transform_node(
         validators = self.get_validators(node_)
         validator_nodes = []
         validator_name_map = {}
+        validator_name_count = defaultdict(int)
```
elijahbenizzy, line 42: OK, so I realized, this won't work with multiple `@check_output_custom` decorators... Which means that we should error out but really, just add that to the docs for now. That they have to be in the same decorator. It's also just strictly better.

elijahbenizzy (approved):
Add disclaimer to the docs but otherwise looks good

hamilton/function_modifiers/validation.py
skrawcz: it never did work. added comment in docs.

PR #962, opened 2024-06-14
Fixes issues in which foreign key IDs were not displayed in the UI (+ a few others)

elijahbenizzy:
This is a bit ugly -- we're still identifying the underlying cause, but it turns out that in the SQL orm interaction with django ninja, foreign keys will not show up as IDs.

If you have foreign_key_id: int in the model it'll work. So adding this in fixed it.

### DIFF
--- a/ui/backend/server/trackingserver_template/api.py
+++ b/ui/backend/server/trackingserver_template/api.py
@@ -298,6 +298,10 @@ async def get_dag_template_catalog(
     node_templates = [
         NodeTemplateOut.from_orm(node_template) for node_template in all_node_templates
     ]
+    for i, node_template in enumerate(node_templates):
+        node_template.dag_template = all_node_templates[i].dag_template_id
+    # node_templates = []
+    # dag_template = node_template.dag_template_id
     code_artifacts = list(
         {
             code_artifact
@@ -306,7 +310,7 @@ async def get_dag_template_catalog(
         }
     )
     out = CatalogResponse(
-        nodes=node_templates,
+        nodes=node_templates[:],
         code_artifacts=code_artifacts,
     )
     logger.info(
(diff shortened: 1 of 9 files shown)


ui/frontend/openapi-config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "schemaFile": "http://127.0.0.1:8000/api/openapi.json",
+  "schemaFile": "http://127.0.0.1:8241/api/openapi.json",
```
skrawcz, line 2: just double checking, this is the right port?

ui/backend/server/trackingserver_template/api.py
```diff
@@ -298,6 +298,10 @@ async def get_dag_template_catalog(
     node_templates = [
         NodeTemplateOut.from_orm(node_template) for node_template in all_node_templates
     ]
+    for i, node_template in enumerate(node_templates):
+        node_template.dag_template = all_node_templates[i].dag_template_id
+    # node_templates = []
+    # dag_template = node_template.dag_template_id
```
skrawcz, line 304: remove?

ui/backend/server/trackingserver_template/schema.py
```diff
@@ -22,13 +22,20 @@ class Meta:
...
-        # exclude =
 
     primary_code_artifact: Optional[str] = None
 
+    # @classmethod
+    # def from_orm(cls, obj: Any, **kw: Any) -> "NodeTemplateOut":
+    #     out = super().from_orm(obj, **kw)
+    #     out.dag_template_id = obj.dag_template_id
+    #     return out
+
```
skrawcz, line 38: remove?

skrawcz (approved):

ui/frontend/openapi-config.json
elijahbenizzy: yes

PR #964, opened 2024-06-14
Multiple attributes per node

elijahbenizzy:


elijahbenizzy:
image

elijahbenizzy:
Didn't add tests to the sdk cause this wasn't explicitly tested before, this is a very small change

ui/sdk/src/hamilton_sdk/api/constants.py
```diff
@@ -1,2 +1,2 @@
 HAMILTON_API_URL = "http://localhost:8241"
-HAMILTON_UI_URL = "http://localhost:8242"
+HAMILTON_UI_URL = "http://localhost:8241"  # optimizing for mini-mode
```
skrawcz, line 2: err -- docs need to be updated then for use with docker then right?

skrawcz: or there needs to be a way to know which URL to use that the SDK can figure out

elijahbenizzy: See commit notes, this was the quickest change but yeah we can update docs

zilto: side note: `127.0.0.1` should be preferred to `localhost`  [ref](https://www.youtube.com/watch?v=98SYTvNw1kw)

elijahbenizzy: Hmm I'm generally against using a hardcoded value for the purpose of small optimizations, although there are other reasons... Good video.

skrawcz (approved):
yes, but you need to add to docs on how to change the URL to point to wherever the UI lives.

PR #970, opened 2024-06-20
async tracking

elijahbenizzy:
Async tracking client:

```python
from hamilton_sdk import tracker
        tracker_async = adapters.AsyncHamiltonTracker(
            project_id=1,
            username="elijah",
            dag_name="async_tracker",
        )
        dr = (
            await h_async
            .Builder()
            .with_modules(async_module)
            .with_adapters(tracking_async)
            .build()
        )
```

## Changes

Async hooks + tracking client

## How I tested this

Unit tests, manually (will be adding more.

## Notes

New async builder -- have to call `await h_async.Builder()....build()`

TODO:
- [ ] Add better  description here
- [x] Clean up code 
    - [x] Multiple commits
- [x] Batch mode
    - [x] Transfer over batch mode impl from sync version
    - [x] Figure out how to not have `init`, or make it simpler
- [x] Get it working with builder 
    - [x] API -- decide on an easy integration (`.as_async()` in `driver.Builder` is current pref)
- [x] Add any relevant documentation
    - [x] How to use it
    - [ ] Caveats
    - [x] Docstrings
- [x] Determine whether I want `.ainit()` or not...
- [x] Add example into the async example
- [x] Add more unit tests
## Checklist


examples/async/fastapi_example.py
```diff
@@ -1,13 +1,42 @@
+from contextlib import asynccontextmanager
+
 import async_module
 import fastapi
+import hamilton_sdk.adapters
```
skrawcz, line 5: We try to do module imports...
```suggestion
from hamilton_sdk import adapters 
```

examples/async/fastapi_example.py
```diff
@@ -1,13 +1,42 @@
...
+    global dr
+    dr = await h_async.AsyncDriver(
+        {},
+        async_module,
+        result_builder=base.DictResult(),
+        adapters=[
+            tracker_async,
+            # tracker_sync
+        ],
+    ).ainit()
```
skrawcz, line 35: ideally we also add a builder to make this match the new way we're doing things?

examples/async/fastapi_example.py
elijahbenizzy: This is throwaway code

examples/async/fastapi_example.py
elijahbenizzy: See TODO

hamilton/experimental/h_async.py
```diff
@@ -95,45 +192,140 @@ class AsyncDriver(driver.Driver):
...
         :param display_graph: whether or not to display graph -- this is not supported.
         :param inputs:  Inputs for DAG runtime calculation
+        :param _fn_graph: Function graph for compatibility with superclass -- unused
         :return: A dict of key -> result
         """
+        assert _fn_graph is None, (
+            "_fn_graph must not be provided "
+            "-- the only reason you'd do this is to use materialize(), which is not supported yet.."
+        )
+        run_id = str(uuid.uuid4())
```
skrawcz, line 286: this should be passed in if we're trying to be consistent.

hamilton/experimental/h_async.py
```diff
@@ -57,34 +77,111 @@ def execute_node(self, node: node.Node, kwargs: typing.Dict[str, typing.Any]) ->
...
+                # This is messy as it is a method called within a method...
+                # if self.adapter.does_method("do_node_execute", is_async=False):
+                #     result = self.adapter.call_lifecycle_method_sync(
+                #         "do_node_execute",
+                #         run_id=run_id,
+                #         node_=node_,
+                #         kwargs=kwargs,
+                #         task_id=task_id,
+                #     )
+                # else:
```
skrawcz, line 119: ?

hamilton/lifecycle/base.py
```diff
@@ -745,11 +745,20 @@ def __init__(self, *adapters: LifecycleAdapter):
...
         :param adapters: Adapters to group together
         """
-        self._adapters = list(adapters)
+        self._adapters = self._uniqify_adapters(adapters)
         self.sync_hooks, self.async_hooks = self._get_lifecycle_hooks()
         self.sync_methods, self.async_methods = self._get_lifecycle_methods()
         self.sync_validators = self._get_lifecycle_validators()
 
+    def _uniqify_adapters(self, adapters: List[LifecycleAdapter]) -> List[LifecycleAdapter]:
+        """Removes duplicate adapters from the list of adapters -- this often happens on how they're passed in
```
skrawcz, line 754: how would this happen in practice?

ui/sdk/src/hamilton_sdk/adapters.py
```diff
@@ -416,11 +414,36 @@ def __init__(
...
+        #         f"Authentication failed. Please check your username and try again. "
+        #         f"Username: {self.username}"
+        #     )
+        #     raise
+        # except clients.ResourceDoesNotExistException:
+        #     logger.error(
+        #         f"Project {self.project_id} does not exist/is accessible. Please create it first in the UI! "
+        #         f"You can do so at {self.hamilton_ui_url}/dashboard/projects"
+        #     )
+        #     raise
```
skrawcz, line 430: ?

skrawcz, line 430: should this be uncommented? or?

skrawcz (commented):
we should probably schedule time to walkthrough this -- e.g. how to fix not flushing properly...

elijahbenizzy:
> we should probably schedule time to walkthrough this -- e.g. how to fix not flushing properly...

Sure. Flushing does work properly, it occasionally flushes after the end of a request rather than before it, but the task will continue running and it will continue executing.

ui/sdk/src/hamilton_sdk/adapters.py
elijahbenizzy: Should be deleted

hamilton/experimental/h_async.py
elijahbenizzy: Methods versus hooks. Currently we don't respect methods (E.G. `do_node_execute`), as we're using a hook to implement this.

hamilton/lifecycle/base.py
elijahbenizzy: Yeah, it's not great, but deduplicating makes life easier. Will add a TODO.

What happens is that we think about adapters as keys (hook/method names), whereas they are objects. When we do some cleverness (separate them out, etc...) they end up getting merged back in. Might be worth digging in more but for now this is OK.

hamilton/experimental/h_async.py
elijahbenizzy: It's consistent? https://github.com/DAGWorks-Inc/hamilton/blob/5585c24c7b3813fcfadf180c8c24771e43781e15/hamilton/driver.py#L642

examples/async/fastapi_example.py
```diff
@@ -1,13 +1,61 @@
...
+                if response.status == 200:
+                    return True
+                else:
+                    return False
+        except client_exceptions.ClientConnectionError:
+            return False
+
+
+@asynccontextmanager
+async def lifespan(app: fastapi.FastAPI):
```
skrawcz, line 32: It would be nice to have a second endpoint to compare latency/penalty for the tracker.

skrawcz (approved):

PR #976, opened 2024-06-21
`plugin` add schema tracking support for PySpark 

zilto:
PySpark doesn't offer built-in conversion to pyarrow. Instead, I added a manual mapping between the two.

## Changes
- added another databackend implementation for pyspark in `hamilton.plugins.h_schema`

## How I tested this
- added tests for type conversion and schema conversion in `plugin_tests/h_spark/test_h_schema.py`

### DIFF
--- a/hamilton/plugins/h_schema.py
+++ b/hamilton/plugins/h_schema.py
@@ -274,6 +274,72 @@ def _(df: h_databackends.AbstractIbisDataFrame, **kwargs) -> pyarrow.Schema:
     return df.schema().to_pyarrow()
 
 
+def _spark_to_arrow(type_):
+    import pyspark.sql.types as pt
+
+    _from_pyspark_dtypes = {
+        pt.NullType: pyarrow.null(),
+        pt.BooleanType: pyarrow.bool_(),
+        pt.BinaryType: pyarrow.binary(),
+        pt.ByteType: pyarrow.int8(),
+        pt.ShortType: pyarrow.int16(),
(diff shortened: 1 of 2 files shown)


hamilton/plugins/h_schema.py
```diff
@@ -274,6 +274,66 @@ def _(df: h_databackends.AbstractIbisDataFrame, **kwargs) -> pyarrow.Schema:
...
+            arrow_type = _from_pyspark_dtypes[type(type_)]
+        except KeyError:
+            raise NotImplementedError(f"Can't convert {type_} to pyarrow type.")
+
+    return arrow_type
+
+
+@_get_arrow_schema.register(h_databackends.AbstractSparkSQLDataFrame)
+def _get_spark_schema(df, **kwargs) -> pyarrow.Schema:
+    """Convert the Ibis schema to pyarrow Schema. The operation is lazy
```
skrawcz, line 324:
```suggestion
    """Convert the pyspark schema to pyarrow Schema. The operation is lazy
```

hamilton/plugins/h_schema.py
```diff
@@ -274,6 +274,66 @@ def _(df: h_databackends.AbstractIbisDataFrame, **kwargs) -> pyarrow.Schema:
...
+    and doesn't require Ibis execution"""
```
skrawcz, line 325:
```suggestion
    and doesn't require pyspark execution"""
```

skrawcz (approved):

PR #983, opened 2024-06-24
Feature: inline data saver & loaders

skrawcz:
This change enables one to have inline loaders / savers and to expose metadata for them.

```python
import pandas as pd
from sklearn import datasets

from hamilton.function_modifiers import dataloader, datasaver
from hamilton.io import utils as io_utils


@dataloader()
def raw_data() -> tuple[pd.DataFrame, dict]:
    data = datasets.load_digits()
    df = pd.DataFrame(data.data, columns=[f"feature_{i}" for i in range(data.data.shape[1])])
    metadata = io_utils.get_dataframe_metadata(df)
    return df, metadata


def transformed_data(raw_data: pd.DataFrame) -> pd.DataFrame:
    return raw_data


@datasaver()
def saved_data(transformed_data: pd.DataFrame, filepath: str) -> dict:
    transformed_data.to_csv(filepath)
    metadata = io_utils.get_file_and_dataframe_metadata(filepath, transformed_data)
    return metadata
```

This results in the following being captured in the tracker / UI
Screen Shot 2024-07-22 at 1 38 45 PM
Screen Shot 2024-07-22 at 1 38 50 PM
Screen Shot 2024-07-22 at 1 38 57 PM
Screen Shot 2024-07-22 at 1 39 09 PM
Screen Shot 2024-07-22 at 1 46 43 PM

## Changes

- adds decorators & docs
- tests
- example

## How I tested this
 - locally
 - unit tests

## Checklist


skrawcz:
Okay let's instead go for the following if it's simpler to implement:

```python
from hamilton.function_modifiers import loader, saver

@loader # injects node to pull out result
def foo() -> tuple[pd.DataFrame, dict]:
   ...
   return DF, metadata

@saver # all it does is add the right tags
def write_foo(...) -> dict:
   ...
   return metadata
```

elijahbenizzy:
> Okay let's instead go for the following if it's simpler to implement:
> 
> 
> 
> ```python
> 
> from hamilton.function_modifiers import loader, saver
> 
> from hamilton.io import utils as io_utils
> 
> 
> 
> @loader # injects node to pull out result
> 
> def foo() -> tuple[pd.DataFrame, dict]:
> 
>    ...
> 
>    metadata = io_utils....(file, df)
> 
>    return DF, metadata
> 
> 
> 
> @saver # all it does is add the right tags
> 
> def write_foo(...) -> dict:
> 
>    ...
> 
>    metadata = io_utils....(file)
> 
>    return metadata
> 
> ```

I also find it clearer :)

ui/sdk/src/hamilton_sdk/tracking/stats.py
```diff
@@ -100,10 +100,16 @@ def compute_stats_tuple(result: tuple, node_name: str, node_tags: dict) -> Stats
         if isinstance(result[1], dict):
             try:
                 # double check that it's JSON serializable
-                raw_data = json.dumps(result[1])
+                raw_data = (
+                    json.dumps(result[1])
+                    if isinstance(result[1], dict)
+                    else json.dumps(result[1].to_dict())
+                )
```
skrawcz, line 107: need to revert this

hamilton/function_modifiers/__init__.py
```diff
@@ -92,3 +92,5 @@
 # materialization stuff
 load_from = adapters.load_from
 save_to = adapters.save_to
+loader = macros.loader
```
elijahbenizzy, line 95: Feels like these should go in adapters, not macros, like the others

hamilton/function_modifiers/macros.py
```diff
@@ -870,3 +872,106 @@ def optional_config(self) -> Dict[str, Any]:
...
+        :param fn:
+        :param config:
+        :return:
+        """
+        _name = "loader"
+        og_node = node.Node.from_fn(fn, name=_name)
+        new_tags = og_node.tags.copy()
+        new_tags.update(
+            {
+                "hamilton.data_loader": True,
```
elijahbenizzy, line 919: Can/should this share code with this? https://github.com/DAGWorks-Inc/hamilton/blob/d2f3c7d1e1d6b9a8209cc97406bac41ddd84ea39/hamilton/function_modifiers/adapters.py#L601

hamilton/graph.py
```diff
@@ -500,7 +500,9 @@ def _get_legend(
             node_style.update(**modifier_style)
             seen_node_types.add("materializer")
 
-        if n.tags.get("hamilton.data_loader") and "load_data." in n.name:
+        if n.tags.get("hamilton.data_loader") and (
```
elijahbenizzy, line 503: Comments + reason for this change? Confusing.

hamilton/function_modifiers/macros.py
```diff
@@ -870,3 +872,106 @@ def optional_config(self) -> Dict[str, Any]:
 #
 #     def __init__(self, *transforms: Applicable, collapse=False):
 #         super(flow, self).__init__(*transforms, collapse=collapse, _chain=False)
+
+
+class loader(base.NodeCreator):
```
elijahbenizzy, line 877: This calls with `@loader()`. 

Do you ever foresee having parameters? Or should be make it so `loader=Loader()` and have it all be a static instance? Then we call it with `@loader`.

hamilton/function_modifiers/macros.py
```diff
@@ -870,3 +872,106 @@ def optional_config(self) -> Dict[str, Any]:
...
+                f"Function: {fn.__qualname__} must have a return annotation."
+            )
+        # check that the type is a tuple[TYPE, dict]:
+        if not typing_inspect.is_tuple_type(return_annotation):
+            raise InvalidDecoratorException(f"Function: {fn.__qualname__} must return a tuple.")
+        if len(typing_inspect.get_args(return_annotation)) != 2:
+            raise InvalidDecoratorException(
+                f"Function: {fn.__qualname__} must return a tuple of length 2."
+            )
+        if not typing_inspect.get_args(return_annotation)[1] == dict:
```
elijahbenizzy, line 898: Should be able to just use this: https://github.com/DAGWorks-Inc/hamilton/blob/d2f3c7d1e1d6b9a8209cc97406bac41ddd84ea39/hamilton/htypes.py#L116 if you want to kill some code.

hamilton/function_modifiers/macros.py
```diff
@@ -870,3 +872,106 @@ def optional_config(self) -> Dict[str, Any]:
...
+                "hamilton.data_loader.source": f"{fn.__name__}",
+                "hamilton.data_loader.classname": f"{fn.__name__}()",
+                "hamilton.data_loader.node": fn.__name__,
+            },
+        )
+
+        return [og_node.copy_with(tags=new_tags, namespace=(fn.__name__,)), filter_node]
+
+
+class saver(base.NodeCreator):
```
elijahbenizzy, line 947: naming:
`saver`/`loader` (as it is now)
`datasaver`/`dataloader` (a little more explicit)

elijahbenizzy (commented):
Didn't look for polish, otherwise a few comments, looks good

hamilton/function_modifiers/macros.py
skrawcz: tags, maybe.

hamilton/graph.py
skrawcz: needed because of how tags are done -- between data loaders

hamilton/function_modifiers/macros.py
skrawcz: `target_` ?

hamilton/function_modifiers/macros.py
skrawcz: sure.

PR #984, opened 2024-06-25
`fix/examples` updated README and requirements

zilto:
Address a user's bug submission

> I’ve been working through the hamilton_ui example (https://github.com/DAGWorks-Inc/hamilton/tree/main/examples/hamilton_ui), and a a minor error popped up (I wouldn’t even call it that). When running the example, the python run.py  bit, it gave a ValueError: The 'dtype_backend' argument is not supported for the fastparquet engine error. I fixed this by installing pyarrow, which is not currently listed in the requirements.txt file.
Second, the REAMDE.MD file is a bit outdated.
Current:
python run.py --email <email> --project_id <project_id>
should be
python run.py --username <email> --project_id <project_id>
and
python run.py --email <email> --project_id <project_id> --load-from-parquet True
should be
python run.py --username <email> --project_id <project_id> --load-from-parquet
without the True  flag.

### DIFF
--- a/examples/hamilton_ui/README.md
+++ b/examples/hamilton_ui/README.md
@@ -1,65 +1,63 @@
-# machine\_learning
+# Hamilton UI - Machine learning example
 
-This template shows a ML pipeline.
+Learn how to use the `HamiltonTracker` and the Hamilton UI to track a simple machine learning pipeline.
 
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
👍

PR #986, opened 2024-06-25
`.with_materializers()` added helpful error message

zilto:
When first using `.with_materializers()`, it's not obvious that it expects:
```python
.with_materializers(to.json(...), to.parquet(...))
```
Rather than the invalid:
```python
.with_materializers([to.json(...), to.parquet(...)])
```

If encountering the above case, the error will now specify to unpack the sequence with `*`, leading to the valid code
```python
.with_materializers(*[to.json(...), to.parquet(...)])
```

## Changes
- explicitly catch an invalid argument for `.with_materializers()` when building the driver

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -9,6 +9,7 @@ import time
 # required if we want to run this code stand alone.
 import typing
 import uuid
+from collections.abc import Sequence  # typing.Sequence is deprecated in >=3.9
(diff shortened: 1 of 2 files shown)


skrawcz (commented):
Add a unit test for this please.

skrawcz (approved):

PR #994, opened 2024-06-28
Adds extra guards for plugin imports

skrawcz:
We should probably just blanket catch all errors -- but not doing that right now. Hamilton fails to be imported in a spark serverless environment because there is some interaction with DLT that databricks doesn't like.

So catching more exceptions -- and shoe horning it into an import error.

## Checklist


skrawcz:
Verified this works with RC version.

PR #1001, opened 2024-07-01
Handles polars upgrade to 1.0.0

elijahbenizzy:
Decisions:
1. Not backwards compatible -- we're effectively wrapping the library and we don't want to provide a bridge. This is polars versioning, not Hamilton versioning, however, so it's OK to keep this a patch bump (any polars code would have broken already).
2. They deprecated the usage of a connection string (sqllite://) in the PolarsDatabase{reader/writer} -- we've deprecated/changed the type as well
3. It switches based on polars version so people will only get the error for new polars upgrades

Otherwise, longer term, we're going to want to fix this -- moving away from these blind wrappers.


hamilton/plugins/polars_post_1_0_0_extension.py
```diff
@@ -0,0 +1,805 @@
...
+from polars._typing import CsvEncoding, SchemaDefinition
+
+if has_alias and hasattr(pl.type_aliases, "CsvQuoteStyle"):
+    from polars.type_aliases import CsvQuoteStyle
+else:
+    CsvQuoteStyle = Type
+if has_alias and hasattr(pl.type_aliases, "IpcCompression"):
+    from polars.type_aliases import IpcCompression
+else:
+    IpcCompression = Type
```
skrawcz, line 43: this should be simplified with 1.0+

skrawcz (approved):

PR #1008, opened 2024-07-03
Raise error on empty arguments to `group`

### LINKED ISSUE #1007: Misleading error reporting for `parameterize.validate` on `group` dependency with empty list
The `parameterize` decorator throws a misleading error when the dependency for one of the generated nodes is a `group` that has been passed no arguments.

# Current behavior

The error message raised [here](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/function_modifiers/expanders.py#L355) by the `parameterize` decorator is inaccurate. See below.

```python
from hamilton.function_modifiers import parameterize, group, source

parents = []  # generated elsewhere

@parameterize(beaver={
    "x": group(*[source(name) for name in parents])
})
def summation(x: list[int]) -> int:
    return sum(x)
```

Throws the following error:

```python
---------------------------------------------------------------------------
InvalidDecoratorException                 Traceback (most recent call last)
Cell In[6], line 5
      1 from hamilton.function_modifiers import parameterize, group, source
      3 parents = []  # generated elsewhere
----> 5 @parameterize(beaver={
      6     "x": group(*[source(name) for name in parents])
      7 })
      8 def summation(x: list[int]) -> int:
      9     return sum(x)

File ~/Developer/hamilton/hamilton/function_modifiers/base.py:89, in track_decorator_usage.<locals>.replace__call__(self, fn)
     87 else:
     88     DECORATOR_COUNTER["custom_decorator"] = DECORATOR_COUNTER["custom_decorator"] + 1
---> 89 return call_fn(self, fn)

File ~/Developer/hamilton/hamilton/function_modifiers/base.py:131, in NodeTransformLifecycle.__call__(self, fn)
    123 @track_decorator_usage
    124 def __call__(self, fn: Callable):
    125     """Calls the decorator by adding attributes using the get_lifecycle_name string.
    126     These attributes are the pointer to the decorator object itself, and used later in resolve_nodes below.
    127 
    128     :param fn: Function to decorate
    129     :return: The function again, with the desired properties.
    130     """
--> 131     self.validate(fn)
    132     lifecycle_name = self.__class__.get_lifecycle_name()
    133     if hasattr(fn, self.get_lifecycle_name()):

File ~/Developer/hamilton/hamilton/function_modifiers/expanders.py:356, in parameterize.validate(self, fn)
    354                 invalid_types.append((param, param_annotation))
    355 if invalid_types:
--> 356     raise base.InvalidDecoratorException(
    357         f"Validation for fn: {fn.__qualname__} All parameters with a group() parameterization must be annotated as a list: "
    358         f"the following are not: {', '.join([f'{param} ({annotation})' for param, annotation in invalid_types])}"
    359     )

InvalidDecoratorException: Validation for fn: summation All parameters with a group() parameterization must be annotated as a list: the following are not: x (list[int]), x (list[int])
```

Python 3.12.4 on MacOS 14.4

Hamilton 1.67.0

# Expected behavior

This is misleading because the `x` parameter *is* annotated as a list. The offending line is in the [`group`](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/function_modifiers/dependencies.py#L204) dependency annotator, which, if the arguments list is empty, tries to construct a `GroupedDictDependency` from the keyword args instead.

Expected behaviour: [`_validate_group_params`](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/function_modifiers/dependencies.py#L149) should throw an error if no arguments are passed to `group`. This is also what the documentation suggests.

Should be a quick fix, I'll submit a PR.

adzcai:
Fixes #1007 

## How I tested this

Hand testing.

## Checklist

- [x] Dataflow documentation has been updated if adding/changing functionality.

### DIFF
--- a/hamilton/function_modifiers/dependencies.py
+++ b/hamilton/function_modifiers/dependencies.py
@@ -148,37 +148,38 @@ def source(dependency_on: Any) -> UpstreamDependency:
 
 def _validate_group_params(
     dependency_args: List[ParametrizedDependency],
-    dependecy_kwargs: Dict[str, ParametrizedDependency],
+    dependency_kwargs: Dict[str, ParametrizedDependency],
 ):
     """Validates the following for params to group(...):
     1. That either dependency_args or dependency_kwargs is non-empty, but not both.
     2. That all values in dependency_args are of type either LiteralDependency or UpstreamDependency.
 
     :param dependency_args: List of dependencies.
-    :param dependecy_kwargs: Dict of dependencies.
+    :param dependency_kwargs: Dict of dependencies.
     :raises: InvalidDecoratorException if the above conditions are not met.
     """
-    if dependency_args and dependecy_kwargs:
+    if dependency_args and dependency_kwargs:
         raise InvalidDecoratorException(
             "group() can either represent a dictionary or a list of dependencies, not both!"
         )
-    if dependency_args:
+    elif dependency_args:
         for dependency in dependency_args:
             if not isinstance(dependency, (LiteralDependency, UpstreamDependency)):
                 raise InvalidDecoratorException(
                     f"Dependency: {dependency} is not a valid dependency type for group(), must be "
                     f"a LiteralDependency or UpstreamDependency."
                 )
-    if dependecy_kwargs:
-        for dependency in dependecy_kwargs.values():
+    elif dependency_kwargs:
+        for dependency in dependency_kwargs.values():
             if not isinstance(dependency, (LiteralDependency, UpstreamDependency)):
                 raise InvalidDecoratorException(
                     f"Dependency: {dependency} is not a valid dependency type for group(), must be "
                     f"a LiteralDependency or UpstreamDependency."
                 )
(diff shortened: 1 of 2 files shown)


hamilton/function_modifiers/dependencies.py
```diff
@@ -176,6 +176,9 @@ def _validate_group_params(
                     f"Dependency: {dependency} is not a valid dependency type for group(), must be "
                     f"a LiteralDependency or UpstreamDependency."
                 )
+    raise InvalidDecoratorException(
```
elijahbenizzy, line 179: This will always raise, no? Will want to add `return` statements above or something to get it to work.

adzcai: Oops! Thanks 👍

elijahbenizzy:
Looking good! Mind rebasing off of `main` to fix the polars issue + ensure the tests pass? Then we can merge!

PR #1015, opened 2024-07-04
Sketches Schema UI view

skrawcz:
Original description follows. This has been fixed up to work correctly.

This is hacky UI code. It does not handle "run ids".

1. Adds to pandas & pyspark schema capture using `h_schema.py`.
2. Adds SchemaView and related types to try to show a table...

So it's either we've set something up with too many things, or this is just a factor of using typescript... Basically I dont like how many things I had to add to do this..

Otherwise it's non-obvious to me what the pattern should be to handle comparing runs -- this UI change for the schema table doesn't take that into account -- it should probably use Generic Table but I couldn't wrap my head around it.

Screen Shot 2024-07-04 at 3 18 26 PM
Screen Shot 2024-07-04 at 3 17 54 PM
Screen Shot 2024-07-04 at 3 26 52 PM

## Changes
 - SDK
 - DataObservabilityView

## How I tested this
- locally

## Notes
- need help to handle multiple run Ids and styling it appropriately.
- we might not need to use `h_schema` since I kind of already have some of this... we should align on what the "schema" view actually takes in too.

## Checklist


skrawcz:
To complete this:
 - decide on how to create schema stuff — do we use pyarrow stuff or not… (it would mean a dependency on pyarrow is required)
 - redo the table view to enable comparing multiple run IDs.
 - fix SDK tests

elijahbenizzy:
Fixed up schema view: 
image

Added expand/contract now that we have multiple: 
image

ui/frontend/src/components/dashboard/Runs/Task/result-summaries/DataObservability.tsx
```diff
@@ -15,76 +15,148 @@ import {
...
+        // dataTypeDisplay={(item: string) => {
+        //   return (
+        //     <RunLink
+        //       projectId={props.projectId}
+        //       runId={parseInt(item) as number}
+        //       setHighlightedRun={() => void 0}
+        //       highlightedRun={null}
+        //     ></RunLink>
+        //   );
+        // }}
```
skrawcz, line 159: commented out?

ui/sdk/src/hamilton_sdk/tracking/pyspark_stats.py
```diff
@@ -73,53 +74,35 @@ def _introspect(df: ps.DataFrame) -> Dict[str, Any]:
...
-            "name": "Cost Explain",
-        },
-        {
-            "observability_type": "primitive",
-            "observability_value": {
-                "type": str(str),
-                "value": o_value["extended_explain"],
-            },
-            "observability_schema_version": "0.0.1",
-            "name": "Extended Explain",
```
skrawcz: I don't see these added to additional...

elijahbenizzy (approved):
This has been fixed up + approved

ui/sdk/src/hamilton_sdk/tracking/langchain_stats.py
```diff
@@ -43,7 +43,7 @@ def compute_stats_lc_docs(
     from langchain_core import messages
 
     msg = messages.BaseMessage(content="Hello, World!", type="greeting")
-    print(stats.compute_stats(msg, "greeting", {}))
+    print(data_observation.compute_stats(msg, "greeting", {}))
```
skrawcz, line 46: got to remove these 😆

Issue #1009, opened 2024-07-03
GracefulErrorAdapter fails with Parallelizable


JamesArruda:
`GracefulErrorAdapter` does not know to return a list of the sentinel value for `Parallelizable` nodes (ones that are of type `EXPAND`).

# Current behavior

## Stack Traces

The error can be reproduced with the code in the steps to replicate.

```python
Traceback (most recent call last):
  File "c:\git\github\hamilton\tests\test_parallel_error.py", line 42, in <module>
    ans = dr.execute(
          ^^^^^^^^^^^
  File "c:\git\github\hamilton\hamilton\driver.py", line 566, in execute
    raise e
  File "c:\git\github\hamilton\hamilton\driver.py", line 556, in execute
    outputs = self.raw_execute(_final_vars, overrides, display_graph, inputs=inputs)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\git\github\hamilton\hamilton\driver.py", line 685, in raw_execute
    raise e
  File "c:\git\github\hamilton\hamilton\driver.py", line 674, in raw_execute
    results = self.graph_executor.execute(
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "c:\git\github\hamilton\hamilton\driver.py", line 232, in execute
    executors.run_graph_to_completion(execution_state, self.execution_manager)
  File "c:\git\github\hamilton\hamilton\execution\executors.py", line 399, in run_graph_to_completion
    execution_state.update_task_state(task_name, state, result)
  File "c:\git\github\hamilton\hamilton\execution\state.py", line 363, in update_task_state
    self.write_task_results(task_to_update, results)
  File "c:\git\github\hamilton\hamilton\execution\state.py", line 323, in write_task_results
    result_for_expansion = list(results_to_write.pop(result_name_for_expansion))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

## Steps to replicate behavior

```python
from hamilton.htypes import Collect, Parallelizable
from hamilton import driver
from hamilton.execution.executors import SynchronousLocalTaskExecutor
from hamilton.lifecycle import GracefulErrorAdapter


[...]

Issue #742, opened 2024-03-05
Parallelizable cannot aggregate or return multiple Collects


JamesArruda:
# Current behavior

A DAG with one `Parallelizable` and two `Collect` statements cannot return results from  both `Collect` nodes.

## Stack Traces
`KeyError: 'Key metric_2 not found in cache'`

## Steps to replicate behavior

```python
import logging

from hamilton import driver
from hamilton.execution.executors import SynchronousLocalTaskExecutor
from hamilton.htypes import Collect, Parallelizable
import pandas as pd


ANALYSIS_OB = tuple[tuple[str,...], pd.DataFrame]
ANALYSIS_RES = dict[str, str | float]


def split_by_cols(full_data: pd.DataFrame, columns: list[str]) -> Parallelizable[ANALYSIS_OB]:
    for idx, grp in full_data.groupby(columns):
        yield (idx, grp)


def sub_metric_1(split_by_cols: ANALYSIS_OB, number: float=1.0) -> ANALYSIS_RES:
    idx, grp = split_by_cols
    return {"key": idx, "mean": grp["spend"].mean() + number}


def sub_metric_2(split_by_cols: ANALYSIS_OB) -> ANALYSIS_RES:
    idx, grp = split_by_cols
    return {"key": idx, "mean": grp["signups"].mean()}


def metric_1(sub_metric_1: Collect[ANALYSIS_RES], columns: list[str]) -> pd.DataFrame:
    data = [[k for k in d["key"]] + [d["mean"], "spend"] for d in sub_metric_1]
    cols = list(columns) + ["mean", "metric"]
    return pd.DataFrame(data, columns=cols)


def metric_2(sub_metric_2: Collect[ANALYSIS_RES], columns: list[str]) -> pd.DataFrame:
    data = [[k for k in d["key"]] + [d["mean"], "signups"] for d in sub_metric_2]
    cols = list(columns) + ["mean", "metric"]
    return pd.DataFrame(data, columns=cols)


def all_agg(metric_1: pd.DataFrame, metric_2: pd.DataFrame) -> pd.DataFrame:
    return pd.concat([metric_1, metric_2])


if __name__ == "__main__":
    from hamilton.execution import executors
    import __main__

    from hamilton.log_setup import setup_logging
    setup_logging(log_level=logging.DEBUG)

    local_executor = executors.SynchronousLocalTaskExecutor()

    dr = (
        driver.Builder()
[...]

PR #1017, opened 2024-07-05
Graceful failure for parallelism

JamesArruda:
This PR updates the `GracefulErrorAdapter` to handle several situations for `Parallelizable` nodes, as described in https://github.com/DAGWorks-Inc/hamilton/issues/1009.

It also includes a wish-list item (sentinel input) for a broader use case, which is motivated here: https://github.com/DAGWorks-Inc/hamilton/issues/742

This PR allows for sentinel injection into tagged functions, which will support pre-`Collect` blocks that use `hamilton`'s `resolve` to allow runtime-definable DAGs that still aggregate as expected.

## Changes

Using the newly available `is_expand` input, modify the adapter's run logic to:

1. Iterate over the node's generator function
2. Gather successful generator results
3. On an error, push one sentinel into the results list and return
4. Or, treat the generator as pass/fail and any failure returns a single sentinel only
5. Allow sentinel value injection to a node, rather than skipping a node entirely. This is keyed on a node tag.

## How I tested this

The `test_parallel_graceful.py` test file checks for:

1. Two kinds of failure in the `Parallelizable` node (early failure and failure with successful yields)
2. Setting `fail_all_parallel` even with successful yields
3. No `Parallelizable` failure, but checks that failures within the sub-dag are collected.
4. Sentinel injection rules are followed

## Notes

The `Parallelizable` node gets a pre-treatment from `_modify_callable` that handles the list creation from the generator. This makes it hard to specifically handle per-iteration success/fail in the adapter. I opted to pull the original node callable from the partial object out to avoid touching the base implementation. This seems reasonable since it's an optional adapter, but if `_modify_callable` gets more advanced for an `EXPAND` node, this may break the expectations of the adapter. 

It also only pulls out the callable if the user wants to get each successful output, otherwise it goes back to the original behavior.

If "inject" is a loaded term (such as in the `resolve` feature), please suggest a better one!

My intent is that, once this is accepted, to make a more advanced version that has a custom class for the sentinel that stores error traceback information and is easily type-checked for when the sentinel is injected. 

## Checklist


elijahbenizzy:
This is looking great, I'll be able to look in detail tomorrow. If you feel like adding an example (in the examples/ directory), that would be much appreciated! Otherwise no worries, I'm happy to add one later. I also think this would make a nice blog post (we are doing guest authors), so let me know if you're interested!).

Going to download and play around a bit to get a feel for it.

elijahbenizzy:
OK, played around, this is really cool. Here's my testing script btw -- you'll have to replace the project/name with your own and run `hamilton ui`. It's a fun way to explore what they'e doing:

```python
from hamilton import driver
from hamilton.lifecycle import GracefulErrorAdapter
from hamilton_sdk import adapters

import logging

logger = logging.getLogger(__name__)


class SpecificException(Exception):
    pass


def first_node(fail_first_node: bool = False) -> int:
    # First node that can fail
    if fail_first_node:
        raise SpecificException()
    return 1


def parallelizable_block(
        first_node: int,
        total_iterations: int,
        fail_iterations: list[int] = None) -> Parallelizable[int]:
    if fail_iterations is None:
        fail_iterations = []
    # Parallelizable block that can fail
    for i in range(total_iterations):
        if i in fail_iterations:
            raise SpecificException
        yield i


def computed_1(parallelizable_block: int, fail_computed_1: bool = False) -> int:
    # Computed node that can fail
    if fail_computed_1:
        raise SpecificException()
    import time
    time.sleep(1)
    return parallelizable_block + 1


def computed_2(computed_1: int, fail_computed_2: bool = False) -> int:
    # Computed node that can fail
    if fail_computed_2:
        raise SpecificException()
    import time
    time.sleep(1)
    return computed_1 + 2


def computed_3(computed_2: int, fail_computed_3: bool = False) -> int:
    # Computed node that can fail
    if fail_computed_3:
        raise SpecificException()
    import time
    time.sleep(1)
    return computed_2 + 3


def collected(computed_3: Collect[int], fail_collected: bool = False) -> int:
    filtered=[item for item in computed_3 if item is not None]
    if fail_collected:
        raise SpecificException()
    # fail in advance
    return sum(filtered)


def final_node(collected: int, fail_final: bool = False) -> int:
    if fail_final:
        raise SpecificException()
    return collected


if __name__ == "__main__":
    import __main__

    TOTAL_ITERATIONS = 10
    for params in [
        {"fail_first_node": True},
        {"fail_iterations": [3, 5]},
        {"fail_computed_1": True},
        {"fail_computed_2": True},
        {"fail_computed_3": True},
        {"fail_collected": True},
        {"fail_final": True},
    ]:
        try:
            fail_all_parallel = True
            dag_name = str(params.items()).replace("[", "").replace("]", "").replace(", ", "_").replace("'", "")
            tags = {
                **{key: str(value) for key, value in params.items()},
                **{"test_run": "2", "with_adapter": "true", "fail_all_parallel": str(fail_all_parallel).lower()}
            }
            print(f"Running with params: {params}")
            params["total_iterations"] = TOTAL_ITERATIONS
            tracker = adapters.HamiltonTracker(
                project_id=10,  # modify this as needed
                username="elijah@dagworks.io",
                dag_name="initial_test",
                tags=tags
            )
            dr = (
                driver
                .Builder()
                .with_adapters(
                    tracker,
                    GracefulErrorAdapter(
                        error_to_catch=SpecificException
                    )
                )
                .with_modules(__main__)
                .build()
            )
            dr.execute(["final_node"], inputs=params)
        except Exception as e:
            logger.exception(e)

```

hamilton/lifecycle/default.py
```diff
@@ -563,11 +569,19 @@ def never_reached(wont_proceed: int) -> int:
...
         Note you can customize the error you want it to fail on and the sentinel value to use in place of a node's result if it fails.
 
+        For Parallelizable nodes, this adapter will attempt to iterate over the node outputs. If an error occurs, the sentinel value is
+        returned and no more iterations over the node will occur. If you set ``fail_all_parallel`` to be True, it only sends on sentinel
+        value into the parallelize sub-dag.
+
         :param error_to_catch: The error to catch
         :param sentinel_value: The sentinel value to use in place of a node's result if it fails
+        :param fail_all_parallel: Treat a Parallelizable as 1 failure (True) or allow the successful ones to go through (False).
+        :param sentinel_injection_tags: Node tag key:value pairs that allow sentinel injection
```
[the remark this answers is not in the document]
elijahbenizzy: Also, is there any reason you would want to set specific ones? Why not just have a fixed tag? Then you could create a simple decorator that calls out to it:

```python
@drop_errors(...)
```

(still not 100% sure how to use these so this might be wrong...)

hamilton/lifecycle/default.py
```diff
@@ -584,10 +598,43 @@ def run_to_execute_node(
...
+                        return default_return
+        if not is_expand:
+            try:
+                return node_callable(**node_kwargs)
+            except self.error_to_catch:
+                return self.sentinel_value
+
+        # Grab the partial-ized function that is a parallelizable.
+        # Be very specific...
+        if len(node_callable.keywords) == 1 and "_callable" in node_callable.keywords:
```
elijahbenizzy, line 619: This is specifically coupling to this, right? https://github.com/DAGWorks-Inc/hamilton/blob/87c5de0a8a3ff8b31eb4946887735b46aa95fede/hamilton/execution/executors.py#L63? Would it be better to just check if its an expand and make the assumption? Not perfect, but I think it cleans a bit of the coupling up/makes it clearer. Still coupled, but slightly cleaner conditions.

hamilton/lifecycle/default.py
```diff
@@ -584,10 +598,43 @@ def run_to_execute_node(
         # and truncate it/provide sentinels for every failure)
         # TODO -- decide what to do with collect
         """Executes a node. If the node fails, returns the sentinel value."""
-        for key, value in node_kwargs.items():
-            if value == self.sentinel_value:  # == versus is
-                return self.sentinel_value  # cascade it through
+        default_return = [self.sentinel_value] if is_expand else self.sentinel_value
```
elijahbenizzy, line 601: Does empty list not work? Or do you think a single list is cleaner?

hamilton/lifecycle/default.py
```diff
@@ -584,10 +598,43 @@ def run_to_execute_node(
...
+            return [self.sentinel_value]
+        results: list[Any] = []
+        try:
+            for _res in gen:
+                results.append(_res)
+        except self.error_to_catch:
+            if self.fail_all_parallel:
+                results = [self.sentinel_value]
+            else:
+                results.append(self.sentinel_value)
```
elijahbenizzy, line 639: Just to check, if item (3) fails out of 1,2,3,4,5, 4/5 will still be run? Might be worth adding in the docstring.

tests/test_parallel_graceful.py
```diff
@@ -0,0 +1,223 @@
...
+        .build()
+    )
+    return dr
+
+
+def test_parallel_graceful_simple() -> None:
+    # Fail before the anything is yielded: should get 1 failure and skip all
+    dr = _make_driver(
+        temp_module_simple,
+        "early",
```
elijahbenizzy, line 127: These `test_state`, `find_parallel`, and the expected value would be a good use for `@parameterize` -- it'll break it into future tests and make finding the bug one introduces easier. LMK if you want help. figuring that out! Thinking something like:


```python
@pytest.mark.parameterize(
    "test_state,find_parallel,fail_first,expected_value", # add more
    ["early", False, False, [None]]
)
def test_parallel_graceful_simple(test_state,find_parallel,fail_first,expected_value):
    dr = _make_driver(temp_module_simple, test_state, find_prallel=find_parallel, fail_first=fail_first)
    assert dr.execute(["distro_gather"], inputs={"n" : 10}) == expected_value
```

(but modify a bit, just jotting down some ideas)

tests/test_parallel_graceful.py
```diff
@@ -0,0 +1,223 @@
...
+    input_maker__bad,
+    distro__early,
+    distro__middle,
+    distro__pass,
+    some_math,
+    distro_gather,
+    module_name="parallel_example",
+)
+
+temp_module = ad_hoc_utils.create_temporary_module(
```
elijahbenizzy, line 81: So I actually like `create_temporary_module` better for this, but it might be worth putting these two modules in `resources` (as it's consistent with the rest of the way we're doing things): https://github.com/DAGWorks-Inc/hamilton/tree/main/tests/resources.

hamilton/lifecycle/default.py
```diff
@@ -584,10 +598,43 @@ def run_to_execute_node(
...
+        _node_tags = future_kwargs["node_tags"]
+        can_inject = any(
+            _node_tags.get(k, "") == v for k, v in self.sentinel_injection_tags.items()
+        )
+        if not can_inject:
+            for key, value in node_kwargs.items():
+                if type(self.sentinel_value) is type(value):
+                    if self.sentinel_value == value:  # == versus is
+                        return default_return
+        if not is_expand:
```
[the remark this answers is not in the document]
JamesArruda: I don't know all the little details well enough to say for sure. The preceding parts are just supposed to handle if the node is skipped because an input node had an error.

The tests check what happens if the expand node gets an error sentinel, and goes to the collect and beyond, passing the sentinel. If it's not complex enough to detect what you're thinking of, I'm happy to update it.

hamilton/lifecycle/default.py
JamesArruda: I just learned that Hamilton behaves nicely when an expand yields an empty list and takes that empty list to the collect! 

Since sentinel injection is a feature, a user might want to use it after a collect node. If the collect node got an empty list, there wouldn't be a way to tell if it's because the expand node didn't yield anything or if the expand node was skipped. So I lean toward single list being more explicit about what happened.

examples/parallelism/graceful_running/README.md
```diff
@@ -0,0 +1,31 @@
+# Parallelism with GracefulErrorAdapter Example
```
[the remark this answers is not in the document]
JamesArruda: [4308d0a](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/4308d0a83c200609ae45c02efa6dc97ec147456a)

hamilton/lifecycle/default.py
```diff
@@ -563,11 +569,19 @@ def never_reached(wont_proceed: int) -> int:
 
         Note you can customize the error you want it to fail on and the sentinel value to use in place of a node's result if it fails.
 
+        For Parallelizable nodes, this adapter will attempt to iterate over the node outputs. If an error occurs, the sentinel value is
+        returned and no more iterations over the node will occur. If you set ``fail_all_parallel`` to be True, it only sends on sentinel
```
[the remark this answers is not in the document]
JamesArruda: [4308d0a](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/4308d0a83c200609ae45c02efa6dc97ec147456a)

hamilton/lifecycle/default.py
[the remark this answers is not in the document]
JamesArruda: Attempted to make a decorator that creates a specific tag to allow sentinel injection.

[4308d0a](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/4308d0a83c200609ae45c02efa6dc97ec147456a)

hamilton/lifecycle/default.py
JamesArruda: [f6c7339](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/f6c7339fd47d7224d29b22621ffaeff506a7079b)

hamilton/lifecycle/default.py
JamesArruda: 4 and 5 will not be run.

Updated the docstring with an example:

[f7c0e94](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/f7c0e94de67aa969574e25ca2919768a534b8b31)

tests/test_parallel_graceful.py
JamesArruda: Moved the module over to resources.

[a299289](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/a299289c71558537909606e6c817a75280efa8f6)

hamilton/lifecycle/default.py
elijahbenizzy: Heh I had to test to ensure, and yes it turns out it does :)

I think that does make sense -- there's something very clean about a single sentinel value. That said, it *will* still execute on the executor (if this is, say, ray, that would be a slightly more expensive operation). But it'll look better in the UI and cascade it through. So yeah, I think a single list is OK here.

hamilton/lifecycle/default.py
```diff
@@ -584,10 +659,38 @@ def run_to_execute_node(
         # and truncate it/provide sentinels for every failure)
         # TODO -- decide what to do with collect
         """Executes a node. If the node fails, returns the sentinel value."""
-        for key, value in node_kwargs.items():
-            if value == self.sentinel_value:  # == versus is
-                return self.sentinel_value  # cascade it through
+        default_return = [self.sentinel_value] if is_expand else self.sentinel_value
+        _node_tags = future_kwargs["node_tags"]
+        can_inject = _node_tags.get("ERROR_SENTINEL", "false") == "True"
```
elijahbenizzy, line 664: Constant?

hamilton/lifecycle/default.py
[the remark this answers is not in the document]
elijahbenizzy: Yes, on second glance, I think you're right here, I was overly optimizing.

hamilton/lifecycle/default.py
```diff
@@ -512,13 +513,28 @@ def run_after_node_execution(
 SENTINEL_DEFAULT = None  # sentinel value -- lazy for now
 
 
+def accept_error_sentinels(func: Callable):
+    """Tag a function to allow passing in error sentinels.
+
+    For use with ``GracefulErrorAdapter``.
```
elijahbenizzy, line 519: Nit -- it's nice to add this to the docs -- here (https://github.com/DAGWorks-Inc/hamilton/blob/87c5de0a8a3ff8b31eb4946887735b46aa95fede/docs/reference/lifecycle-hooks/GracefulErrorAdapter.rst#L4). I'm happy to add it in afterwards cause it takes a bit of dev work to test out. Plus side is we can also cross-link the reference.

hamilton/function_modifiers/metadata.py
```diff
@@ -57,6 +57,7 @@ def my_function(...) -> ...:
         "ccpa",
         "dag",
         "module",
+        "ERROR_SENTINEL",
```
elijahbenizzy, line 60: I don't think this is a namespace? I'd set it as `hamilton.error_sentinel=True)` then keep the ypassing of reserved namespaces. That said, we can always change it later as its an internal contract...

hamilton/lifecycle/default.py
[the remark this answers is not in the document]
elijahbenizzy: Awesome, thanks! Would you mind adding a little documentation on that decorator with an example? I think its a little confusing what it does.

elijahbenizzy:
@JamesArruda nice work, almost there! Added some comments. Do let me know if you want me to take over getting it the last little bit (don't want to bog you down with nitpicks...).

hamilton/function_modifiers/metadata.py
JamesArruda: [96d9c47](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/96d9c47d9558529668ec30a78721baa9dd230794)

hamilton/lifecycle/default.py
JamesArruda: [96d9c47](https://github.com/DAGWorks-Inc/hamilton/pull/1017/commits/96d9c47d9558529668ec30a78721baa9dd230794)

hamilton/lifecycle/default.py
JamesArruda: If you don't mind adding that, I would appreciate the help. You'll get it done much faster, I think!

elijahbenizzy: You got it!

elijahbenizzy:
@JamesArruda -- turns out the test failure is coming from the inclusion of `@tag` as an import in `default.py` (yes, really, I downloaded it and messed around). 

This reeks of some sloppy dependency stuff elsewhere. I think this is due to a circular dependency, but I haven't been able to dig in. So, the cleanest way to do this is to move the decorator to somewhere in the function_modifiers package, and it'll be loosely coupled. Could also do a dynamic (inline) import. I think that might be cleaner with a TODO + reference to this conversation.

That said, I think this is looking pretty much ready, so I'm happy to merge this and then I'll take this the last mile. Sounds good?

JamesArruda:
@elijahbenizzy Sounds great, thanks!

PR #1031, opened 2024-07-17
fix: updated imports for polars extension 1.0­.0

zilto:
Polars upgrading to `>= 1.0.0` introduced breaking changes, in particular going from `polars.typing_aliases` to `polars._typing`.

> venv/lib/python3.11/site-packages/hamilton/plugins/polars_post_1_0_0_extension.py:597: DeprecationWarning: The `polars.type_aliases` module is deprecated. The type aliases have moved to the `polars._typing` module to explicitly mark them as private. Please define your own type aliases, or temporarily import from the `polars._typing` module. A public `polars.typing` module will be added in the future.
    from polars.type_aliases import ColumnTotalsDefinition, RowTotalsDefinition

We made changes accordingly, but missed two imports that were within a materializer definition. The warning is triggered when the materializer is registered during extension loading. This explains why we missed it and it wasn't covered by our tests.

Also, I renamed `polars_post_1_0_0_extension.py` to the plural `polars_post_1_0_0_extensions.py` to match convention.

### DIFF
--- a/hamilton/plugins/polars_extensions.py
+++ b/hamilton/plugins/polars_extensions.py
@@ -26,7 +26,7 @@ if pl_version < version.Version("1.0.0"):
         f"Current version: {pl_version}, minimum required version: 1.0.0."
     )
 else:
-    from hamilton.plugins.polars_post_1_0_0_extension import register_data_loaders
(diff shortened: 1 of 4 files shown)


elijahbenizzy (approved):
Thanks for the fix

PR #1035, opened 2024-07-18
Implements #710 optional value exposure

### LINKED ISSUE #710: Expose default values in `HamiltonNode`
We have `optional_dependencies` but its just a set. We don't expose it.

We should have a `default_values` field or something like that. `Dict[str, Any]`.

People can inspect the function -- see https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/plugins/h_experiments/hook.py for instance.

**Additional context**
Feature request from OS. This is tricky as we don't capture it, but should be possible with a bit of rewiring.

skrawcz:
This fixes #710.

This hasn't been extensively tested, but this
seems to work and pass things through as
expected. There is a JSON serialization
caveat, however we expect default
values to be primitives so that should
be fine. If not the user can augment the
dictionary returned ...

## Changes
 - graph and node

## How I tested this
 - via unit tests

## Notes
 - need to try examples and test more extensively

## Checklist


zilto:
Why would we need to pass this through `hamilton.node.Node` if no internal components rely on this info? Since the value is derived from the function signature, it can always be accessed through `originating_function`, both on `hamilton.node.Node` and `hamilton.graph_types.HamiltonNode`. The operation could be done more lazily rather than carrying it around (we already do that for `source` code)

```python
import inspect

class HamiltonNode:
  ...
  
  @functools.cached_property
  def default_values(self):
    sig = inspect.signature(self.originating_functions[0])
    return {
      param: param.default
      for param in sig.parameters
      if not param.default is inspect._empty
    }
```

Other points:
- these values may not be JSON serializable for `as_dict()`
- the function signature stores the values, not pointers, therefore default parameters aren't mutable. For example:

```python
DEFAULT_OBJ = "foo"

def my_func(a: Any = DEFAULT_OBJ):
  return a
  
DEFAULT_OBJ = object()
```
This would always return `"foo"` as default value for `a`

elijahbenizzy:
> Why would we need to pass this through `hamilton.node.Node` if no internal components rely on this info? Since the value is derived from the function signature, it can always be accessed through `originating_function`, both on `hamilton.node.Node` and `hamilton.graph_types.HamiltonNode`. The operation could be done more lazily rather than carrying it around (we already do that for `source` code)
> 
> ```python
> import inspect
> 
> class HamiltonNode:
>   ...
>   
>   @functools.cached_property
>   def default_values(self):
>     sig = inspect.signature(self.originating_functions[0])
>     return {
>       param: param.default
>       for param in sig.parameters
>       if not param.default is inspect._empty
>     }
> ```
> 
> Other points:
> 
> * these values may not be JSON serializable for `as_dict()`. I expect this to be particularly common for scikit-learn and LangChain-type workflows where a processor can be swapped on-the-fly (mock example)
> 
> ```python
> def cleaned_data(data: Any, processor: BaseEstimator = sklearn.Imputer(mode="mean")):
>   processor.fit_transform(data)
>   return data
> ```
> 
> * the function signature stores the values, not pointers, therefore default parameters aren't mutable. For example:
> 
> ```python
> DEFAULT_OBJ = "foo"
> 
> def my_func(a: Any = DEFAULT_OBJ):
>   return a
>   
> DEFAULT_OBJ = object()
> ```
> 
> This would always return `"foo"` as default value for `a`. This means we can use `@cached_property` with confidence

While, in most cases, we probably can derive it, I'm not sure we should. There isn't a 1:1 value mapping with it -- we'd be doing extra work to derive it, and making assumptions about the originating function. This information is exposed in the node, so we don't want to re-derive it: https://github.com/DAGWorks-Inc/hamilton/blob/3b2253b1f703fd940da97a356eb4a2ac3aa86691/hamilton/node.py#L35.

hamilton/graph_types.py
```diff
@@ -100,15 +100,20 @@ class HamiltonNode:
...
+
+        Note: optional values could be anything and might not be JSON serializable.
+        """
         return {
             "name": self.name,
             "tags": self.tags,
             "output_type": (get_type_as_string(self.type) if get_type_as_string(self.type) else ""),
             "required_dependencies": sorted(self.required_dependencies),
             "optional_dependencies": sorted(self.optional_dependencies),
+            "optional_input_values": self.optional_input_values,
```
skrawcz, line 116: `optional_dependency_default_values`

hamilton/graph_types.py
```diff
@@ -100,15 +100,20 @@ class HamiltonNode:
     documentation: typing.Optional[str]
     required_dependencies: typing.Set[str]
     optional_dependencies: typing.Set[str]
+    optional_input_values: typing.Dict[str, typing.Any]
 
     def as_dict(self):
```
skrawcz, line 105: add `to_json()`

skrawcz, line 105: add `include_optional_default_values=False`

elijahbenizzy (approved):

Issue #751, opened 2024-03-08
Add version field to data saver/ loader metadata


skrawcz:
**Is your feature request related to a problem? Please describe.**
The data save and loaders return metadata. We should version this schema so that we can iterate and make changes-- people can then know how to version/write code against it.

**Describe the solution you'd like**
Add a `__version__` field or something. It should follow semantic versioning. Major = breaking. Minor = new addition. Patch = bug fix.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

roothsmz:
@skrawcz ,

One thought is to provide for an overridable Callable object.

```
def get_dataframe_metadata(df: pd.DataFrame, metadata_func: Optional[Callable[[pd.DataFrame], Dict[str, Any]]] = None) -> Dict[str, Any]:
    """Gives metadata from loading a dataframe.

    Note: we reserve the right to change this schema. So if you're using this come
    chat so that we can make sure we don't break your code.

    This will default to include:
    - the number of rows
    - the number of columns
    - the column names
    - the data types
    
    If you provide an override, then your DataFrame will be passed into that lambda and the results will be as you decide.
    """
    if metadata_func is not None:
        return {DATAFRAME_METADATA: metadata_func(df)}
    else:
        return {
            DATAFRAME_METADATA: {
                "rows": len(df),
                "columns": len(df.columns),
                "column_names": list(df.columns),
                "datatypes": [str(t) for t in list(df.dtypes)],  # for serialization purposes
            }
        }
```

PR #1036, opened 2024-07-18
Add version field to metadata collected from files and dataframes

skrawcz:
Adds version field to metadata collected for files and dataframes.

Implements https://github.com/DAGWorks-Inc/hamilton/issues/751.

This is to formalize schema at a Hamilton level.

We can always extend this to have someone provide
their own callback function to produce this.

We use the __ syntax to help show that this is internal.

## Changes
io/utils metadata functions.

## How I tested this
 - locally via unit tests

## Checklist


elijahbenizzy (approved):
Seems reasonable

Issue #743, opened 2024-03-05
Add ability to mark function outputs as unserializable 


skrawcz:
**Is your feature request related to a problem? Please describe.**
In Parallel & Collect blocks, there could be a process boundary. Inputs that are invariant to each chunk of the block need to be serialized and deserialized -- but for things like clients (e.g. openai, db client) those objects aren't serializable and things break.

**Describe the solution you'd like**
```python

@unserializable
def client() -> object:
   """Some client, e.g. openai, etc"""
   return ... 
```
Then Hamilton will know that for each parallel block, we rerun this function, versus trying to run it once and trying to serialize and deserialize it.

**Describe alternatives you've considered**
You can get around this by making the `client()` function depend on some part of the parallel block thereby forcing it to be run each block.

**Additional context**
This would be a quality of life improvement.

elijahbenizzy:
Similar to this: https://github.com/DAGWorks-Inc/hamilton/issues/90

skrawcz:
Another alternative is to create a wrapper object and use that instead:

```python
class SerializeableOpenAIClient:
    def __init__(self):
        self.client = OpenAI()

    def __getstate__(self):
        return {}

    def __setstate__(self, d):
        self.client = OpenAI()

def client() -> SerializeableOpenAIClient:
    return SerializeableOpenAIClient()
```

parthrastogicoder:
.take

Issue #940, opened 2024-06-05
Better support for caching / checkpointing development workflow - umbrella issue


skrawcz:
**Is your feature request related to a problem? Please describe.**
We need a simpler caching & checkpointing story that has full service visibility into what's going on.

**Describe the solution you'd like**

1. Checkpointing -- i.e. cache outputs and restart from the latest point.
3. Intelligent Caching -- i.e. cache nodes and only rerun things if code or data has changed.

These should come with the ability to:

1. visualize what is going on when using them.
2. work in a notebook / cli / library context
4. extend how to hash data & where to store it

**Prior art**
1. You do it yourself outside of Hamilton and use the [overrides](https://hamilton.dagworks.io/en/latest/reference/drivers/#short-circuiting-some-dag-computation) argument in .execute/.materialize(..., overrides={...}) to inject pre-computed values into the graph. That is, you run your code, save the things you want, and then you load them and inject them using overrides=. TODO: show example.
2. You use the data savers & data loaders. This is similar to the above, but instead you use the [Data Savers & Data Loaders (i.e. materializers)](https://hamilton.dagworks.io/en/latest/reference/io/available-data-adapters/) to save & then load and inject data in. TODO: show example.
3. You use the CachingGraphAdapter, which requires you to tag functions to cache along with the serialization format.
5. You use the DiskCacheAdapter, which uses the diskcache library to store the results on disk.

Could use https://books.ropensci.org/targets/walkthrough.html#change-code as inspiration.

**Additional context**
Slack threads:
 - https://hamilton-opensource.slack.com/archives/C03M33QB4M8/p1717602458129939

## Next steps:

TODO: write up tasks in this issue into smaller and manageable chunks.

jmbuhr:
I took one of the `targets` examples and transferred it one-to-one to `hamilton` to see how the concepts compare. Both workflows are implemented in modules and make use of helper functions from a separate module. Then both are started interactively from quarto documents and their results and graphs visualized in the rendered output of said notebooks: http://jmbuhr.de/targets-hamilton-comparison/ (source code: https://github.com/jmbuhr/targets-hamilton-comparison)

Again, this is for exploration of possibilities, not to impose paradigms on you :)

In this first pass I noticed two things I was missing in hamilton compared to targets when it comes to caching:

- changing a function that is used by a node, but is not itself a node, should also invalidated the cache of the node
[...]

PR #1039, opened 2024-07-18
Feat/fingerprinting

zilto:
Following up our design discussion on caching and fingerprinting. Here are the basic working mechanisms for caching as we discussed.

Composable parts:
- caching algorithm: decide when to compute or use cache
- fingerprint store (repository): how to store `{execution context: fingerprint}`
- result store (cache): how to store `{fingerprint: result / data}`
- serialization and deserialization (related to result store)

TODO
- figure out ideal module structure
- documentation
- standardize and agree on naming
- introspection mechanisms
- good logger for debugging
- special caching algorithm behavior
  - always recompute: does the recompute and produces a new fingerprint; this fingerprint affects the keys of downstream nodes
  - don't fingerprint: can be recomputed or read from cache; in the eye of the downstream nodes, the value / fingerprint is constant (related: #743 )
  - read historical value: if a value exists, use it (i.e., the existing CachingGraphAdapter). It would require storing a `node_name` metadata in the fingerprint store. This is very easy with `diskcache`
- surface to user post-run caching explanations (visualization, rule-based text explanations)
- surface to user pre-run caching explanations; this will use node versions (must recompute), top-level inputs and overrides, and the state of the fingerprint store (if it's empty, we know we must recompute everything same if `node_name` metadata is absent)
- checkpointing is a subset of this cache feature. All you need to do is explicitly pass a mapping of fingerprints to `SmartCache(fingerprints={...})` (which could be hidden beind a single `run_id`). Then the adapter will use these fingerprints to read from the result store instead of trying to compute these fingerprints or load them from the fingerprint store.
- improve the protocol definition for `Store` objects. `get` and `set` are required, but unclear if `open()` and `close()` are necessary for all (could be null operations)
- create a `MaterializerStore` where `get()` and `set()` are materializers


skrawcz:
this is related to https://github.com/DAGWorks-Inc/hamilton/issues/940

PR #1067, opened 2024-07-26
Adds NoEdgeAndInputTypeChecking adapter

skrawcz:
So that people can turn off type checking if they really want to.

This won't necessarily mean you can get away without types -- it just means they don't need to match.

## Changes
 - adds adapter
 - updates docs

## How I tested this
- locally

## Checklist

### DIFF
--- a/docs/reference/lifecycle-hooks/NoEdgeAndInputTypeChecking.rst
+++ b/docs/reference/lifecycle-hooks/NoEdgeAndInputTypeChecking.rst
@@ -1,0 +1,9 @@
+====================================
+lifecycle.NoEdgeAndInputTypeChecking
+====================================
+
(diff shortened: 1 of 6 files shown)


docs/reference/lifecycle-hooks/NoEdgeAndInputTypeChecking.rst
```diff
@@ -0,0 +1,9 @@
+====================================
+lifecycle.NoEdgeAndInputTypeChecking
+====================================
+
+Use this hook turn off edge and input type checking during graph construction and execution
```
elijahbenizzy, line 5: Might be worth saying that this isn't recommended

elijahbenizzy (approved):

PR #1068, opened 2024-07-26
Plugin/mlflow custom flavor - fixed small bug

jernejfrank:
Thanks for adding the custom flavor option to mlflow. I tested it and only the flavor specific kwargs somehow got mixed up and tried to be called somewhere else which lead to a fatal error.

## Changes
- renamed class attribute to avoid conflicting calls

## How I tested this
- added unit test that doubles as minimal example for reproducing the error

## Notes
- only tested the DataSaver (due to internal legacy structure, DataLoader is not utilised/hard to implement)

## Checklist


skrawcz:
Thanks @jernejfrank ! For the pre-commit failure you just need to do something like:

`pip install pre-commit`
`pre-commit install`
`pre-commit run --all-files`
And then check that change in  and you should be good to go. Thanks for the fix!

skrawcz:
I'll plan to push this in the release on Monday. If you don't get to the `pre-commit` before I'll do it. Thanks!

examples/hamilton_ui/data_set_v1.parquet
skrawcz, line 1: delete this please

examples/hamilton_ui/lr_model.pkl
skrawcz, line 1: delete this change please

examples/hamilton_ui/svm_model.pkl
skrawcz, line 1: delete this change please

skrawcz (changes requested):
looks good. just a few things:

1. pre-commit.
2. you added some files in by accident I assume from the hamilton_ui example.

examples/hamilton_ui/data_set_v1.parquet
jernejfrank: missed this got generated, sorry.

examples/hamilton_ui/lr_model.pkl
jernejfrank: same here.

examples/hamilton_ui/svm_model.pkl
jernejfrank: and here.

skrawcz (approved):

skrawcz:
Thanks @jernejfrank !

PR #1081, opened 2024-08-05
Adds code for pregnancy due date modeling blog post

elijahbenizzy:
Models pregnancy due dates with some assumptions in a notebook using Hamilton.


skrawcz (changes requested):
Can you:

1. Add google collab link to readme
2. In the top of the notebook include pip install requirements (e.g. to run in the notebook)
3. add google collab link in notebook
4. drop `-m` from the cell_to_module magic -- don't need that flag anymore, module name is the first argument.
5. an intro to the notebook saying what it is and how to use it.
6.

skrawcz (approved):

PR #1088, opened 2024-08-07
Transition to packaging & configuration using pyproject.toml

Dev-iL:
Transition to the more modern `pyproject.toml` from `setup.cfg`, `requirements-*.txt` etc.

## Changes
Combined multiple files into one + added some placeholder for ruff & mypy.

## How I tested this

```bash
python -m build --wheel
```

## Notes
Addresses #10

## Checklist

- [x] PR has an informative and human-readable title (this will be pulled into the release notes)
- [x] Changes are limited to a single goal (no scope creep)
- [x] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [ ] Any _change_ in functionality is tested
- [ ] New functions are documented (with a description, list of inputs, and expected output)
- [ ] Placeholder code is flagged / future TODOs are captured in comments
- [ ] Project documentation has been updated if adding/changing functionality.

PR #1089, opened 2024-08-08
Use Ruff instead of Black+isort+Flake8

Dev-iL:
[Ruff](https://docs.astral.sh/ruff/) is faster, and some might say more modern, alternative to multiple other tools.
- This PR turns on a subset of Ruff that is equivalent to Black, isort, and no-plugin flake8, combined.
- This is a continuation of #1088.

## Changes
(See individual commit titles)

## How I tested this
Evaluated the pre-commit hooks until they passed.

## Notes
As discussed on Slack, some inspections, when enabled (or unexcluded), detect a lot of violations, so enabling rules and rule families will need to be done carefully. This will require an additional effort, potentially outside the scope of this PR.

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,21 +5,16 @@
 #   $ pre-commit install
 exclude: '^docs/code-comparisons/'  # skip the code comparisons directory
 repos:
--   repo: https://github.com/ambv/black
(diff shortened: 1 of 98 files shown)


skrawcz (approved):
Exciting!

Issue #1059, opened 2024-07-23
Slow import of hamilton's Builder


giladrubin1:
## Current behavior
Importing the `Builder` from `hamilton.driver` takes approximately 4-5 seconds, while the rest of the package imports are instant.

### Steps to replicate behavior
1. Open a VSCode notebook
2. Run the following import:
   ```python
   from hamilton.driver import Builder
   ```
3. Observe that the import takes 4-5 seconds to complete

## Library & System Information
- Hamilton library version: 1.72.1
- Environment: VSCode notebooks

## Expected behavior
The import of `Builder` from `hamilton.driver` should be as fast as other package imports.

## Additional context
This slow import is specifically noticed when using VSCode notebooks. It would be helpful to know if this behavior is reproducible in other environments or if it's specific to VSCode notebooks.

skrawcz:
Thanks @giladrubin1 .

This is likely the plugins being loaded (my hypothesis). If you have a lot of libraries in your environment this can happen - I assume you have a lot in your python environment? 

Otherwise the plugin loading could be optimized for sure.

giladrubin1:
I do have several libraries in my environment... I did some basic testing with import-waterfall and if I read the graph right - it does show that the import is affected by other libraries. 

However, I would expect Hamilton's builder to not be affected by other libraries in this way. Especially since most of them are not used in the context of that driver.

skrawcz:
Yes, agreed. If you haven't imported them beforehand, Hamilton shouldn't
either -- that's at least my view.

Dev-iL:
@giladrubin1 which python version are you on? If >=3.12, could you please see if [enabling lazy imports](https://peps.python.org/pep-0690/#specification) helps?

```python
importlib.set_lazy_imports()
```

giladrubin1:
I'm using python 3.10

Dev-iL:
That's fine - disregard that suggestion - looks like that PEP was rejected anyway. Maybe something like [this](https://stackoverflow.com/a/54616590/3372061) could work:

```python
def __getattr__(name):
    if name == "module_xyz_it_takes_long_to_load":
        return _lazy_init_function_for_xyz()

    # Implicit else
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```
 I'll give that approach a try in one of my personal projects and see if it's worth considering.

giladrubin1:
I want to emphasize that I'm not looking for a workaround.
[...]

PR #1095, opened 2024-08-14
added environment variable to disable extensions autoload

zilto:
The current approach to loading plugins is greedy. When doing `import hamilton.driver`, it tries to import all the extension modules available to the users (e.g., pandas, matplotlib, mlflow, pyspark) even if these libraries aren't used in the dataflow.

This can introduce slowdowns, memory usage (many Mbs), and warning messages by 3rd party libraries (e.g., `PYARROW_TIMEZONE` by pyspark).

Since many features (`@extract_columns`, materializers, and more) rely on this extension registry, changing the default behavior would introduce many hard-to-track breaking changes. 

Instead, we'll adopt an opt-out approach. Users can now do:
1. disable autoload programmatically
  ```python
  from hamilton import registry
  registry.disable_autoload()
  ```
2. set an environment variable
  ```bash
  export HAMILTON_AUTOLOAD_EXTENSIONS=0
  ```
  or (note that this is a string)
  ```python
  import os
  os.environ["HAMILTON_AUTOLOAD_EXTENSIONS"] = "0"
  ```
  or in jupyter notebooks
  ```python
  %env HAMILTON_AUTOLOAD_EXTENSIONS=0
  ```
3. set a config value that will persist and be used whenever Hamilton is used via the command line
  ```bash
  hamilton-disable-autoload-extensions
  ```
  and re-enable it via
  ```bash
  hamilton-enable-autoload-extensions
  ```
  
Then, users that disabled autoload would need to either:
1. directly import hamilton plugins anywhere in their code
  ```python
  from hamilton.plugins import mlflow_extensions
  ```
2. directly register extensions in the registry (has good IDE support via `typing.Literal`)
  ```python
  from hamilton import registry
  registry.load_extension("mlflow")
  ```
  
Relates to #1059.

## Changes
- moved the content of `hamilton.function_modifiers.base` to `hamilton.registry`
- added entrypoints to  `pyproject.toml`
- added functions to `hamilton.registry` to interact with the `~/.hamilton.conf`

## How I tested this
- tested some code path manually in a notebook and enabling/disabling via the CLI

## Checklist

### DIFF
--- a/hamilton/registry.py
+++ b/hamilton/registry.py
@@ -1,19 +1,129 @@
 import collections
+import configparser
 import functools
 import importlib
 import logging
-from typing import Any, Dict, Optional, Type
+import os
+import pathlib
+import sys
+from typing import Any, Dict, Literal, Optional, Tuple, Type, get_args
 
 logger = logging.getLogger(__name__)
 
 # Use this to ensure the registry is loaded only once.
 INITIALIZED = False
+ExtensionName = Literal[
+    "yaml",
+    "matplotlib",
+    "numpy",
+    "pandas",
+    "plotly",
+    "polars",
+    "polars_lazyframe",
+    "pyspark_pandas",
+    "spark",
+    "dask",
+    "geopandas",
+    "xgboost",
+    "lightgbm",
+    "sklearn_plot",
+    "vaex",
+    "ibis",
+    "dlt",
+    "kedro",
+    "huggingface",
+    "mlflow",
+]
+HAMILTON_EXTENSIONS: Tuple[ExtensionName, ...] = get_args(ExtensionName)
+HAMILTON_AUTOLOAD_CONFIG_KEY = "autoload_extensions"
+HAMILTON_AUTOLOAD_ENV = "HAMILTON_AUTOLOAD_EXTENSIONS"
+DEFAULT_CONFIG_LOCATION = pathlib.Path("~/.hamilton.conf").expanduser()
 
 # This is a dictionary of extension name -> dict with dataframe and column types.
 DF_TYPE_AND_COLUMN_TYPES: Dict[str, Dict[str, Type]] = {}
 
 COLUMN_TYPE = "column_type"
 DATAFRAME_TYPE = "dataframe_type"
+
+
+def load_extension(plugin_module: ExtensionName):
+    """Given a module name, loads it for Hamilton to use.
+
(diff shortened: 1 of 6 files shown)


hamilton/registry.py
```diff
@@ -16,6 +45,86 @@
...
+        ), f"Error extension missing fill_with_scalar_{plugin_module}"
+        logger.info(f"Detected {plugin_module} and successfully loaded Hamilton extensions.")
+
+
+def initialize():
+    load_autoload_config()
+    logger.debug(f"{HAMILTON_AUTOLOAD_ENV}={os.environ.get(HAMILTON_AUTOLOAD_ENV)}")
+    for extension_name in HAMILTON_EXTENSIONS:
+        # skip modules that aren't explicitly imported by the user
+        if os.environ.get(HAMILTON_AUTOLOAD_ENV) == "0" and extension_name not in sys.modules:
```
skrawcz, line 74: wont this skip importing names like polars_lazyframe ?

skrawcz: actually why is the second clause needed? I thought the API is that someone has to manually register things? either via load_extension, or importing the extension which will register itself.

zilto: Ok, maybe we have different specs. The behavior I implemented is described in the detailed in the first PR message.

### Current behavior
- When `import hamilton.function_modifiers.base`: autoload all plugins by try/except `load_extension()`
- Can't be disabled or changed

### This PR
the current behavior can be disabled to instead do the following:
- When `import hamilton.function_modifiers.base`: autoload all plugins that have been imported in `sys`. In other words,
```python
import matplotlib
from hamilton import driver
```
will import `matplotlib_extensions.py`
- after `import hamilton.function_modifiers.base`, you would need to register other extensions via `import hamilton.plugins.matplotlib_extensions` or `registry.register_extension("matplotlib")`

### The behavior you are implying
disable the current behavior can be disabled to remove *all autoloading* and only allow explicit `import hamilton.plugins.matplotlib_extensions` or `registry.register_extension("matplotlib")`

## Conclusion
I realize that the phrasing "disable autoload" might be unappropriate for this PR's behavior. Should we have 3 modes instead or would that be confusing?

1. Greedy: current behavior, default
2. Lazy: allows autoload for `import matplotlib` then `from hamilton import driver`
3. Disabled: explicit registration only

Lazy might be more valuable for development and Disabled better in prod ("disable all magic").

Side note: this is probably also why I find the CLI so slow

skrawcz (approved):

PR #1097, opened 2024-08-15
Update graph_functions.py

elijahbenizzy:
What to do for adding tracking support for ray remote


elijahbenizzy:
See commits for all the TODO.

One thing to note -- the lifecycle method should take in a function as well as the node, that's the function that you define in graph_functions.

Then to test, just run it -- we can help add some unit tests if you want, there's a few properties we want to make sure work

PR #1100, opened 2024-08-18
Adds setstate getstate to driver and fixes 1093

### LINKED ISSUE #1093: Multi-processing pickle issue with pickling driver within Hamilton
# Current behavior
This breaks on the below code -- it's hamilton within hamilton.

## Stack Traces
Pickle error on Module Type.

The mapper-worker-reducer pattern works well on multithreadingexecutor, but not work for multiprocessingexecutor. The error is cannot pickle 'module' object. Here is a minimal reproduce environment:

```python
import worker

def double(a:int) -> int:
    return a*2
```
```python
import mapper

from typing import Any

def mapper(
    drivers: list,
    inputs: list,
    final_vars: list = [],
) -> Parallelizable[dict]:
    for dr, input_ in zip(drivers, inputs):
        yield {
            "dr": dr,
            "final_vars": final_vars or dr.list_available_variables(),
            "input": input_,
        }


def worker(mapper: dict) -> dict:
    _dr = mapper["dr"]
    _inputs = mapper["input"]
    _final_var = mapper["final_vars"]
    return _dr.execute(final_vars=_final_var, inputs=_inputs)


def reducer(worker: Collect[dict]) -> Any:

    return worker
```
```python
from hamilton import driver
from hamilton.execution import executors

drivers = []
inputs = []
for i in range(4):
    dr = driver.Builder().with_modules(worker).build()
    drivers.append(dr)
    inputs.append({'a': i})


dr = (
    driver.Builder()
    .with_modules(mapper)
    .with_local_executor(executors.SynchronousLocalTaskExecutor())
    .with_remote_executor(executors.MultiProcessingExecutor(8))
    .build()
)
dr.execute(
    final_vars=["reducer"],
    inputs={"drivers": drivers, "inputs": inputs, "final_vars": ['double']},
)
```

Latest.

# Expected behavior
This work.

# Additional context

skrawcz:
This fixes #1093.

Modules were not picklable. So this fixes that by serializing their fully qualified names, and then when the driver object is deserialized, they are reinstantiated as module objects.

## Checklist

### DIFF
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -263,21 +263,31 @@ class Driver:
     """
 
     def __getstate__(self):
+        """Used for serialization."""
         # Copy the object's state from self.__dict__
         state = self.__dict__.copy()
         # Remove the unpicklable entries -- right now it's the modules tracked.
-        state["graph_module_names"] = [
+        state["__graph_module_names"] = [
             importlib.util.find_spec(m.__name__).name for m in state["graph_modules"]
         ]
         del state["graph_modules"]  # remove from state
         return state
 
     def __setstate__(self, state):
+        """Used for deserialization."""
         # Restore instance attributes
         self.__dict__.update(state)
         # Reinitialize the unpicklable entries
         # assumption is that the modules are importable in the new process
-        self.graph_modules = [importlib.import_module(n) for n in state["graph_module_names"]]
+        self.graph_modules = []
+        for n in state["__graph_module_names"]:
+            try:
+                g_module = importlib.import_module(n)
+            except ImportError:
+                logger.error(f"Could not import module {n}")
+                continue
+            else:
+                self.graph_modules.append(g_module)
 
     @staticmethod
     def normalize_adapter_input(
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -1,5 +1,7 @@
 import abc
 import functools
+import importlib
+import importlib.util
 import json
 import logging
 import operator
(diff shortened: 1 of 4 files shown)


hamilton/driver.py
```diff
@@ -260,6 +262,23 @@ class Driver:
...
+        ]
+        del state["graph_modules"]  # remove from state
+        return state
+
+    def __setstate__(self, state):
+        # Restore instance attributes
+        self.__dict__.update(state)
+        # Reinitialize the unpicklable entries
+        # assumption is that the modules are importable in the new process
+        self.graph_modules = [importlib.import_module(n) for n in state["graph_module_names"]]
```
elijahbenizzy, line 280: We should add a try/catch with an error and instructions on how to do this. Error in `__setstate__` can ber confusing.

hamilton/driver.py
```diff
@@ -260,6 +262,23 @@ class Driver:
         dr = driver.Driver(config, module, adapter=adapter)
     """
 
+    def __getstate__(self):
+        # Copy the object's state from self.__dict__
+        state = self.__dict__.copy()
+        # Remove the unpicklable entries -- right now it's the modules tracked.
+        state["graph_module_names"] = [
```
elijahbenizzy, line 269: Should be `__` start or something to show that this is special

tests/test_hamilton_driver.py
```diff
@@ -665,3 +667,29 @@ def func_to_test(a: int) -> int:
...
+        dr = Builder().with_modules(tests.resources.test_driver_serde_worker).build()
+        drivers.append(dr)
+        inputs.append({"a": i})
+
+    dr = (
+        Builder()
+        .with_modules(tests.resources.test_driver_serde_mapper)
+        .enable_dynamic_execution(allow_experimental_mode=True)
+        # .with_local_executor(executors.SynchronousLocalTaskExecutor())
+        .with_remote_executor(executors.MultiProcessingExecutor(8))
```
elijahbenizzy, line 688: Why 8?

skrawcz: no reason. changed to 4.

elijahbenizzy (approved):

PR #1103, opened 2024-08-20
Refactor to enable RayGraphAdapter and HamiltonTracker to work well together

jernejfrank:
Hi Elijah,

I have done some things, but got stuck in what seems to be the final stages (unless I missed something). Please let me know if you have some time this week to discuss.

## Changes
- added inline function to execute lifecycle hooks in remote environment
- added new lifecycle class for remote execution that passes the inline function to the adapters that handle remote environments
- changed RayGraphAdapter to execute the full lifecycle (including pre- and post-hooks) within Ray remote

## How I tested this
- There is a script ```z_test_implementation.py``` that executes a single node in ray remote and waits for 5s to test the telemetry in HamiltonUI

## Notes
- Unable to overcome
```
AssertionError: The @ray.remote decorator must be applied either with no arguments and no parentheses
```
- Unsure if I handled catching errors, success, results, exceptions with the inline function in ```graph_functions``` correctly.
- Added ```RayGraphAdapter.do_remote_execute(..., **future_kwargs)``` to collect things such as ```run_id``` and ```task_id``` that are not used in function.
## Checklist


hamilton/execution/graph_functions.py
```diff
@@ -214,57 +215,102 @@ def dfs_traverse(
...
+                    execute_lifecycle_for_node = execute_lifecycle_for_node(
+                        node_ = node_,
+                        adapter = adapter,
+                        run_id = run_id,
+                        task_id = task_id,
+                        kwargs = kwargs,
+                        success=success,
+                        pre_node_execute_errored=pre_node_execute_errored,
+                        error=error,
+                        result=result),
```
skrawcz, line 296: @jernejfrank I think you're actually calling the function here -- what I think we want is to pass a pointer to the function, along with the arguments for it.

skrawcz: Some of these could be bound via functools.partial

hamilton/plugins/h_ray.py
```diff
@@ -105,25 +111,36 @@ def __init__(self, result_builder: base.ResultMixin):
...
         """Function that is called as we walk the graph to determine how to execute a hamilton function.
 
-        :param node: the node from the graph.
+        :param execute_lifecycle_for_node: wrapper function that executes lifecycle hooks and methods
         :param kwargs: the arguments that should be passed to it.
         :return: returns a ray object reference.
         """
         ray_options = parse_ray_remote_options_from_tags(node.tags)
-        return ray.remote(raify(node.callable)).options(**ray_options).remote(**kwargs)
+        return ray.remote(raify(execute_lifecycle_for_node))#.options(**ray_options).remote()#**kwargs)
```
skrawcz, line 142: So here we want to:

1. raify `execute_lifecycle_for_node` -- this will inject `ray.remote` around that function.
2. pass in options from node tags
3. then  **kwargs here should include all the arguments to execute `execute_lifecycle_for_node`

skrawcz: So the trickiest part here is connecting the arguments required for the new function, as well as the arguments required for the actual node function call itself.

hamilton/execution/graph_functions.py
jernejfrank: > @jernejfrank I think you're actually calling the function here -- what I think we want is to pass a pointer to the function, along with the arguments for it.

@skrawcz thanks a lot for pointing that out, I wasn't exactly sure where the arguments need to be passed into the Callable.



> Some of these could be bound via functools.partial

We discussed that option with @elijahbenizzy , but agreed that we both find inline functions clearer to read vs partial. Having said that, I am happy to change that since I have never done partials and would learn something new!

hamilton/execution/graph_functions.py
```diff
@@ -250,7 +258,7 @@ def execute_lifecycle_for_node(
                     else:
                         result = node_(**kwargs)
                     
-                    return error, result, success, pre_node_execute_errored
+                    return result#, success, pre_node_execute_errored
```
jernejfrank, line 261: I am struggling to capture these, since now ```ray.remote``` is returning the reference that needs to be passed into ```ray.get``` to get the results of the 'sandwich' function that helps us get the hooks through. I feel like there a step missing in ```RayGraphAdapter.build_result()``` where these would be unpacked and passed back to ```execute_subdag```

hamilton/execution/graph_functions.py
skrawcz: yep so you can either:

1. Bind some of the arguments here -- i.e. the ones that don't depend on the inputs to the node function callable.
2. Or pass in a dictionary of those arguments for the function.

In either case, you need to construct the right kwargs for the `.remote(...)` call.

hamilton/execution/graph_functions.py
skrawcz: so the way it works right now is that we get ray object references back. The `.remote(**kwargs)` call on the ray.remote function is where Ray figures out what to resolve. 

I think we want the `num_returns` construct here in that case.

```
@ray.remote(num_gpus=1, max_calls=1, num_returns=2)
def f():
    return 1, 2

f_with_2_gpus = f.options(num_gpus=2)
object_refs = f_with_2_gpus.remote()
assert ray.get(object_refs) == [1, 2]
```
i.e. you'll then get separate object references for each return value that you can wire through.

hamilton/execution/graph_functions.py
```diff
@@ -270,7 +260,13 @@ def execute_lifecycle_for_node(
                     raise
                 finally:
```
jernejfrank, line 261: I played around and got a deterministic outcome for this try-except-finally clause:
1) If the node executes normally, finally does not get called
2) If node raises exception finally gets called

ui/sdk/src/hamilton_sdk/adapters.py
```diff
@@ -250,6 +250,7 @@ def post_node_execute(
         result: Optional[Any],
         task_id: Optional[str] = None,
     ):
+        print("I got to hamilton tracker")
```
jernejfrank, line 253: We never make it to here, so the UI shows the nodes as still running no matter what happened.

hamilton/execution/graph_functions.py
jernejfrank: Opposed to this case where there is no ValueError.
```
2024-08-21 23:43:49,243 INFO worker.py:1781 -- Started a local Ray instance.

Capturing execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/65


Captured execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/65

(execute_lifecycle_for_node pid=361616) try lifecycle
(execute_lifecycle_for_node pid=361616) pre-hook
(execute_lifecycle_for_node pid=361616) try pre-hook
(execute_lifecycle_for_node pid=361616) node execute as callable
(execute_lifecycle_for_node pid=361616) 5s executed
    node_5s
0  5.025094
```

jernejfrank (commented):
Ignore, I had a local mistake that the Tracker was called from pip installed package and not the executable dir...

z_test_implementation.py
```diff
@@ -1,63 +1,65 @@
...
-def node_5s()->float:
+def node_5s() -> float:
     print("5s executed")
     start = time.time()
     time.sleep(5)
     return time.time() - start
 
-def node_5s_error()->float:
+
+def at_1_to_previous(node_5s: float) -> float:
```
jernejfrank, line 11: Added to see how two connecting nodes would respond and ran into a different issue:

```
Captured execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/101

(execute_lifecycle_for_node pid=631084) 5s executed
                                    at_1_to_previous
0  ObjectRef(c8ef45ccd0112571ffffffffffffffffffff...
```

z_test_implementation.py
```diff
@@ -1,63 +1,65 @@
 import time
 
 
-def node_5s()->float:
+def node_5s() -> float:
```
jernejfrank, line 4: This works now and tracks correctly on the node level in the ui

```
Captured execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/102

(execute_lifecycle_for_node pid=631939) 5s executed
    node_5s
0  5.025091
    node_5s
0  5.025091
```

z_test_implementation.py
elijahbenizzy: Hmm, the problem is it's getting passed an object ref rather than a function? Is this reflected in the UI? Or just the result?

jernejfrank: Here's the weird part that I am troubleshooting at the moment. Asking for the ```node_5s``` displays correctly on the UI that it took 5s to run.

I then added this function into the DAG
```
def at_1_to_previous(node_5s: float) -> float:
    print("Value of 5s ", node_5s)
    print("1s executed")
    start = time.time()
    time.sleep(1)
    return node_5s   + (time.time()-start)
```

If I request ```at_1_to_previous``` I get an error message and the UI shows the run but no nodes are listed as executed or failed, it's just an empty waterfall and list.

```
2024-08-24 21:56:32,934 INFO worker.py:1781 -- Started a local Ray instance.

Capturing execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/111


Captured execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/111

(execute_lifecycle_for_node pid=661223) Value of 5s  ObjectRef(c8ef45ccd0112571ffffffffffffffffffffffff0100000001000000)
(execute_lifecycle_for_node pid=661223) 1s executed
-------------------------------------------------------------------
Oh no an error! Need help with Hamilton?
Join our slack and ask for help! https://join.slack.com/t/hamilton-opensource/shared_invite/zt-2niepkra8-DGKGf_tTYhXuJWBTXtIs4g
-------------------------------------------------------------------

  File "/home/jfrank/code/hamilton/z_test_implementation.py", line 46, in <module>
    result_ray = dr_ray.execute(
  File "/home/jfrank/code/hamilton/hamilton/driver.py", line 567, in execute
    raise e
  File "/home/jfrank/code/hamilton/hamilton/driver.py", line 560, in execute
    return self.adapter.call_lifecycle_method_sync("do_build_result", outputs=outputs)
  File "/home/jfrank/code/hamilton/hamilton/lifecycle/base.py", line 959, in call_lifecycle_method_sync
    return getattr(adapter, method_name)(**kwargs)
  File "/home/jfrank/code/hamilton/hamilton/plugins/h_ray.py", line 154, in do_build_result
    result = ray.get(remote_combine)  # this materializes the object locally
  File "/home/jfrank/miniconda3/envs/hamilton/lib/python3.10/site-packages/ray/_private/auto_init_hook.py", line 21, in auto_init_wrapper
    return fn(*args, **kwargs)
  File "/home/jfrank/miniconda3/envs/hamilton/lib/python3.10/site-packages/ray/_private/client_mode_hook.py", line 103, in wrapper
    return func(*args, **kwargs)
  File "/home/jfrank/miniconda3/envs/hamilton/lib/python3.10/site-packages/ray/_private/worker.py", line 2659, in get
    values, debugger_breakpoint = worker.get_objects(object_refs, timeout=timeout)
  File "/home/jfrank/miniconda3/envs/hamilton/lib/python3.10/site-packages/ray/_private/worker.py", line 871, in get_objects
    raise value.as_instanceof_cause()
ray.exceptions.RayTaskError(TypeError): ray::build_result() (pid=661223, ip=10.21.249.30)
  At least one of the input arguments for this task could not be computed:
ray.exceptions.RayTaskError: ray::execute_lifecycle_for_node() (pid=661223, ip=10.21.249.30)
  File "/home/jfrank/code/hamilton/hamilton/execution/graph_functions.py", line 243, in execute_lifecycle_for_node
    result = node_(**kwargs)
  File "/home/jfrank/code/hamilton/hamilton/node.py", line 249, in __call__
    return self.callable(*args, **kwargs)
  File "/home/jfrank/code/hamilton/z_test_implementation.py", line 16, in at_1_to_previous
    return node_5s   + (time.time()-start)
TypeError: unsupported operand type(s) for +: 'ray._raylet.ObjectRef' and 'float'
(execute_lifecycle_for_node pid=661223) 
(execute_lifecycle_for_node pid=661223) ********************************************************************************
(execute_lifecycle_for_node pid=661223) > at_1_to_previous [__main__.at_1_to_previous()] encountered an error          <
(execute_lifecycle_for_node pid=661223) > Node inputs:
(execute_lifecycle_for_node pid=661223) {'node_5s': 'ObjectRef(c8ef45ccd0112571ffffffffffffffffffffffff...'}
(execute_lifecycle_for_node pid=661223) ********************************************************************************
(execute_lifecycle_for_node pid=661223) Traceback (most recent call last):
(execute_lifecycle_for_node pid=661223)   File "/home/jfrank/code/hamilton/hamilton/execution/graph_functions.py", line 243, in execute_lifecycle_for_node
(execute_lifecycle_for_node pid=661223)     result = node_(**kwargs)
(execute_lifecycle_for_node pid=661223)   File "/home/jfrank/code/hamilton/hamilton/node.py", line 249, in __call__
(execute_lifecycle_for_node pid=661223)     return self.callable(*args, **kwargs)
(execute_lifecycle_for_node pid=661223)   File "/home/jfrank/code/hamilton/z_test_implementation.py", line 16, in at_1_to_previous
(execute_lifecycle_for_node pid=661223)     return node_5s   + (time.time()-start)
(execute_lifecycle_for_node pid=661223) TypeError: unsupported operand type(s) for +: 'ray._raylet.ObjectRef' and 'float'
(execute_lifecycle_for_node pid=661226) 5s executed
```
Lastly, if I change the return to
```
def at_1_to_previous(node_5s: float) -> float:
    print("Value of 5s ", node_5s)
    print("1s executed")
    start = time.time()
    time.sleep(1)
    return node_5s   #+ (time.time()-start)
```
The UI again doesn't show anything and I get
```
/home/jfrank/miniconda3/envs/hamilton/bin/python /home/jfrank/code/hamilton/z_test_implementation.py
2024-08-24 21:59:25,593 INFO worker.py:1781 -- Started a local Ray instance.

Capturing execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/113


Captured execution run. Results can be found at http://localhost:8241/dashboard/project/1/runs/113

(execute_lifecycle_for_node pid=662808) Value of 5s  ObjectRef(c8ef45ccd0112571ffffffffffffffffffffffff0100000001000000)
(execute_lifecycle_for_node pid=662808) 1s executed
                                    at_1_to_previous
0  ObjectRef(c8ef45ccd0112571ffffffffffffffffffff...
(execute_lifecycle_for_node pid=662810) 5s executed
                                    at_1_to_previous
0  ObjectRef(c8ef45ccd0112571ffffffffffffffffffff...
```

hamilton/execution/graph_functions.py
```diff
@@ -201,69 +201,95 @@ def dfs_traverse(
...
+                            )
+                        except Exception:
+                            message = create_error_message(kwargs, node_, "[post-node-execute]")
+                            logger.exception(message)
+                            raise
+
+                        if remote_execute:
+                            for a in adapter._adapters:
+                                if hasattr(a, "stop"):
+                                    a.stop()
```
skrawcz, line 277: this can be dropped.

ui/sdk/src/hamilton_sdk/adapters.py
```diff
@@ -101,6 +101,10 @@ def __init__(
         # if you're using a float value.
         self.seed = None
 
+    def stop(self):
+        """Initiates stop if run in remote environment"""
+        self.client.stop()
+
```
skrawcz, line 107: this can be removed

ui/sdk/src/hamilton_sdk/api/clients.py
```diff
@@ -191,6 +192,9 @@ def __init__(
         ).start()
         self.worker_thread.start()
 
+        # Makes sure the process stops even if in remote environment
+        atexit.register(self.stop)
+
```
skrawcz, line 197: this registers and ensures stop is called at process exit -- that's why that other code can be removed.

z_test_implementation.py
skrawcz: so without diving in the details above this is my mental model for what should happen.

1. Hamilton is building a delayed execution graph with ray.remote functions. This means that as kwargs input to a function we will pass around ray object references.
2. this happens "driver" side -- i.e. functions that actually run on ray have the object references resolved properly by ray, and are none the wiser.
3. so "driver side" the result from the node will include a ray object reference.
4. the builder resolves all of these to pull the results locally to the driver side.

Does that make sense? I will look more closely tonight.

jernejfrank:
Awesome, I moved the post-execute-graph hooks to be executed after the results builder and it fixed the simple issue of displaying correct telemetry on the graph level. However, I don't have the bigger picture if this would mess things up for other adapters?

jernejfrank:
So the one thing missing is this weird behaviour:
If the second node is executed with time.sleep(i):
1) where i<5s (which is the first node), the UI doesn't shows only the first node and keeps thinking its running.
2) where i=5s, the UI shows first node complete and the second node as running without finishing
3) where i>>5s, the UI shows both nodes correct

jernejfrank:
Only thing I haven't done is written any tests beyond running the small script. Let me know if we should add these.

skrawcz:
@jernejfrank I think there's a few tests that need to be investigated. Some look like it's due to polars changing, others could be related to the changes here.

skrawcz:
@jernejfrank how much time do you have over the next week? I'd like to get this over the line.

The main thing is the unit tests -- else there's a few things to refactor around `raw_execute` that we should probably do. If you don't have time one of us can take that part on.

jernejfrank:
> @jernejfrank how much time do you have over the next week? I'd like to get this over the line.

Agreed, let's get this sorted. I can dedicated 2-3 hours in the evening/night next couple of days to finish what's missing.

> 
> The main thing is the unit tests -- else there's a few things to refactor around `raw_execute` that we should probably do. If you don't have time one of us can take that part on.

All unit tests pass for me now. I would suggest one more change --> changing ```raw_execute``` into ```__raw_execute``` to make it as private/protected as possible in Python because the ```post_graph_execute``` hooks are not part of it anymore and it will potentially break other people's stuff (even-though you urge to use ```execute```, this will enforce it more ;) )

skrawcz:
> I would suggest one more change --> changing raw_execute into __raw_execute to make it as private/protected as possible in Python because the post_graph_execute hooks are not part of it anymore and it will potentially break other people's stuff (even-though you urge to use execute, this will enforce it more ;) )

Correct, that's in line with the refactor we were thinking of. What we plan to do is that, but keep `raw_execute` there in case someone was using it, and add a deprecation warning on it -- I think we have a decorator somewhere for that...

hamilton/execution/graph_functions.py
```diff
@@ -285,6 +251,84 @@ def dfs_traverse(
     return computed
 
 
+# TODO: better function name
+def execute_lifecycle_for_node(
+    node_: node.Node,
+    adapter: LifecycleAdapterSet,
+    run_id: str,
+    task_id: str,
```
skrawcz, line 259: these will need to have `__` prefixes. Since we don't want to clash with possible values in`**kwargs` argument below; e.g. most people will not name variables with a `__` prefix so I think we should be good here.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dask-distributed = ["dask[distributed]"]
 datadog = ["ddtrace"]
 dev = [
   "pre-commit",
-  "ruff",
+  "ruff==0.5.7",
```
skrawcz, line 47:
```suggestion
  "ruff==0.5.7",  # this should match `.pre-commit-config.yaml`
```

ui/sdk/src/hamilton_sdk/adapters.py
```diff
@@ -103,6 +103,10 @@ def __init__(
...
```
skrawcz, line 109:
```suggestion
```
Not required

z_test_implementation.py
```diff
@@ -0,0 +1,64 @@
+import time
```
skrawcz, line 1: We should move this file to be an example under examples/ray/...

hamilton/execution/graph_functions.py
```diff
@@ -285,6 +251,84 @@ def dfs_traverse(
     return computed
 
 
+# TODO: better function name
+def execute_lifecycle_for_node(
```
skrawcz, line 255: We should probably have some dedicated unit tests? I know existing tests probably cover most of it. You could have a look at code coverage to see if any lines are missed?

skrawcz (changes requested):
Looking good! Only a few things to change - let me know if I misinterpreted anything.

hamilton/driver.py
```diff
@@ -657,24 +677,55 @@ def raw_execute(
...
         :param display_graph: DEPRECATED. DO NOT USE. Whether or not to display the graph when running it
         :param inputs: Runtime inputs to the DAG
         :return:
         """
-        function_graph = _fn_graph if _fn_graph is not None else self.graph
-        run_id = str(uuid.uuid4())
-        nodes, user_nodes = function_graph.get_upstream_nodes(final_vars, inputs, overrides)
+        self.function_graph = _fn_graph if _fn_graph is not None else self.graph
+        self.run_id = str(uuid.uuid4())
+        nodes, user_nodes = self.function_graph.get_upstream_nodes(final_vars, inputs, overrides)
```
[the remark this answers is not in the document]
jernejfrank: I don't like this structure as is.. the pre-execute hooks are in one function and the post-execute hooks are outside. The ```self``` was a workaround to get the test to pass and not running into issues with unspecified ```run_id``` when an exception occurs.

Maybe we can simplify this and just put the result_builder into __raw_execute? This would make everything neater, planning to code this up tonight.

dag_example_module.png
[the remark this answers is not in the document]
jernejfrank: I think it got created when I ran pytest on the whole test folder.

hamilton/driver.py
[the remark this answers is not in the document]
jernejfrank: Let me know if this solution is acceptable or if it breaks something somewhere else?

hamilton/execution/graph_functions.py
jernejfrank: My bad, I got confused with ```partial``` freezing them, but they can clash before that..

hamilton/driver.py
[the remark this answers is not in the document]
elijahbenizzy: So yeah, results builder is not in raw execute. The problem here is that it's not thread-safe -- this is something we rely on (E.G. being able to have multiple calls to the driver at once). So I don't think this will work?

IMO it's OK having it in two different places, but I think there are better solution here:
1. (simple, hackyish) Have `__raw_execute` return the run_id/function graph -- function graph can actually always be passed in.
2. Cleaner -- pass run_id into `__raw_execute` -- why should it generate? Then if we pass that + the graph in we can easily call it at the beginning. 
3. Probably best -- just move some of the logic outside of `__raw_execute`, always surrounding it by the pre/post graph hooks. Then it look like a sandwich, but has the right tooling. The `raw_execute` function is the only one that does not call the result builder, the rest call the post graph execute hook after calling the result builder.

One thing to note -- the earlier the pre graph execute and later the post graph execute hooks are called the better. This is because it can capture errors -- thus any configuration problems/hamilton bugs will be caught in the right place and displayed in the UI.

hamilton/execution/graph_functions.py
skrawcz: No worries - functools.partial will actually swallow any error it seems:
e.g.
```python
>>> def foo(a, b, c, **kwargs):
     print(a,b,c,kwargs)
     return "hi"
 
>>> f = functools.partial(foo, a=1, b=2, c=3)
>>> f(a=5)  # this will just override it -- won't error
5 2 3 {}
'hi'

>>> f(d=3)
1 2 3 {'d': 3}
'hi'

```

hamilton/driver.py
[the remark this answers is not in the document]
jernejfrank: I was 90% sure my suggestion will not work; looked too easy :) I adjusted the ```.materialize()``` entyr point in the same way since it also calls ```__raw_execute``` and all the test pass now as well.

hamilton/execution/graph_functions.py
jernejfrank: That is good to know, but also wth - somebody somewhere is probably having a big headache because of this...

z_test_implementation.py
jernejfrank: This file is weird for me and I don't know why, but when I put it under examples I get back the complaining from ray that it tries to add 1s to a ray reference, i.e. as if it were nested again. If I run it where it is, everything is fine and works as intended.

jernejfrank:
> Looking good! Only a few things to change - let me know if I misinterpreted anything.

Hi @skrawcz , let me know if there is anything I should add to finish the PR.

examples/ray/ray_Hamilton_UI_tracking/ray_lineage.py
```diff
@@ -27,6 +29,8 @@ def node_1s_error() -> float:
     username = "admin"
 
     try:
+        # ray.init()
+        ray.shutdown()
```
skrawcz, line 33: I assume this was just to test something?

examples/ray/ray_Hamilton_UI_tracking/ray_lineage.py
skrawcz, line 1: typically we have a README, requirements.txt, and a notebook version of the script in an example. Would you mind, please?

skrawcz (commented):
@jernejfrank what's the rationale for initializing ray in the adapter now?

My concern here now that I look more closely at it is any new behaviors should be opt-in, and it seems to have some specific behavior checks that I'm not sure we should be that opinionated on, e.g. always adding `ignore_reinit_error` kwarg.

So with that in mind:

1. I think initializing ray is fine, but you should opt-in to do it.
2. stopping Ray should again, be opt-in, or something you specify explicitly.

So my suggestion is:

1. allow someone to pass in ray config.
2. if that value is not None, do ray init with that config, else assume user has initialized ray.
3. have flag `shutdown_ray_on_completion=False` as a default, and use that to know when to shutdown or not.

Thoughts?

hamilton/driver.py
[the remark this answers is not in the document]
elijahbenizzy: Nice, thanks! Yeah we should have tests for this, but it's always an issue of how much you can reasonably cover...

skrawcz:
@jernejfrank I've tested locally -- so functionally things seem to work without regressions; I ran a few of the Hamilton examples to validate things.

So only outstanding thing is around the ray_config bit + tidying up the example.

skrawcz:
I also changed the brain to be main - so we'd do a squash merge here as a single commit... unless you want to tidy up the commits to be more atomic.

jernejfrank:
> 3\. shutdown_ray_on_completion

Fair point, we really shouldn't be adding a hidden init of a cluster. The initial idea was to abstract away from ray and create a context manager that would:

1) spin up cluster
2) execute DAG
3)  clean up after itself 


but I can see that it is too restrictive and changed it as suggested.

jernejfrank:
> I also changed the brain to be main - so we'd do a squash merge here as a single commit... unless you want to tidy up the commits to be more atomic.

happy with a squash commit!

hamilton/plugins/h_ray.py
```diff
@@ -135,6 +163,14 @@ def build_result(self, **outputs: typing.Dict[str, typing.Any]) -> typing.Any:
         result = ray.get(remote_combine)  # this materializes the object locally
         return result
 
+    def post_graph_execute(self, *args, **kwargs):
+        """We have the option to close the cluster down after execution."""
+
+        if not self.shutdown_ray_on_completion:
```
skrawcz, line 169: I think we want this right?
```suggestion
        if self.shutdown_ray_on_completion:
```

skrawcz (approved):
So just that one thing. But otherwise I think this LGTM. Will ask @elijahbenizzy to take a pass, but we'll merge this weekend if there isn't anything else! Thanks @jernejfrank !

elijahbenizzy:
Will play tomorrow with it then we can merge! I think this is the most surgery anyone who is not a core maintainer has done on the library -- good stuff! Special award to you 🥇

PR #1104, opened 2024-08-23
feat: First-class caching

zilto:
This follows-up the closed PR #1039 

Current TODOs:

- [x] Clean up docstrings and comments that are outdated
- [~] Write documentation
- [~] Add special behavior (always recompute, dont fingerprint, dont cache)
- [ ] Review the assumptions around the data model

Other  TODOs:

- [x] structured the Hamilton modules
- [x] standardized naming: `result_store` and `metadata_store` (closer to S3 lingo), `fingerprinting`, `fingerprint`
- [ ] introspection mechanism (pre-run, and post-run)
- [ ] introspection visualization
- [~] logging for cache adapter; could be improved for storage
- [x] checkpointing feature using `run_id`. Passing a `run_id` string or `latest` to `result_from` should do the trick
- [~ ]  OOP or protocols for `Store` subclasses. It's currently a bit confusing that `result_store` and `metadata_store` both inherit from the same `BaseStore` but they have different method signatures for `get` and `set`
- [~] Works with materializers. It does work, but has rough edges:
   - for `to.file` or other destinations where the suffix doesn't match the ideal file extension.
   - Can't pass kwargs to materializers
   - Assumes the materializer's only required arg is `path`


tests/caching/cases.py
```diff
@@ -0,0 +1,76 @@
...
+def module_6() -> ModuleType:
+    """Node with node and external dependency"""
+
+    def A() -> int:
+        return 1
+
+    def B(A: int, external: int) -> int:
+        return A + external + 3
+
+    return create_temporary_module(A, B)
```
skrawcz, line 66: these two are the same?

tests/caching/test_integration.py
```diff
@@ -0,0 +1,85 @@
...
+    repo = InMemoryMetadataStore(path=tmp_path)
+    cache = ShelveResultStore(path=tmp_path)
+    adapter = SmartCacheAdapter(repo=repo, result_store=cache)
+
+    return driver.Builder().with_modules(module).with_adapters(adapter).build()
+
+
+@pytest.mark.parametrize("h_driver", cases.ALL_MODULES, indirect=True)
+def test_single_run_is_successful(h_driver):
+    """The Driver with the adapter can be executed at least once
```
skrawcz, line 22:
```suggestion
    """The Driver with the adapter can be executed once
```

tests/caching/test_integration.py
```diff
@@ -0,0 +1,85 @@
...
+        if node.is_external_input:
+            inputs[node.name] = idx
+
+        final_vars.append(node.name)
+
+    h_driver.execute(final_vars, inputs=inputs)
+
+
+@pytest.mark.parametrize("h_driver", cases.ALL_MODULES, indirect=True)
+def test_run_twice_is_successful(h_driver):
```
skrawcz, line 43: we probably need some more integration tests:

1. Run graph once with input X, then re-run with Input Y. Validate what was computed.
2. Run graph once with input X, change some code. Re-run with input X. Validate what was computed.
3. Run graph once with input X, change some code. Re-run with input Y. Validate what was computed.

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+
+Control caching behavior
+-------------------------
+
+Hamilton's default caching behavior is ideal to get started and typically what you want during iterative development. However, you might want more control over the cache behavior in specific scenario or when moving to production. The caching behavior can be specified at the node-level via either:
+
+- **Dataflow definition** by adding to the function modifiers ``@tag(caching_behavior="ignore")``
+- **Driver definition** by passing a list of node name to ``.with_cache(ignore=["foo", "bar"])``
+
+The "Driver definition" approach will always override the behavior specified using ``@tag``. Passing empty lists to ``ignore=[]`` effectively disables all ``@tag``. Querying node names directly from the ``Driver`` is an effective way to select multiple nodes at once:
```
skrawcz, line 105: we probably want `None` here versus empty list.

skrawcz: or instead be explicit:
```python
.with_cache(ignore="*")
```

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+   * - ✅
+     - ✅
+     - ❌
+     - ❌
+
+Large and complex data objects (e.g., dataframes, machine learning models) can be expensive or unreliable to fingerprint, but may be valuable to cache nonetheless.
+
+In that case, using ``caching_behavior="dont_fingerprint"`` will store the result with a constant fingerprint making the result only retrievable via ``resume_from="latest"``.
+
+For instance, we don't want to store the ``weather_message`` because it's only a set of ``if/else`` that are trivial to compute.
```
skrawcz, line 264: where is weather message?

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+     - Store fingerprint
+   * - ✅
+     - ❌
+     - ✅
+     - ❌
+
+If a node produces results that are large in size, not picklable, or very inexpensive to compute, it might be preferrable to not store it. Also, your environment may have storage or I/O limitations (e.g., web service).
+
+
+Specifying ``caching_behavior="dont_store_result"`` means Hamilton won't add the result to the result store, but the fingerprint won't be added to the metadata store since it would point to nothing (i.e., "result store miss"). The fingerprint is only used to create the key for children nodes (e.g., ``recommended_outfit``).
```
skrawcz, line 312:
```suggestion
Specifying ``caching_behavior="dont_store_result"`` means Hamilton won't add the result to the result store, and the fingerprint won't be added to the metadata store since it would point to nothing (i.e., "result store miss"). The fingerprint is only used to create the key for downstream nodes (e.g., ``recommended_outfit``).
```

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+
+    This aims to match and replace the features of ``hamilton.experimental.h_cache.CachingGraphAdater`` but the API is preliminary.
+
+
+Resume from checkpoint
+-----------------------
+
+The caching feature stores results after each node execution. If execution fails half-way, the ``Driver`` can use the cache metadata and results to resume the run. You can think of it as automatically passing ``.execute(overrides={...})`` from disk.
+
+To resume, you can specify ``.with_cache(resume_from=...)`` which accepts a specific ``run_id`` or the string ``"latest"`` which will automatically get the results from the latest execution, whether it succeeded or failed. The ``run_id`` is automatically generated internally and can be viewed via the cache logger.
```
skrawcz, line 76: This ignores fingerprinting right? 
E.g.
1. run dataflow with `latest=True`. 
2. there is nothing in the cache -- so compute everything.
3. it fails due to a code error
4. we fix the code.
5. we restart with `latest=True`.
6. since there is data in the cache, for any node names that are the same, we will load the prior value, and continue computation from where it left off.

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+        .build()
+    )
+
+    dr.execute(...)
+
+.. note::
+
+    Using ``resume_from="latest"`` gets the latest run at the moment of building the ``Driver``. Consequently, calling ``Driver.execute()`` multiple times afterwards will always retrieve the same results from that specific point in time.
+
+
```
skrawcz, line 96: There needs to be a  "Only recompute what has changed" section to complement the above "checkpointing section".

E.g. 
2. run dataflow with caching.
3. there is nothing in the cache -- so compute everything.
4. it runs through to completion
5. we then run the code again but change an input
6. we only recompute the parts of the dataflow impacted by that new data
7. we then run the code again, but change code
8. we only recompute the parts of the dataflow impacted by that code change

elijahbenizzy:
High-level:

1. We should have some user-stories that match to common cases. THe docs are good for explaining the very high-level, but we want people to attach it to their workflow (maybe multiple notebooks?)
2. In the docs it feels like there's some assumed knowledge about terms (fingerprinting, etc...). I think we're exposing a bit at the lower level here -- user stories should help.

A few more user-stories can help me evaluate the use-cases. Seems technically largely solid, although I would want instructions on extending (E.G. adding a global cache)

docs/concepts/caching.rst
```diff
@@ -0,0 +1,489 @@
...
+.. list-table::
+   :header-rows: 1
+   :widths: auto
+   :align: center
+
+   * - Behavior
+     - Try to retrieve result
+     - Store result
+     - Fingerprint result
+     - Store fingerprint
```
skrawcz, line 366: When reading this table I don't know what I need to do.
This should be `@tag(...)` since that's what you want people to do, right?

docs/concepts/caching.rst
zilto: What you describe as "only recompute what has changed" is the default behavior described under "Basic". Unless you mean something else

docs/concepts/caching.rst
zilto: It doesn't ignore fingerprinting and I believe doesn't work as you expect. 

The `resume_from=` options means "resume from this specific `run_id`". If you have previous runs ("1", "2", "3", "4"), calling `resume_from="latest"` will pick the latest available `run_id` at Driver build time (i.e., `"4"`).

Node names are never used for caching. If code versions changed, this will be a cache miss. However, if your DAG has nodes
`foo -> bar -> baz`. 
1. You execute it
2. `baz` fails
3. you tweak baz
4. you rebuild the Driver with `resume_from="latest"`
5. execute again
6. execution will start at `bar` assuming its code version didn't change

docs/concepts/caching.rst
zilto: This is the summary table from other subheaders

skrawcz: I understand -- but I need to scroll up to make that connection.

docs/concepts/caching.rst
zilto: I believe what I meant here is the opposite of what you describe. The section refers to a dataflow like this

```python
@tag(cache_behavior="ignore")
def  foo() -> int: ...

def bar(foo: int) -> int: ...
```

What if the user wants to "disable" `cache="ignore"` from the `Driver` code? To me, having
```python
dr = driver.Builder().with_modules(...).with_cache(ignore=[ ])
```
Means "disregard all `cache_behavior="ignore"` statements". Since there's more than 2 behaviors, "disregard all `ignore`" isn't equivalent to "set all to `default`". 

```python
.with_cache(ignore="*")
#not equal to 
.with_cache(default="*")
```

hamilton/driver.py
```diff
@@ -1807,6 +1810,34 @@ def with_materializers(
         self.materializers.extend(materializers)
         return self
 
+    def with_cache(
+        self,
+        path: Optional[Union[str, pathlib.Path]] = None,
+        metadata_store: Optional[cache_store.BaseStore] = None,
+        result_store: Optional[cache_store.BaseStore] = None,
+        resume_from: Optional[Union[str, Literal["latest"]]] = None,
+    ) -> "Builder":
```
skrawcz, line 1819:
```suggestion
    def with_cache(
        self,
        path: Optional[Union[str, pathlib.Path]] = None,
        metadata_store: Optional[cache_store.BaseStore] = None,
        result_store: Optional[cache_store.BaseStore] = None,
        **kwargs = None,
    ) -> "Builder":
```

skrawcz: I think this API is fine -- kwargs can then handle strategy specific things, like `ignore`, etc.

hamilton/driver.py
```diff
@@ -1807,6 +1810,34 @@ def with_materializers(
...
+        """
+        metadata_store = (
+            metadata_store if metadata_store else cache_store.SQLiteMetadataStore(path=path)
+        )
+        result_store = result_store if result_store else cache_store.ShelveResultStore(path=path)
+
+        adapter = cache_adapters.SmartCacheAdapter(
+            metadata_store=metadata_store,
+            result_store=result_store,
+            resume_from=resume_from,
```
skrawcz, line 1836: to go with the above:
```suggestion
            **kwargs
```

hamilton/driver.py
```diff
@@ -1807,6 +1810,34 @@ def with_materializers(
...
+        metadata_store: Optional[cache_store.BaseStore] = None,
+        result_store: Optional[cache_store.BaseStore] = None,
+        resume_from: Optional[Union[str, Literal["latest"]]] = None,
+    ) -> "Builder":
+        """Add the caching adapter to the `Driver`
+
+        :param path: path where the cache metadata and results will be stored
+        :param metadata_store: BaseStore handling metadata for the cache adapter
+        :param result_store: BaseStore caching dataflow execution results
+        :param resume_from: Run id or "latest" to load metadata from at execution time.
```
skrawcz, line 1825:
```suggestion
        :param kwargs: See the specific cache strategy for kwargs that you can pass through.
```

hamilton/driver.py
skrawcz: Decisions:

1. with_cache is ONLY for fingerprinting
2. remove **kwargs - and push configurable kwargs for the fingerprinting strategy.

hamilton/driver.py
skrawcz, line 1836: we agreed to couple some kwargs from this in the `with_cache` API.

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+        If a fingerprint is retrieved, but reading from cache fails, invalidate the metadata store
+        and force recompute and storage of fingerprint + result.
+        """
+        # gather the attributes typically available through NodeExecutionMethod
+        node_name = node_.name
+        node_callable = node_.callable
+        node_kwargs = kwargs
+
+        if node_.node_role == hamilton.node.NodeType.EXPAND:
+            self._expand_nodes.add(node_name)
```
elijahbenizzy, line 963: Why can't we do this at the beginning? We know all the node types statically, pre-graph-execute could work.

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+            # Counter is the right data structure: it's unordered, arbitrary length, allow duplicates
+            data_version_counter = collections.Counter(collected_values)
+            collected_data_version = fingerprinting.hash_mapping(dict(data_version_counter))
+            # TODO apply the right caching behavior; usually applied in `run_after_node_execution`
+            self.set_memory_metadata(
+                node_name=collected_name, task_id=task_id, data_version=collected_data_version
+            )
+
+        # node "inside" a parallel block
+        elif task_id != node_name:
```
elijahbenizzy, line 978: OK, so I feel like we can do a better job with this type of stuff. Will likely be a TODO for later, but checking equality is a non-transparent way of figuring out what we want... E.G. we should have the task object and ask it if it's inside a grouped block.

Add a TODO with caveats then we can take a look at what will be involved in wiritg it through, should be trivial.

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+            result = self.execute_node(
+                node_name=node_name,
+                node_callable=node_callable,
+                node_kwargs=node_kwargs,
+                task_id=task_id,
+            )
+            return result
+
+        # create context_key before execution; will be reused after execution
+        context_key = self.create_context_key(
```
elijahbenizzy, line 998: A little unclear what/how a context key is used. Mind adding more explanation?

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+                node_name=node_name, task_id=task_id, context_key=context_key
+            )
+
+        # if data_version is found, try to retrieve stored result
+        if data_version is not None:
+            try:
+                result = self.get_stored_result(
+                    node_name=node_name, task_id=task_id, data_version=data_version
+                )
+            except ResultRetrievalError:
```
elijahbenizzy, line 1018: Comments about why we delete it -- a little confusing

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+        # create context_key before execution; will be reused after execution
+        context_key = self.create_context_key(
+            node_name=node_name, dep_names=node_kwargs.keys(), task_id=task_id
+        )
+        self.set_context_key(node_name=node_name, task_id=task_id, context_key=context_key)
+
+        # check in-memory metadata
+        data_version = self.get_memory_metadata(node_name=node_name, task_id=task_id)
+
+        # check stored metadata
```
elijahbenizzy, line 1006: Design question -- why are we separating stored from in-memory metadata? When do we sync them?

elijahbenizzy (commented):
Lots of thoughts here, exciting stuff

hamilton/driver.py
```diff
@@ -1807,6 +1828,39 @@ def with_materializers(
         self.materializers.extend(materializers)
         return self
 
+    def with_cache(
+        self,
+        path: Union[str, pathlib.Path] = "hamilton_cache",
```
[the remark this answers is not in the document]
zilto: Local subdirectory is a safer default IMO. The user *sees* the data in his filesystem and it becomes easier to delete files to clear disk space. Setting a cache per project seems like a reasonable pattern.

Once we have better cache eviction and storage management mechanisms, we could use a global cache `~/.hamilton`. We could use an environment variable to toggle between the two.

hamilton/driver.py
```diff
@@ -1807,6 +1828,39 @@ def with_materializers(
         self.materializers.extend(materializers)
         return self
 
+    def with_cache(
```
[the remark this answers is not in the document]
zilto: Added a check for it and a test.

hamilton/driver.py
```diff
@@ -1807,6 +1828,39 @@ def with_materializers(
...
+        self,
+        path: Union[str, pathlib.Path] = "hamilton_cache",
+        metadata_store: Optional[cache_store.BaseStore] = None,
+        result_store: Optional[cache_store.BaseStore] = None,
+        recompute: Optional[Union[Literal[True], Sequence[str]]] = None,
+        ignore: Optional[Union[Literal[True], Sequence[str]]] = None,
+        disable: Optional[Union[Literal[True], Sequence[str]]] = None,
+        log_to_file: bool = False,
+    ) -> "Builder":
+        """Add the caching adapter to the `Driver`
```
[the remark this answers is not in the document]
zilto: Added a simple example and a reference to the caching Concepts page.

hamilton/io/store.py
```diff
@@ -0,0 +1,361 @@
...
+    @abc.abstractmethod
+    def set(self, **future_kwargs) -> None:
+        """Add to the cache `key: value` or `key: materialization_instructions`"""
+
+    @abc.abstractmethod
+    def get(self, **future_kwargs) -> Any:
+        """Get the value from cache, potentially loading materialized data"""
+
+    @abc.abstractmethod
+    def delete(self, **future_kwargs) -> None:
```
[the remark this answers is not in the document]
zilto: My thinking was that individual implementations might have different kwargs / function signatures.

hamilton/io/fingerprinting.py
```diff
@@ -0,0 +1,117 @@
...
+import hashlib
+import logging
+from collections.abc import Mapping, Sequence, Set
+
+from hamilton.experimental import h_databackends
+
+logger = logging.getLogger(__name__)
+
+
+MAX_DEPTH = 4
```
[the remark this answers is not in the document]
zilto: There's no particular reason. I wanted sufficient depth to reduce hashing collision as default.

To hit depth 4, means you went through 4 times of non-primitives / supported types and tried to hash it's `__dict__` content.

hamilton/io/fingerprinting.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@hash_value.register(str)
+@hash_value.register(int)
+@hash_value.register(float)
+@hash_value.register(bool)
+@hash_value.register(bytes)
+def hash_primitive(obj, *args, **kwargs) -> str:
+    """Convert the primitive to a string and hash it"""
+    hash_object = hashlib.md5(str(obj).encode())
```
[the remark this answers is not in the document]
zilto: This conversion is need because `hash_primitive` is registered for `str`, `int`, `float`,  `bool`, which don't have an `.encode()` method.

I will create a separate method for `bytes`

hamilton/io/fingerprinting.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+@hash_value.register(Mapping)
+def hash_mapping(obj, *, sort: bool = False, **kwargs) -> str:
+    """Hash each key then its value.
+
+    The mapping is always sorted first because order shouldn't matter
+    in a mapping.
+
+    NOTE this may clash with Python dictionary ordering since >=3.7
```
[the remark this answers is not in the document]
zilto: I clarified the note. Here's the case:

```python
foo = {"key": 3, "key2": 13}
bar = {"key2": 13, "key": 3}
```

- Before Python 3.7, dictionaries don't store insertion order. To create a unique hash of a dictionary, it makes sense to sort keys and hash them. Then, `foo` and `bar` have the same hash.

- After Python 3.7 (all Hamilton support-versions), dictionaries store insertion order. In *some* cases, the users might expect `foo` to be unequal to `bar` because they have different insertion order. 

The note was to indicate the "unique hash" produced by `hash_mapping` isn't congruent with the definition of a dictionary since Python 3.7. Nonetheless, I think sorting keys and assuming order doesn't matter remains a better default.

Users can register a function for `collections.ordereddict` if they want to

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+    SET_RESULT = "set_result"
+    VERSION_DATA = "version_data"
+    VERSION_CODE = "version_code"
+    EXECUTE_NODE = "execute_node"
+    FAILED_EXECUTION = "failed_execution"
+    CREATE_CONTEXT_KEY = "create_context_key"
+    GET_CONTEXT_KEY = "get_context_key"
+    SET_CONTEXT_KEY = "set_context_key"
+    RESOLVE_BEHAVIOR = "resolve_behavior"
+    AS_OVERRIDE = "as_override"
```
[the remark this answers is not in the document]
zilto: `RESOLVE_BEHAVIOR` is intended to determine if a special behavior was specified via the `@cache` decorator or `.with_cache()`

`as_*` is meant to indicate that this node is an input or override. This one seems pretty self-explanatory. Maybe `IS_OVERRIDE` is better?

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+        """Set in-memory data_version whether a task_id is specified or not"""
+        assert data_version is not None
+        # default behavior; when not using task-based execution
+        if task_id is None:
+            self.data_versions[node_name] = data_version
+        # nodes outside are uniquely identifiable by node_name
+        elif node_name == task_id:
+            self.data_versions[node_name] = data_version
+        # nodes "on the edge" of parallelizable/collect blocks are uniquely identifiable by node_name
+        elif task_id == f"expand-{node_name}":
```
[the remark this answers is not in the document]
zilto: Definitely a pain to manage, but laying out all cases with `if/elif/else` makes for more readable and easier to maintain code. Hooks didn't provide more reliable information other than the `task_id`.

The cases are also pretty explicitly described in the comments. Typical cases:

- `task_id is None`: there's no `Parallelizable/Collect` used
- `task_id == node_name`: this is outside of a block 
- `task_id == f"expand-{node_name}`: this is a `Parallelizable` node
- `"collect-" in task_id`: this is a `Collect` node. The check is different than for `Parallelizable` because the task_id contains the name of the `Parallelizable` node, not the "current" one.
- else (i.e., `task_id != node_name`) this is inside a parallel/collect block

The best solution to avoid these would be to parse the `HamiltonGraph` in `run_before_graph_execution` and assign an enum "inside, outside, etc.", but this is redundant with the `node.NodeType` that's available in node-related hooks

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+    def get_context_key(self, node_name: str, task_id: Optional[str] = None) -> Optional[str]:
+        """Get in-memory context_key whether a task_id is specified or not"""
+        # default behavior; when not using task-based execution
+        if task_id is None:
+            context_key = self.context_keys.get(node_name, SENTINEL)
+        # nodes outside are uniquely identifiable by node_name
+        elif node_name == task_id:
+            context_key = self.context_keys.get(node_name, SENTINEL)
+        # nodes "on the edge" of parallelizable/collect blocks are uniquely identifiable by node_name
+        elif task_id == f"expand-{node_name}":
```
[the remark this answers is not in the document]
zilto: The point is not to get a unique id though. The goal is to retrieve the `node.NodeType` enum via the `task_id`

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+        **kwargs: Any,
+    ):
+        """Get code versions for all, and data version for top-level inputs and overrides
+        and setup state variables before the run
+        """
+        self.run_id = run_id
+        self._logs[run_id] = []
+        self.metadata_store.initialize()
+
+        self.graph = graph
```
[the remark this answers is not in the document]
zilto: What alternative do I have to storing things using self ? also, these attributes are set in run_before_graph_execution aren't they defacto associated with the latest execution?

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+
+        if node_.node_role == hamilton.node.NodeType.EXPAND:
+            self._expand_nodes.add(node_name)
+
+        elif node_.node_role == hamilton.node.NodeType.COLLECT:
+            self._collect_nodes.add(node_name)
+            collected_name = node_.collect_dependency
+            collected_values = node_kwargs[node_.collect_dependency]
+            # Counter is the right data structure: it's unordered, arbitrary length, allow duplicates
+            data_version_counter = collections.Counter(collected_values)
```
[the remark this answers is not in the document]
zilto: Your `Collect` has `[0, 0, 1, 1]`. The order is irrelevant. Although it's a `Collect`, I can't assume that they're all of the same time and sortable.

Therefore, I should hash each individual item and create a counter
`{hash(0): 2, hash(1): 2}`.

Then hashing this counter dictionary is a unique identifier of `[0, 0, 1, 1]` that doesn't rely on order or sorting

hamilton/lifecycle/caching.py
zilto: I tried looking at `HamiltonNode` but I'm pretty sure they don't communicate `Expand` / `Collect`. I could manually investigate the type again though

hamilton/lifecycle/caching.py
zilto: While this is a convoluted, the goal was to limit round-trips to the result store. The physical metadata store is always a super set of what's in memory. 

Invalidating the physical store is a way to force recompute. This should be more explicit as mentioned in another comment

hamilton/lifecycle/caching.py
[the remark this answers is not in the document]
elijahbenizzy: We should store state on run_id, so you'll want a dict of run_id -> graph object

hamilton/lifecycle/caching.py
elijahbenizzy: You can use the function graph directly: https://github.com/DAGWorks-Inc/hamilton/blob/72866175bb1e5dc65cedfd2de1096cc082fa19ee/hamilton/lifecycle/api.py#L300

hamilton/lifecycle/caching.py
elijahbenizzy: 👍

hamilton/io/fingerprinting.py
```diff
@@ -0,0 +1,117 @@
...
+    full Python object.
+
+    The default case hashes the `__dict__` attribute of the
+    object (recursive).
+    """
+    if hasattr(obj, "__dict__") and depth < MAX_DEPTH:
+        depth += 1
+        return hash_value(obj.__dict__, depth=depth)
+
+    hash_object = hashlib.md5("<unhashable>".encode())
```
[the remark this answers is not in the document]
zilto: From sync: 
- log a warning when hitting depth 2
- mention potential collision risks
- solutions: custom type or increase max depth

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1087 @@
...
+        """
+        # default behavior; when not using task-based execution
+        if task_id is None:
+            data_version = self.data_versions.get(node_name, SENTINEL)
+        # nodes outside are uniquely identifiable by node_name
+        elif node_name == task_id:
+            data_version = self.data_versions.get(node_name, SENTINEL)
+        # because `Parallelizable` node return the collected generator as a result
+        # this is not the data we want to version.
+        elif task_id == f"expand-{node_name}":
```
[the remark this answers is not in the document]
zilto: notes from sync:
- add detailed comments for `NodeRole`
- mention relationships to `parallel/collect`

hamilton/driver.py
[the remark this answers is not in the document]
elijahbenizzy: I guess? Seems very error-prone to me -- easy to accidentally invalidate by running it from the wrong place. It's generally bad (IMO) to rely on the CWD for program behavior, especially if it's a hard-to-debug side-effect. But if this is too much to change let's keep it like this for now.

hamilton/experimental/h_cache.py
```diff
@@ -185,6 +187,17 @@ def read_pickle_object(data: object, filepath: str) -> object:
         return pickle.load(file)
 
 
+@deprecation.deprecated(
```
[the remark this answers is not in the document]
elijahbenizzy: Does this work?

hamilton/driver.py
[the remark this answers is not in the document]
zilto: That's how MLFlow works by default, so it would feel familiar. It wouldn't be difficult to change the default location, but having a single central storage would require more utility for disk space management

docs/how-tos/caching.rst
```diff
@@ -0,0 +1,90 @@
+========
+Caching
+========
+
+In Hamilton, the term *caching* broadly refers to "reusing results from previous executions to skip redundant computation". When caching is enabled, node results are stored in a **result store** and where they can be retrieved by subsequent runs.
```
elijahbenizzy, line 5:
```suggestion
In Hamilton, the term *caching* broadly refers to "reusing results from previous executions to skip redundant computation". When caching is enabled, node results are stored in a **result store**, where they can be retrieved by subsequent runs.
```

docs/how-tos/caching.rst
```diff
@@ -0,0 +1,90 @@
...
+
+.. seealso::
+
+    For a broader overview of how caching works, see the :doc:`/concepts/caching` Concepts page .
+
+    For implementation details, see the :doc:`/reference/caching/caching-logic` Reference page .
+
+
+Use cases
+-----------
```
elijahbenizzy, line 15: Length of --- should match (everywhere)

docs/how-tos/caching.rst
```diff
@@ -0,0 +1,90 @@
...
+    from sklearn.linear import LinearRegression
+
+    def raw_data(file_path: str) -> pd.DataFrame:
+        return pd.read_parquet(file_path)
+
+    def cleaned_data(raw_data: pd.DataFrame) -> pd.DataFrame:
+        cleaned_df = ...  # fill nulls, filter rows, rename columns
+        return cleaned_df
+
+    def statistics(cleaned_data: pd.DataFrame) -> dict:
```
elijahbenizzy, line 37: Add comments to say what users will change. E.G. you want to keep the same raw data but mess around with cleaning...

docs/how-tos/caching.rst
```diff
@@ -0,0 +1,90 @@
...
+    For implementation details, see the :doc:`/reference/caching/caching-logic` Reference page .
+
+
+Use cases
+-----------
+
+Caching is beneficial for almost all use cases, and it's likely a good idea to have it on. The next sections present a illustrative scenarios.
+
+
+Iterative notebook development
```
elijahbenizzy, line 20: This is just one use-case. Are you planning to add more? Or should this just be "getting started"?

docs/how-tos/caching.rst
elijahbenizzy: But yeah, these don't capture all the use-cases we discussed/the more horizontal one (iterating on a node, changing the data, changing the parameters, etc...) I think just "getting started" is a good way, unless you're already planning to rewrite this.

docs/reference/caching/caching-logic.rst
```diff
@@ -0,0 +1,52 @@
+=======================
+Caching logic
```
elijahbenizzy, line 2: Brief overview on what this section is

docs/reference/caching/index.rst
```diff
@@ -0,0 +1,12 @@
...
+==============
+
+Reference
+---------
+
+.. toctree::
+    :maxdepth: 2
+
+    caching-logic
+    data-versioning
```
elijahbenizzy, line 12: I don't htink we need these to be nested

hamilton/io/fingerprinting.py
```diff
@@ -0,0 +1,178 @@
...
+import logging
+from collections.abc import Mapping, Sequence, Set
+from typing import Dict
+
+from hamilton.experimental import h_databackends
+
+logger = logging.getLogger(__name__)
+
+
+MAX_DEPTH = 2
```
elijahbenizzy, line 13: So the pattern is for people to change it? Might be worth having a method to do that. Or, add a TODO to make it easier to change. Generally bad to have people alter a constant in a module for config.

hamilton/lifecycle/caching.py
```diff
@@ -0,0 +1,1138 @@
...
+            CachingBehavior.from_string("recompute")
+
+        """
+        try:
+            return cls[string.upper()]
+        except KeyError as e:
+            raise KeyError(f"{string} is an invalid `CachingBehavior` value") from e
+
+
+class NodeRoleInTaskExecution(enum.Enum):
```
elijahbenizzy, line 144: Add comments that this is internal and we might end up changing this (I think we can align it with the way we do tasks better, just haven't scoped that out)

hamilton/plugins/h_diskcache.py
```diff
@@ -69,6 +70,19 @@ def evict_all_except_driver(dr: driver.Driver) -> dict:
 MAX_KWARGS_REPR_LENGTH = 200
 
 
+@deprecation.deprecated(
```
elijahbenizzy, line 73: Does this work on a class?

hamilton/stores/base.py
```diff
@@ -0,0 +1,246 @@
...
+
+class ResultRetrievalError(Exception):
+    pass
+
+
+class MetadataStoreIndexingError(Exception):
+    """Internal error that should be caught during testing."""
+
+
+# def search_data_adapter_registry(name: str, type_: type) -> Tuple[Type[DataSaver], Type[DataLoader]]:
```
elijahbenizzy: Remove comments?

hamilton/stores/base.py
```diff
@@ -0,0 +1,246 @@
...
+                for applicable_type in loader_cls.applicable_types()
+            )
+        )
+    except StopIteration as e:
+        raise KeyError(f"{name} doesn't have any DataLoader supporting type {type_}") from e
+
+    return saver_cls, loader_cls
+
+
+class ResultStore(abc.ABC):
```
elijahbenizzy, line 50: I think we should move this to caching, it's not top-level (as nobody else uses it). It's purely a tool for caching -- we can generalize later if we want.

hamilton/stores/file.py
```diff
@@ -0,0 +1,80 @@
+import shutil
```
elijahbenizzy, line 1: Calling this `file.py` will overwrite the default `file` by default. Something like `file_store` might be better.

examples/caching_nodes/caching.ipynb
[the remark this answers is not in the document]
zilto: I haven't updated this notebook over many commits. I will be updating it now that the visualization function is added.

We don't have the setup to embed notebooks in the Hamilton docs. Should this be added?

hamilton/function_modifiers/metadata.py
```diff
@@ -315,3 +315,61 @@ def ray_remote_options(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
...
+        if behavior:
+            cache_tags[Cache.BEHAVIOR_KEY] = behavior
+
+        if format:
+            cache_tags[Cache.FORMAT_KEY] = format
+
+        super(Cache, self).__init__(**cache_tags)
+
+
+def cache(
```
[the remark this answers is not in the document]
zilto: I just followed how the [recent ray decorator](https://github.com/DAGWorks-Inc/hamilton/blob/1365e8860fa98745bfe2da4cad1daa374711ef0b/hamilton/function_modifiers/metadata.py#L296) was done. Any suggestion on a model to follow?

hamilton/stores/file.py
zilto: AFAIK, there is no `file` in the standard Python library. As you pointed out, this module is internal so the following patterns seem safe 

```python
from hamilton.caching.stores.file import FileResultStore
```  
```python
# hamilton/caching/stores/__init__.py
from .file import FileResultStore

# elsewhere
from hamilton.caching.stores import FileResultStore
```

Let me know if I should prioritize one or another

hamilton/driver.py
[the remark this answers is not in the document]
zilto: Changing this string is sufficient to modify the default location in `.with_cache()`,  which would affect the default location of the `metadata_store` and the `result_store`. I can make the change tomorrow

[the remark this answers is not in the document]
elijahbenizzy: Yeah I'm happy with whatever you feel is the best user experience, let's just make it clear in the docs

hamilton/stores/file.py
elijahbenizzy: heh, you're right, I'm hallucinating. my bad

examples/caching_nodes/caching.ipynb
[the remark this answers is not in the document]
elijahbenizzy: Don't upscope, just link out + copy/paste code

hamilton/function_modifiers/metadata.py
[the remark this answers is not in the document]
elijahbenizzy: Ahh got it. So I think that's marginally different as it has specific logic in the class (E.G. not purely indirection for the sake of naming), but I'd argue that that should be lowercase as well -- very little in the class that couldn't be in a single one. No biggie either way I guess. Should have caught consistency -- you can see that most of the other decorators are just lowercase classnames. 

I think most of the rest of them follow the lowercase pattern.

skrawcz:
Looking good! Does this work with adapters like ray, or the graceful error one? 
If there are things that this doesn't work with or behavior would be weird -- we should have a section mentioning that.

hamilton/function_modifiers/metadata.py
```diff
@@ -315,3 +315,102 @@ def ray_remote_options(**kwargs: Union[int, Dict[str, int]]) -> RayRemote:
...
+
+            @cache(format="json", target_="performance")
+            @extract_fields(trained_model=LinearRegression, performance: dict)
+            def model_training() -> dict:
+                # ...
+                performance = {"rmse": 0.1, "mae": 0.2}
+                return {"trained_model": trained_model, "performance": performance}
+
+
+        :param behavior: The behavior of the cache. This can be one of the following:
```
elijahbenizzy, line 376: Note the fact that there is precedence here -- this is complicated but essential to getting it right

elijahbenizzy (commented):
Looking great! A few points on documentation. Caveats + testing:
- mlflow adapter 
- hamilton UI (curious what this looks like)

But yeah, gotta get this out, merge tonight/tomorrow morning?

hamilton/driver.py
```diff
@@ -1818,6 +1832,18 @@ def validate_materialization(
         all_nodes = nodes | user_nodes
         self.graph_executor.validate(list(all_nodes))
 
+    @property
+    def cache(self) -> SmartCacheAdapter:
+        """Directly access the cache adapter"""
+        if self.adapter:
+            for adapter in self.adapter.adapters:
+                if isinstance(adapter, SmartCacheAdapter):
```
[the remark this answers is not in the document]
zilto: the current name is dumb, but I wanted to avoid clashes with `hamilton.lifecycle.default.CacheAdapter`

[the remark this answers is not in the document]
elijahbenizzy: Sure, maybe `HamiltonCacheAdapter`? But yeah. Or just `Cache`

hamilton/caching/adapter.py
```diff
@@ -0,0 +1,1335 @@
...
+            raise ValueError(
+                "Can only set one of (`recompute`, `disable`, `ignore`) to True. Please pass mutually exclusive sets of node names"
+            )
+        self._recompute = recompute
+        self._disable = disable
+        self._ignore = ignore
+
+        # attributes populated at execution time
+        self.run_ids: List[str] = []
+        self._fn_graphs: Dict[str, FunctionGraph] = {}  # {run_id: graph}
```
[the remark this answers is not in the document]
zilto: I thoroughly reviewed the code last week and intentionally set the `_` prefix for internal and external attributes and methods. 

Is there a specific question or concern?

examples/caching/README.md
```diff
@@ -0,0 +1 @@
+# Caching
```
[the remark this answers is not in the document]
zilto: Should I delete the tutorials for adapters to be deprecated? I was hesitant and thought we should keep them until we actually remove those features.

I created `examples/caching` so that we can direct people to that link (which will display the `README`) and they won't see / peak at the older features to be deprecated.

docs/how-tos/caching.rst
zilto: Now covered in the tutorial notebook

docs/how-tos/caching.rst
zilto: Now covered in the tutorial notebook

hamilton/caching/adapter.py
```diff
@@ -0,0 +1,1335 @@
...
+    BaseDoNodeExecute,
+    BasePostNodeExecute,
+    BasePreGraphExecute,
+    BasePreNodeExecute,
+)
+
+logger = logging.getLogger("hamilton.caching")
+
+SENTINEL = object()
+S = TypeVar("S", object, object)
```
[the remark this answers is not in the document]
zilto: Having `-> Union[str, S]` improves readability by specifying that the function can return the `SENTINEL` value. It's much less ambiguous than not using a `TypeVar` such as  `-> Union[str, object]`.

It was your recommendation to use a `SENTINEL` value and it did help clarify the code and catch errors.

examples/caching/README.md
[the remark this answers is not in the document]
elijahbenizzy: Ahh, sure, as long as it's clear they're deprecated (big letters with a note/link to the right one) then it's fine to hold them around.

hamilton/caching/adapter.py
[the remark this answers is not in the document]
elijahbenizzy: IMO a typevar really shouldn't be used if it's not linked across two places. E.G. if this class were implemented using a typevar as a generic, then we have value from it. But I guess this is fine -- pylance isn't complaining (although it does if it's only used in one place).

Specifically, type-vars are best (IMO) when cascaded through different parts of the code. E.G. if you pass in an str, you get out an str. Hence what I mean by code completion. In this case it's really just a union type, and not a variable.

Better (IMO) would be `S=Union[str, object]`. But I think it's clearer to have `Union[str, object]` everywhere. 

Pylance doesn't like it because it can't actually parse that. So this is fine for now.

Funnily enough, `Union[str, object]` is kind of silly, as a `str` is an object.

Anyway, not a huge deal, but these are my thoughts.

hamilton/caching/adapter.py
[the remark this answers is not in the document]
elijahbenizzy: My point was that variables such as `behaviors` didn't feel public. Might be wrong though? And `run_ids`?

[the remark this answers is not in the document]
zilto: Those are meant for debugging, but I would consider that public still? A common pattern for all of those would be 

```python
# assuming multiple executions occured

run_id = dr.cache.run_ids[3]
cache_key = dr.cache.cache_keys[run_id]["trained_model"]
dr.cache.result_store.get(cache_key)
```

To the contrary, `._disable` isn't "resolved" yet (it might be a list of node names or `True`), so it's internal. For `_fn_graphs`, I felt that this mechanism was a bit of a hack and everything you need should be available through `.context_keys`, `.code_versions`, etc. The access to function graphs would probably change in 2.0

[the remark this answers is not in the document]
elijahbenizzy: So I think better practice is to have them be underscores then expose them with `@property`. This way you can add documentation, and have some semblance of a contract.

Otherwise, exposing instance vars is common, but not great practice. Can just have a TODO here, listing the public vars out if you don't want to make the change.

hamilton/caching/adapter.py
```diff
@@ -917,6 +952,41 @@ def resolve_behaviors(self, run_id: str) -> Dict[str, CachingBehavior]:
...
+        for node in graph.get_nodes():
+            if node.tags.get("hamilton.data_loader") is True:
+                main_node = node.tags["hamilton.data_loader.node"]
+                if main_node == node.name:
+                    continue
+
+                # solution for `@dataloader` and `from_`
+                if behaviors.get(main_node, None) is not None:
+                    behaviors[node.name] = behaviors[main_node]
+                # this hacky section is required to support @load_from and provide
```
elijahbenizzy, line 967: Hmm, not sure I follow what's happening here. Let's add some more docs.

elijahbenizzy: Specifically, document desired behavior here/how the code maps to it

hamilton/caching/adapter.py
```diff
@@ -1173,7 +1265,23 @@ def do_node_execute(
...
+                # nodes collected in `._data_loaders` return tuples of (result, metadata)
+                # where metadata often includes a timestamp. To ensure we provide a consistent
+                # `data_version` / hash, we only hash the result part of the materializer return
+                # value and discard the metadata.
+                if node_name in self._data_loaders[run_id] and isinstance(result, tuple):
+                    result = result[0]
+
                 data_version = self._version_data(node_name=node_name, run_id=run_id, result=result)
+
+                # nodes collected in `._data_savers` return a dictionary of metadata
```
elijahbenizzy, line 1278: I'm not convinced we need to special case these... But for now, let's call this a bit experimental (data savers/loaders) -- it's a very odd case that we don't want ot dwell onmore.

elijahbenizzy (commented):
Looking good! Will glance over when we sync tomorrow, then approve + merge.

elijahbenizzy (approved):
Well done!

PR #1105, opened 2024-08-26
[Async]HamiltonTracker support passing in custom CA cert

flavour:
Our Hamilton UI server is fronted by an SSL using an internal CA, which isn't supported by Hamilton SDK without this PR.

The standard synchronous `HamiltonTracker` can be worked around by using `os.environ["REQUESTS_CA_BUNDLE"]` but that's not ideal. The `AsyncHamiltonTracker` has no such workaround available & we are switching to  using that.

## Changes

Trackers take a new optional `verify` arg which mimics the `requests` arg of the same name.

## How I tested this

Both adapters tested on our internal network

## Checklist

### DIFF
--- a/ui/sdk/src/hamilton_sdk/adapters.py
+++ b/ui/sdk/src/hamilton_sdk/adapters.py
@@ -45,11 +45,12 @@ class HamiltonTracker(
         dag_name: str,
         tags: Dict[str, str] = None,
         client_factory: Callable[
-            [str, str, str], clients.HamiltonClient
+            [str, str, str, str | bool], clients.HamiltonClient
         ] = clients.BasicSynchronousHamiltonClient,
         api_key: str = None,
         hamilton_api_url=os.environ.get("HAMILTON_API_URL", constants.HAMILTON_API_URL),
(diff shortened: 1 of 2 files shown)


ui/sdk/src/hamilton_sdk/adapters.py
```diff
@@ -45,11 +45,12 @@ def __init__(
         dag_name: str,
         tags: Dict[str, str] = None,
         client_factory: Callable[
-            [str, str, str], clients.HamiltonClient
+            [str, str, str, str | bool], clients.HamiltonClient
```
skrawcz: this won't work for python < 3.10, right?

So we either import futures, or we use `Union[str, bool]`, right?

skrawcz: nvm looking at old version.

skrawcz (approved):
@elijahbenizzy any thoughts? I think this look good.

elijahbenizzy:
> @elijahbenizzy any thoughts? I think this look good.

Yep, looks good, maybe a few more comments on this (it's pretty clear, just the true/false/str is a bit confusing.

skrawcz:
@flavour plan is to release the SDK today 🤞 .

PR #1106, opened 2024-08-26
Fixes up the hello_world script

elijahbenizzy:
This is a remnant of the past -- this takes away all the ways to do things, and simplifies the explanation


examples/hello_world/my_script.py
```diff
@@ -7,37 +6,36 @@
...
-# module_name = "my_functions"
-# my_functions = importlib.import_module(module_name)
-# or import module(s) directly:
+# we use module pointers to do this
 import my_functions
 
-dr = driver.Driver(initial_columns, my_functions)  # can pass in multiple modules
-# we need to specify what we want in the final dataframe. These can be string names, or function references.
+# This uses the module above, but you can pass in as many as you want
+dr = driver.Driver({}, my_functions)
```
skrawcz, line 15: we should really change this to the builder pattern.

elijahbenizzy: Done, but it makes it slightly more complicated

PR #1115, opened 2024-08-31
Change `Parallelizable` and `Collect` to structural subtypes

### LINKED ISSUE #1114: ` Parallelizable` type checking error with `pyright` / VS Code
First off, I wanted to say "thanks", this project is a breath of fresh air and exactly what I was looking for. Please keep up the great work!

I was working along with [Dynamic DAGs/Parallel Execution](https://hamilton.dagworks.io/en/latest/concepts/parallel-task/) when I encountered a type checking error using `Parallelizable` as prescribed in the docs.

# Current behavior
Annotating a generator function with `Parallelizable` (see example below) results in the following `pyright` type error in either VS Code (using the `basic` type checking mode) and/or the command line:

```shell
error: Return type of generator function must be compatible with "Generator[str, Any, Any]"
    "Generator[str, Unknown, Unknown]" is not assignable to "Parallelizable[str]" (reportReturnType)
```

Full disclosure: I did *not* encounter any errors with `mypy`, even in strict mode.

## Screenshots
image

Note: The text of the error message is slightly different then the command line.

1. Run `pyright` (either via VS Code or on the command line) on the following code:
```python

def url() -> Parallelizable[str]:
    for url_ in ["web1", "web2", "web3"]:
        yield url_

def url_loaded(url: str) -> str:
    return url

def counts(url_loaded: str) -> int:
    return len(url_loaded.split(" "))

def total_words(counts: Collect[int]) -> int:
    return sum(counts)
```

Python versions: 3.10, 3.11, 3.12
Hamilton version: 1.75
OS: Windows 11

# Expected behavior
I expected the  `Parallelizable` type to be compatible with the generator function, but given that it is a nominal subtype of generator this is not possible.

# Additional context
After looking around the code to see how `Parallelizable` is consumed (namely in conjunction with `typing.get_origin`), wouldn't it be possible to switch to structural subtyping using `Protocol`? Something like the following?

```python
U = TypeVar("U", covariant=True)

class Parallelizable(Iterable[U], Protocol[U]): ...

def is_parallelizable_type(type_: Type) -> bool:
    return _get_origin(type_) == Parallelizable
```

I added these changes to a [fork](https://github.com/cswartzvi/hamilton/tree/parallelizable_typing) and ran the executor tests successfully. Note that I was having issues with other tests before I made the change - maybe because I am on Windows - so I cannot comment on the rest of the test suite. If you think this is a worthwhile addition to Hamilton I would be honored to submit a PR - let me know. Thanks!

cswartzvi:
Attempts to fix #1114 by introducing structural subtyping for `Parallelizable` and `Collect`.

## Changes

Both  `Parallelizable` and `Collect` were converted to generic protocols (vice abstract classes). The function `is_parallelizable_type` was also updated because generic protocols are not runtime checkable via `issubclass` - the function now checks the origin type via `typing.get_origin`.

## How I tested this
As mentioned in #1114, I was having issues running the full test suite locally on my Windows 11 machine, but I was able to run all the tests where `Parallelizable` is used, that includes the following folders:
- `./tests/execution/`
- `./tests/function_modifiers/`
- `./tests/resources/ `

## Notes
I didn't find any usage of `is_parallelizable_type` in the codebase. Is this intended for a future feature?

## Checklist

### DIFF
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -272,17 +272,27 @@ def get_type_information(some_type: Any) -> Tuple[Type[Type], list]:
 
 
 # Type variables for annotations below
-T = TypeVar("T")
-U = TypeVar("U", covariant=True)
-V = TypeVar("V", covariant=True)
+SequentialElement = TypeVar("SequentialElement", covariant=True)
+ParallelizableElement = TypeVar("ParallelizableElement", covariant=True)
+CollectElement = TypeVar("CollectElement", covariant=True)
 
 
 # TODO -- support sequential operation
-# class Sequential(Generator[T, None, None], ABC):
+# class Sequential(Iterable[SequentialElement], Protocol[SequentialElement]):
 #     pass
 
 
-class Parallelizable(Iterable[U], Protocol[U]):
+class Parallelizable(Iterable[ParallelizableElement], Protocol[ParallelizableElement]):
+    """Marks the output of a function node as parallelizable.
+
+    Parallelizable outputs are expected to be iterable, where each element dynamically
+    generates a node. When using dynamic execution, each of these dynamic nodes can be
+    executed in parallel.
+
+    Because this uses dynamic execution, the builder method `enable_dynamic_execution`
+    must be called with `allow_experimental_mode=True`.
+    """
+
     pass
 
 
@@ -290,8 +300,15 @@ def is_parallelizable_type(type_: Type) -> bool:
     return _get_origin(type_) == Parallelizable
 
 
-class Collect(Iterable[V], Protocol[V]):
-    pass
+class Collect(Iterable[CollectElement], Protocol[CollectElement]):
+    """Marks a function node parameter as collectable.
+
+    Collectable inputs are expected to be iterable, where each element is populated with
+    the results of dynamic nodes derived from parallelizable outputs.
+
+    Because this uses dynamic execution, the builder method `enable_dynamic_execution`
+    must be called with `allow_experimental_mode=True`.
+    """
 
 
 def check_input_type(node_type: Type, input_value: Any) -> bool:
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -1,8 +1,7 @@
 import inspect
 import sys
 import typing
-from abc import ABC
-from typing import Any, Generator, Optional, Tuple, Type, TypeVar, Union
+from typing import Any, Iterable, Optional, Protocol, Tuple, Type, TypeVar, Union
 
 import typing_inspect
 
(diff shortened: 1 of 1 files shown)


elijahbenizzy:
Hey! Thanks for this -- makes sense at a high-level. Also the issue is quite helpful in highlighting the problem. 

**TL;DR -- originally I intended it to be a generator but that's over-engineered. Looks good!**

So, two changes here:
(1) changing it from a `Generator` to an `Iterable`
(2) Changing it to be covariant types + using P

Covariant types make sense -- this is a pretty clear-cut case of where any subclass of V would work. Re: iterable -- generator was implemented by design (although not strictly necessary). Specifically, the future of this feature intends to allow more fine-grained pipelining for streaming results -- the send/yield from of a generator would be critical there. 

The problem is that python has no builtin protocol for `Generator` (it's a concrete type) which forces us to use ABC, which then goes up against the same problem as earlier. The "right" way would be to define a protocol, but ughh, that's a bit ridiculous. This is hitting the limitations of python's (quite lacking) type system. All this as far as I understand, I'm not an expert in python's typing system.

All that said, we haven't actually made use of the fact that it's a generator, and we could always have another type (E.G. `Stream`). So talking myself into a circle, I think it's fine to loosen it to `Iterable`, then we can have a special case when we ant for a generator.

Mind adding a few comments on the type? Also while you're working on this, maybe we should change the name of `U` and `V` to `ParallelizableElement` and `CollectElement`? Or better names.

Appreciate this, and the chance to dig into python's type system!

cswartzvi:
Ah! So you were thinking about streaming via `send` - I thought that might be the case when I saw `Generator` as the base class. For what it's worth, I think that adding an explicit `Stream` type down the road, separate from `Parallelizable`, fits nicely with the overall library. 

I feel your frustrations with python's type system - I have been using it pretty heavily over the last 3-4 years (I work with a lot of C# people and that is the only way to get them involved with python) but I still bump into rough paths all the time - like no builtin protocol for Generator, which I have seen before, but definitely did not remember.

Anyway, let me add those comments and update the type variable names - I decided to go with your suggestions for the names. Note that I also updated the commented out `Sequential` so that it will start in a consistent place if/when it is introduced.

elijahbenizzy:
> Ah! So you were thinking about streaming via `send` - I thought that might be the case when I saw `Generator` as the base class. For what it's worth, I think that adding an explicit `Stream` type down the road, separate from `Parallelizable`, fits nicely with the overall library.
> 
> I feel your frustrations with python's type system - I have been using it pretty heavily over the last 3-4 years (I work with a lot of C# people and that is the only way to get them involved with python) but I still bump into rough paths all the time - like no builtin protocol for Generator, which I have seen before, but definitely did not remember.
> 
> Anyway, let me add those comments and update the type variable names - I decided to go with your suggestions for the names. Note that I also updated the commented out `Sequential` so that it will start in a consistent place if/when it is introduced.

Yep, I agree on the way to proceed. I'm just that weird one hoping for true compiled python at some point 😆 

Appreciate the comments + changes + in-depth thoughts! This looks good, merging!

elijahbenizzy (approved):
Thank you!

PR #1116, opened 2024-09-02
Handles partials in source code hash

skrawcz:
This is a stop gap measure to handle
partials for the CacheAdapter.

## Changes
- adds check for partial functions; they have `.func` attribute on a callable.

## How I tested this
This code now works:

```python
from hamilton.function_modifiers import source, inject
from hamilton.ad_hoc_utils import create_temporary_module
import time

@inject(foo=source("foo"))
def worker(seconds: int, foo: int) -> int:
    print(f"start work {seconds}s")
    start = time.time()
    time.sleep(1)
    # yield {
    #     "job_name": f"sleep_{seconds}",
    #     "cmd": [f"sleep {str(seconds)}"],
    #     "block": True,
    # }
    print(f"foo: {foo}")
    end = time.time()
    print(f"end stop {end - start:.2f}s work")
    return seconds


if __name__ == "__main__":
    from hamilton import driver
    from hamilton.execution import executors
    from hamilton.lifecycle.default import CacheAdapter
    logic = create_temporary_module(
        worker,
    )

    dr = (
        driver.Builder()
        .with_modules(logic)
        .with_local_executor(executors.SynchronousLocalTaskExecutor())
        .with_remote_executor(
            executors.MultiThreadingExecutor(max_tasks=4)
        )
        # !!! error when uncomment this line !!!
        # TypeError: module, class, method, function, traceback, frame, or code object was expected, got partial
        .with_adapters(CacheAdapter(cache_path="cache"))
        .with_config({"foo": 42})
        .build()
    )
    start = time.time()
    dr.execute(["worker"], inputs={"seconds": 3})
    print(f"Time taken: {time.time() - start: .2f} seconds")
```

## Checklist

### DIFF
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -376,7 +376,10 @@ class CacheAdapter(NodeExecutionHook, NodeExecutionMethod, GraphExecutionHook):
         if node_name not in self.cache_vars:
             return node_callable(**node_kwargs)
 
-        node_hash = graph_types.hash_source_code(node_callable, strip=True)
+        source_of_node_callable = node_callable
+        while hasattr(source_of_node_callable, "func"):  # handle partials
+            source_of_node_callable = source_of_node_callable.func
+        node_hash = graph_types.hash_source_code(source_of_node_callable, strip=True)
         cache_key = CacheAdapter.create_key(node_hash, node_kwargs)
 
         from_cache = self.cache.get(cache_key, None)
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -8,6 +8,7 @@ import pprint
 import random
 import shelve
 import time
+from functools import partial
 from typing import Any, Callable, Dict, List, Optional, Type, Union
 
 from hamilton import graph_types, htypes
@@ -377,7 +378,7 @@ class CacheAdapter(NodeExecutionHook, NodeExecutionMethod, GraphExecutionHook):
             return node_callable(**node_kwargs)
 
         source_of_node_callable = node_callable
-        while hasattr(source_of_node_callable, "func"):  # handle partials
+        while isinstance(source_of_node_callable, partial):  # handle partials
             source_of_node_callable = source_of_node_callable.func
         node_hash = graph_types.hash_source_code(source_of_node_callable, strip=True)
         cache_key = CacheAdapter.create_key(node_hash, node_kwargs)
--- a/hamilton/lifecycle/default.py
+++ b/hamilton/lifecycle/default.py
@@ -359,7 +359,7 @@ class CacheAdapter(NodeExecutionHook, NodeExecutionMethod, GraphExecutionHook):
     def run_before_graph_execution(self, *, graph: HamiltonGraph, **kwargs):
         """Set `cache_vars` to all nodes if received None during `__init__`"""
         self.cache = shelve.open(self.cache_path)
-        if self.cache_vars == []:
+        if len(self.cache_vars) == 0:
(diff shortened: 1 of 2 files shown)


hamilton/lifecycle/default.py
```diff
@@ -376,7 +376,10 @@ def run_to_execute_node(
         if node_name not in self.cache_vars:
             return node_callable(**node_kwargs)
 
-        node_hash = graph_types.hash_source_code(node_callable, strip=True)
+        source_of_node_callable = node_callable
+        while hasattr(source_of_node_callable, "func"):  # handle partials
```
Roy-Kid, line 380: I this here maybe better to use `isinstance(source_of_node_callable, partial)`, since `func` is a too common name?

skrawcz: sure good idea

hamilton/lifecycle/default.py
```diff
@@ -376,7 +377,10 @@ def run_to_execute_node(
         if node_name not in self.cache_vars:
             return node_callable(**node_kwargs)
 
-        node_hash = graph_types.hash_source_code(node_callable, strip=True)
+        source_of_node_callable = node_callable
+        while isinstance(source_of_node_callable, partial):  # handle partials
```
Roy-Kid, line 381: clever 😂

Roy-Kid (approved):
It looks nice. We can merge it.

hamilton/lifecycle/default.py
elijahbenizzy, line 381: cool -- can make it into a function for clarity: `get_original_function(func: Callable)` + test, but this is good

tests/lifecycle/test_cache_adapter.py
```diff
@@ -138,3 +165,49 @@ def test_commit_nodes_history(hook: CacheAdapter):
     # need to reopen the hook cache
     with shelve.open(hook.cache_path) as cache:
         assert cache.get(CacheAdapter.nodes_history_key) == hook.nodes_history
+
+
+def test_partial_handling(hook: CacheAdapter, node_a_partial: node.Node):
```
elijahbenizzy, line 170: Yeah if the func-derivation is a function then this one is less necessary (although still nice to have), and we can test that/get more confidence.

elijahbenizzy (approved):
Nits, looks good

hamilton/lifecycle/default.py
```diff
@@ -376,7 +377,10 @@ def run_to_execute_node(
...
+            source_of_node_callable = source_of_node_callable.func
+        node_hash = graph_types.hash_source_code(source_of_node_callable, strip=True)
```
zilto, line 383: instead of this code, you should use `node.version` (`HamiltonNode.version`). It uses the same code implementation, but the `CacheAdapter` predates the `.version`

skrawcz: that's not wired through.

zilto: Then, it should be retrieved from `graph: HamiltonGraph` in `run_before_graph_execution`

elijahbenizzy: We shoudln't just wire it through for the sake of  using the same function. But yeah, that brings up a point of why are we computing this here when we compute statically? We should just compute all the hash keys in advance.

Let's just

skrawcz: I think this is fine what I have now -- it's stop gap measure until new caching stuff comes out anyway.

elijahbenizzy: Yep, approved, ship it

PR #1121, opened 2024-09-04
Pydantic validator

### LINKED ISSUE #473: Pydantic datatype validation for hamilton nodes
We have pandera, we should have pydantic as well.

```python
class MyModel(BaseModel):
    ...
# mode 1
@check_output(schema=MyModel)
def foo() -> dict:
    ...

# mode 2
@h_pydantic.check_output_schema()
def foo() -> MyModel:
    ...
```

I, personally, find pydantic to be overkill for most cases and think the world should have united around dataclasses/typed dicts, separate from validation. That said, everyone uses it, so 🤷 .

**Additional context**
Related: https://github.com/DAGWorks-Inc/hamilton/issues/469

cswartzvi:
This PR adds pydantic integration via a new data validator and plugin. Resolves #473.

## Changes

I added a file called `data_quality/pydantic_validators.py` with a new default validator `PydanticModelValidator`. This validator is dynamically added to the list of available default validators if pydantic is available - very similar, some would say identical 😄, to how the pandera validators are added. The pydantic validator is passed a `model` parameter that is then used to validated the output of the decorated function:

```python
class MyModel(BaseModel):
    name: str

@check_output(model=MyModel)
def foo() -> dict:
    return {"name": "hamilton"}
```

I also added a plugin file `plugins/h_pydantic.py` with a variant of the `check_output` decorator that uses the return type annotation to establish the pydantic model to validate:

```python
class MyModel(BaseModel):
    name: str

@h_pydantic.check_output()
def foo() -> MyModel:
    return MyModel(name="hamilton")
```

My implementation uses the pydantic [`TypeAdapter`](https://docs.pydantic.dev/latest/concepts/type_adapter/) - mainly because [`pydantic.validate_call`](https://docs.pydantic.dev/latest/api/validate_call/) does not have an option to *only* check the return value. `TypeAdapter` allows you to specify a `strict` mode where type coercion is turned off, since modifying the outputs is not currently allowed (that is correct, right?). I have enabled `strict` mode for both the validator and the plugin.

Although I think that it is idiomatic to pydantic, I should point out that `strict`mode  does *not* stop you from doing this (i.e. this passes validation):

```python
class MyModel(BaseModel):
    name: str

@h_pydantic.check_output()
def foo() -> MyModel:
    return {name: "hamilton"}
```

One last thing to mention is that `h_pydantic.check_output` currently checks that the return type annotation is a subclass of `pydantic.BaseModel`. In theory, you could use pydantic to check all kinds of things (builtins, `Annotated` types, ...), however I was having trouble getting it to play nicely with validator resolution so I scraped it.

## How I tested this

I added a file to the testing suite `test_pydantic_data_quality.py` that tests the validator and the `check_out` plugin decorator for both basic and complex cases.

## Notes

There are a couple of points I wanted to bring up:
-  `TypeAdapter` is a pydantic 2.0 feature and in conflict with the `[vaex]` extra dependency 😞. I didn't notice this until I was updating `pyproject.toml`, let me know if this is a deal breaker and I will come up with an alternative implementation
- I deviated from the spec in #473 and used `model` (vice `schema`) and `pydantic.check_output` (vice `pydantic.check_output_schema`) - I can change them back if desired, I just thought they fit better with the terminology of pydantic and the ergonomics of the pandera plugin, respectively.
- I will update the documentation and the plugin docstring if you are good with the above notes

## Checklist

Thanks for the opportunity to dig into this!

### DIFF
--- a/hamilton/data_quality/default_validators.py
+++ b/hamilton/data_quality/default_validators.py
@@ -511,9 +511,9 @@ _append_pandera_to_default_validators()
 def _append_pydantic_to_default_validators():
     """Utility method to append pydantic validators as needed"""
     try:
-        from pydantic import BaseModel  # noqa: F401
+        import pydantic  # noqa: F401
     except ModuleNotFoundError:
-        logger.info(
+        logger.debug(
             "Cannot import pydantic from pydantic_validators. Run pip install sf-hamilton[pydantic] if needed."
         )
         return
--- a/hamilton/data_quality/default_validators.py
+++ b/hamilton/data_quality/default_validators.py
@@ -508,6 +508,23 @@ def _append_pandera_to_default_validators():
 _append_pandera_to_default_validators()
 
 
+def _append_pydantic_to_default_validators():
+    """Utility method to append pydantic validators as needed"""
+    try:
+        from pydantic import BaseModel  # noqa: F401
+    except ModuleNotFoundError:
+        logger.info(
+            "Cannot import pydantic from pydantic_validators. Run pip install sf-hamilton[pydantic] if needed."
+        )
+        return
+    from hamilton.data_quality import pydantic_validators
+
+    AVAILABLE_DEFAULT_VALIDATORS.extend(pydantic_validators.PYDANTIC_VALIDATORS)
+
+
+_append_pydantic_to_default_validators()
+
+
 def resolve_default_validators(
     output_type: Type[Type],
     importance: str,
--- a/hamilton/data_quality/pydantic_validators.py
+++ b/hamilton/data_quality/pydantic_validators.py
@@ -1,0 +1,61 @@
+from typing import Any, Type
+
+from pydantic import BaseModel, TypeAdapter, ValidationError
+
+from hamilton.data_quality import base
+from hamilton.htypes import custom_subclass_check
+
+
+class PydanticModelValidator(base.BaseDefaultValidator):
+    """Pydantic model compatibility validator
+
+    Note that this validator uses pydantic's strict mode, which does not allow for
+    coercion of data. This means that if an object does not exactly match the reference
+    type, it will fail validation, regardless of whether it could be coerced into the
+    correct type.
+
+    :param model: Pydantic model to validate against
+    :param importance: Importance of the validator, possible values "warn" and "fail"
+    :param arbitrary_types_allowed: Whether arbitrary types are allowed in the model
+    """
+
+    def __init__(self, model: type[BaseModel], importance: str):
+        super(PydanticModelValidator, self).__init__(importance)
+        self.model = model
+
+        self._model_adapter = TypeAdapter(model)
+
+    @classmethod
+    def applies_to(cls, datatype: Type[Type]) -> bool:
+        # In addition to checking for a subclass of BaseModel, we also check for dict
+        # as this is the standard 'de-serialized' format of pydantic models in python
+        return custom_subclass_check(datatype, BaseModel) or custom_subclass_check(datatype, dict)
+
+    def description(self) -> str:
+        return "Validates that the returned object is compatible with a pydantic model"
+
+    def validate(self, data: Any) -> base.ValidationResult:
+        try:
+            # Currently, validate can not alter the output data, so we must use
+            # strict=True. The downside to this is that data that could be coerced
+            # into the correct type will fail validation.
+            self._model_adapter.validate_python(data, strict=True)
+        except ValidationError as e:
+            return base.ValidationResult(
+                passes=False, message=str(e), diagnostics={"model_errors": e.errors()}
+            )
+        return base.ValidationResult(
+            passes=True,
+            message=f"Data passes pydantic check for model {str(self.model)}",
+        )
+
+    @classmethod
+    def arg(cls) -> str:
+        return "model"
+
+    @classmethod
+    def name(cls) -> str:
+        return "pydantic_validator"
+
+
+PYDANTIC_VALIDATORS = [PydanticModelValidator]
--- a/hamilton/data_quality/pydantic_validators.py
+++ b/hamilton/data_quality/pydantic_validators.py
@@ -19,10 +19,9 @@ class PydanticModelValidator(base.BaseDefaultValidator):
     :param arbitrary_types_allowed: Whether arbitrary types are allowed in the model
     """
 
-    def __init__(self, model: type[BaseModel], importance: str):
+    def __init__(self, model: Type[BaseModel], importance: str):
         super(PydanticModelValidator, self).__init__(importance)
         self.model = model
-
         self._model_adapter = TypeAdapter(model)
 
     @classmethod
@@ -32,7 +31,7 @@ class PydanticModelValidator(base.BaseDefaultValidator):
         return custom_subclass_check(datatype, BaseModel) or custom_subclass_check(datatype, dict)
 
     def description(self) -> str:
-        return "Validates that the returned object is compatible with a pydantic model"
+        return "Validates that the returned object is compatible with the specified pydantic model"
 
     def validate(self, data: Any) -> base.ValidationResult:
         try:
--- a/hamilton/plugins/h_pydantic.py
+++ b/hamilton/plugins/h_pydantic.py
@@ -28,6 +28,7 @@ class check_output(BaseDataValidationDecorator):
         Here is an example of how to use this decorator with a function that returns a pydantic model:
 
         .. code-block:: python
+
             :name: "@check_output using pydantic model output"
 
             from pydantic import BaseModel
@@ -46,6 +47,7 @@ class check_output(BaseDataValidationDecorator):
         complain about this):
 
         .. code-block:: python
+
             :name: "@check_output using dict output"
 
             from pydantic import BaseModel
@@ -65,6 +67,7 @@ class check_output(BaseDataValidationDecorator):
         following function will *fail* validation:
 
         .. code-block:: python
+
             :name: "@check_output invalid output in strict mode"
 
             from pydantic import BaseModel
--- a/hamilton/plugins/h_pydantic.py
+++ b/hamilton/plugins/h_pydantic.py
@@ -29,8 +29,6 @@ class check_output(BaseDataValidationDecorator):
 
         .. code-block:: python
 
-            :name: "@check_output using pydantic model output"
-
             from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
 
@@ -47,8 +45,6 @@ class check_output(BaseDataValidationDecorator):
         complain about this):
 
         .. code-block:: python
-
-            :name: "@check_output using dict output"
 
             from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
@@ -67,8 +63,6 @@ class check_output(BaseDataValidationDecorator):
         following function will *fail* validation:
 
         .. code-block:: python
-
-            :name: "@check_output invalid output in strict mode"
 
             from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
--- a/hamilton/plugins/h_pydantic.py
+++ b/hamilton/plugins/h_pydantic.py
@@ -1,0 +1,32 @@
+from typing import List
+
+from pydantic import BaseModel
+
+from hamilton import node
+from hamilton.data_quality import base as dq_base
+from hamilton.function_modifiers import InvalidDecoratorException
+from hamilton.function_modifiers import base as fm_base
+from hamilton.function_modifiers import check_output as base_check_output
+from hamilton.function_modifiers.validation import BaseDataValidationDecorator
+from hamilton.htypes import custom_subclass_check
+
+
+class check_output(BaseDataValidationDecorator):
+    def __init__(
+        self,
+        importance: str = dq_base.DataValidationLevel.WARN.value,
+        target: fm_base.TargetType = None,
+    ):
+        super(check_output, self).__init__(target)
+        self.importance = importance
+        self.target = target
+
+    def get_validators(self, node_to_validate: node.Node) -> List[dq_base.DataValidator]:
+        output_type = node_to_validate.type
+        if not custom_subclass_check(output_type, BaseModel):
+            raise InvalidDecoratorException(
+                f"Output of function {node_to_validate.name} must be a Pydantic model"
+            )
+        return base_check_output(
+            importance=self.importance, model=output_type, target_=self.target
+        ).get_validators(node_to_validate)
--- a/hamilton/plugins/h_pydantic.py
+++ b/hamilton/plugins/h_pydantic.py
@@ -29,8 +29,8 @@ class check_output(BaseDataValidationDecorator):
 
         .. code-block:: python
 
-            from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
+            from pydantic import BaseModel
 
             class MyModel(BaseModel):
                 a: int
@@ -46,8 +46,8 @@ class check_output(BaseDataValidationDecorator):
 
         .. code-block:: python
 
-            from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
+            from pydantic import BaseModel
 
             class MyModel(BaseModel):
                 a: int
@@ -58,14 +58,33 @@ class check_output(BaseDataValidationDecorator):
             def foo() -> MyModel:
                 return {"a": 1, "b": 2.0, "c": "hello"}
 
+        You can also use pydantic validation through ``function_modifiers.check_output`` by
+        providing the model as an argument:
+
+        .. code-block:: python
+
+            from typing import Any
+
+            from hamilton import function_modifiers
+            from pydantic import BaseModel
+
+            class MyModel(BaseModel):
+                a: int
+                b: float
+                c: str
+
+            @function_modifiers.check_output(model=MyModel)
+            def foo() -> dict[str, Any]:
+                return {"a": 1, "b": 2.0, "c": "hello"}
+
         Note, that because we do not (yet) support modification of the output, the validation is
         performed in strict mode, meaning that no data coercion is performed. For example, the
         following function will *fail* validation:
 
         .. code-block:: python
 
-            from pydantic import BaseModel
             from hamilton.plugins import h_pydantic
+            from pydantic import BaseModel
 
             class MyModel(BaseModel):
                 a: int  # Defined as an int
--- a/hamilton/plugins/h_pydantic.py
+++ b/hamilton/plugins/h_pydantic.py
@@ -17,6 +17,69 @@ class check_output(BaseDataValidationDecorator):
         importance: str = dq_base.DataValidationLevel.WARN.value,
         target: fm_base.TargetType = None,
     ):
+        """Specific output-checker for pydantic models. This decorator utilizes the output type of
+        the function, which can be any subclass of pydantic.BaseModel. The function output must
+        be declared with a type hint.
+
+        :param model: The pydantic model to use for validation. If this is not provided, then the output type of the function is used.
+        :param importance: Importance level (either "warn" or "fail") -- see documentation for check_output for more details.
+        :param target: The target of the decorator -- see documentation for check_output for more details.
+
+        Here is an example of how to use this decorator with a function that returns a pydantic model:
+
+        .. code-block:: python
+            :name: "@check_output using pydantic model output"
+
+            from pydantic import BaseModel
+            from hamilton.plugins import h_pydantic
+
+            class MyModel(BaseModel):
+                a: int
+                b: float
+                c: str
+
+            @h_pydantic.check_output()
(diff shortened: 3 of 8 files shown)


skrawcz:
@cswartzvi would you mind creating an example under `data_quality` showing it in action?

Otherwise just need to make sure the decorator turns up in our documentation.

hamilton/data_quality/pydantic_validators.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def description(self) -> str:
+        return "Validates that the returned object is compatible with a pydantic model"
+
+    def validate(self, data: Any) -> base.ValidationResult:
+        try:
+            # Currently, validate can not alter the output data, so we must use
+            # strict=True. The downside to this is that data that could be coerced
+            # into the correct type will fail validation.
+            self._model_adapter.validate_python(data, strict=True)
```
elijahbenizzy, line 42: Just checking -- `data` can be `dict` or `pydantic.BaseModel` here, right?

hamilton/data_quality/pydantic_validators.py
```diff
@@ -0,0 +1,61 @@
...
+        self._model_adapter = TypeAdapter(model)
+
+    @classmethod
+    def applies_to(cls, datatype: Type[Type]) -> bool:
+        # In addition to checking for a subclass of BaseModel, we also check for dict
+        # as this is the standard 'de-serialized' format of pydantic models in python
+        return custom_subclass_check(datatype, BaseModel) or custom_subclass_check(datatype, dict)
+
+    def description(self) -> str:
+        return "Validates that the returned object is compatible with a pydantic model"
```
elijahbenizzy, line 35: Nit -- with the "specified" pydantic model

hamilton/data_quality/pydantic_validators.py
```diff
@@ -0,0 +1,61 @@
...
+    coercion of data. This means that if an object does not exactly match the reference
+    type, it will fail validation, regardless of whether it could be coerced into the
+    correct type.
+
+    :param model: Pydantic model to validate against
+    :param importance: Importance of the validator, possible values "warn" and "fail"
+    :param arbitrary_types_allowed: Whether arbitrary types are allowed in the model
+    """
+
+    def __init__(self, model: type[BaseModel], importance: str):
```
elijahbenizzy, line 22: OK, so just a month until this is valid in all supported pythons! We'll be deprecating 3.8 support a bit after that, although we tend to move a bit slower than python versions for doing so.

That said, mind doing `Type[BaseModel]` instead?

hamilton/data_quality/pydantic_validators.py
```diff
@@ -0,0 +1,61 @@
...
+        super(PydanticModelValidator, self).__init__(importance)
+        self.model = model
+
```
elijahbenizzy, line 25: whitespace

hamilton/data_quality/pydantic_validators.py
```diff
@@ -0,0 +1,61 @@
...
+from pydantic import BaseModel, TypeAdapter, ValidationError
+
+from hamilton.data_quality import base
+from hamilton.htypes import custom_subclass_check
+
+
+class PydanticModelValidator(base.BaseDefaultValidator):
+    """Pydantic model compatibility validator
+
+    Note that this validator uses pydantic's strict mode, which does not allow for
```
elijahbenizzy, line 12: A code example would be really great here -- especially if you can add to the docs. Ping us if you want help getting that set up, or it's too much of a pain -- you can add a code example and we can add it to the docs.

hamilton/plugins/h_pydantic.py
```diff
@@ -0,0 +1,32 @@
...
+        self,
+        importance: str = dq_base.DataValidationLevel.WARN.value,
+        target: fm_base.TargetType = None,
+    ):
+        super(check_output, self).__init__(target)
+        self.importance = importance
+        self.target = target
+
+    def get_validators(self, node_to_validate: node.Node) -> List[dq_base.DataValidator]:
+        output_type = node_to_validate.type
```
elijahbenizzy, line 25: Oh, cool, really clever, I see what you did here. The reason this is it's own class is specifically so we can dynamically assign it based on the output type!

Was initially going to suggest you try to reuse `check_output_custom`, but this makes for a really nice UX. Mind adding some comments? It's really cool but a bit hidden as to why there's this class structure.

Musing -- I think the better way overall would be to expose the node to the subclass, but that's more work than it's worth now, and could have backwards compatibility. So yeah, some comments. 

And a docstring that we can put in docs!

elijahbenizzy (commented):
Yeah, this is really cool what you did. Nice work! Really appreciate it!

Things we'll want to get this out:

- [ ] A few more docs/code in the docs
- [ ] A comment or two

Let me know if you have bandwidth for adding to docs -- happy to point you in the right direction/get you set up -- if not, happy to discuss.

elijahbenizzy:
Also, I don't think the tests are getting run. Can dig in tomorrow(this is a bit confusing), but it looks like:
1. We check for changes (this is not getting hit)
2. We launch these 5 jobs: https://github.com/DAGWorks-Inc/hamilton/blob/main/.circleci/config.yml#L137
3. Calls this... https://github.com/DAGWorks-Inc/hamilton/blob/004ac5e4efd8aefa45a1888b3c2df375f2843370/.ci/test.sh#L28
4. Note we only have 8 tests run on main: https://app.circleci.com/pipelines/github/DAGWorks-Inc/hamilton/4033/workflows/dfad0448-70b9-4a89-bd80-e6e6c8265640/jobs/68530. And this one doesn't actually detect a change..

cswartzvi:
@elijahbenizzy sorry, I got overloaded with work the past few days. I am going to add the documentation either today or tomorrow - I will take a crack at it and let you know if I run into issues. Did you ever figure out why my test were not running? If not, I can take a look into that as well.

hamilton/data_quality/pydantic_validators.py
cswartzvi: Yes, that is correct - even with `strict=True` the adapter works on a `pydantic.BaseModel` and a `dict` (I got nervous and double checked 😄): 
```python
import pydantic

class Model(pydantic.BaseModel):
    name: str

adapter = pydantic.TypeAdapter(Model)

adapter.validate_python(Model(name="charles"), strict=True)  #  returns Model(name='charles')

adapter.validate_python({"name": "charles"}, strict=True)  #  returns Model(name='charles')
```

elijahbenizzy:
> @elijahbenizzy sorry, I got overloaded with work the past few days. I am going to add the documentation either today or tomorrow - I will take a crack at it and let you know if I run into issues. Did you ever figure out why my test were not running? If not, I can take a look into that as well.

Thank you! I have not yet -- if you don't mind that would be much appreciated, otherwise happy to carve out some time to figure out what's happening (lots of bash scripts...).

hamilton/data_quality/pydantic_validators.py
elijahbenizzy: Great!

cswartzvi:
I was looking into why my tests didn't run when I pushed the initial set of commits: https://app.circleci.com/pipelines/github/DAGWorks-Inc/hamilton/4035/workflows/c7dbcc07-b6df-4fd1-9a38-a76e81b9cb53

I don't have a lot of experience with CircleCI, so please forgive me if I am missing something basic, but I noticed that the `check_for_changes`  job in `.circleci/config.yml` uses the following  form of the `git diff` command:
```shell
git diff --name-only HEAD^ HEAD | grep '^.ci\|^.circleci\|^graph_adapter_tests\|^hamilton\|^plugin_tests\|^tests\|^requirements\|setup' > /dev/null
```
This command compares the current commit (`HEAD`) to the previous commit (`HEAD^`) before piping it to `grep`. However, I pushed multiple commits at once, so if my very last commit didn't change one of the grepped directories wouldn't this test fail? It is my understanding that CircleCI runs on a push and not for each commit, is that correct?

If my hunch is right, I see three solutions
1. Do nothing and encourage people to make single commit pushes - might miss some tests if people forget
2. Change the `git diff` command to something like `git diff --name-only origin/main...HEAD` and check for changes against the main branch - surely this will result in more tests being run within a PR, but it seems safer
3. Beef up the check for changes with something like https://github.com/emmeowzing/dynamic-continuation-orb

Thoughts? Still working on those docs BTW

elijahbenizzy:
> I was looking into why my tests didn't run when I pushed the initial set of commits: https://app.circleci.com/pipelines/github/DAGWorks-Inc/hamilton/4035/workflows/c7dbcc07-b6df-4fd1-9a38-a76e81b9cb53
> 
> I don't have a lot of experience with CircleCI, so please forgive me if I am missing something basic, but I noticed that the `check_for_changes` job in `.circleci/config.yml` uses the following form of the `git diff` command:
> 
> ```shell
> git diff --name-only HEAD^ HEAD | grep '^.ci\|^.circleci\|^graph_adapter_tests\|^hamilton\|^plugin_tests\|^tests\|^requirements\|setup' > /dev/null
> ```
> 
> This command compares the current commit (`HEAD`) to the previous commit (`HEAD^`) before piping it to `grep`. However, I pushed multiple commits at once, so if my very last commit didn't change one of the grepped directories wouldn't this test fail? It is my understanding that CircleCI runs on a push and not for each commit, is that correct?
> 
> If my hunch is right, I see three solutions
> 
> 1. Do nothing and encourage people to make single commit pushes - might miss some tests if people forget
> 2. Change the `git diff` command to something like `git diff --name-only origin/main...HEAD` and check for changes against the main branch - surely this will result in more tests being run within a PR, but it seems safer
> 3. Beef up the check for changes with something like https://github.com/emmeowzing/dynamic-continuation-orb
> 
> Thoughts? Still working on those docs BTW

I think the second one is probably the cleanest -- better to run more tests than needed than undertest... I can make the change, or if you're in the mood to, go for it! These *are* tests we'll catch when merging to main (it should be smart enough to diff against main), but it's nice to catch earlier.

As an edge-case, we can probably get the comparison branch rather than `main`. That said, at some point we'll probably switch to github actions as our CI system, so I think the simplest code-only change is easy and we shouldn't solve too much.

cswartzvi:
@elijahbenizzy I would be happy to make the change. I will start with option 2 and then see how hard it is to incorporate that edge case.

hamilton/data_quality/default_validators.py
```diff
@@ -508,6 +508,23 @@ def _append_pandera_to_default_validators():
 _append_pandera_to_default_validators()
 
 
+def _append_pydantic_to_default_validators():
+    """Utility method to append pydantic validators as needed"""
+    try:
+        from pydantic import BaseModel  # noqa: F401
+    except ModuleNotFoundError:
+        logger.info(
```
elijahbenizzy, line 516: Let's make this a debug? Also, maybe import pydantic rather than BaseModel, as I don't think we use it?

hamilton/plugins/h_pydantic.py
```diff
@@ -0,0 +1,32 @@
...
+from hamilton import node
+from hamilton.data_quality import base as dq_base
+from hamilton.function_modifiers import InvalidDecoratorException
+from hamilton.function_modifiers import base as fm_base
+from hamilton.function_modifiers import check_output as base_check_output
+from hamilton.function_modifiers.validation import BaseDataValidationDecorator
+from hamilton.htypes import custom_subclass_check
+
+
+class check_output(BaseDataValidationDecorator):
```
elijahbenizzy, line 14: Add a docstring -- we can add this to the docs!

elijahbenizzy (commented):
Basically there, just a few notes about docs (which I know is a WIP). To get shipped:
- [ ] Add some more docstrings + add to the right docs reference (ping if you want help, but start with this slack comment)
- [ ] Rebase against the check_for_changes PR after we merge to `main` ensure that it does as intended on this

cswartzvi:
@elijahbenizzy I rebased and added that missing docstring - looks like changes were detected correctly in CI. How deep do you think I should go with the documentation? I see three potential places it could be added:
- https://hamilton.dagworks.io/en/latest/concepts/function-modifiers/
- https://hamilton.dagworks.io/en/latest/how-tos/run-data-quality-checks/
- https://github.com/DAGWorks-Inc/hamilton/tree/main/examples/data_quality

Does that sound reasonable?

cswartzvi:
@elijahbenizzy I rebased and added that missing docstring - looks like changes were detected correctly in CI. How deep do you think I should go with the documentation? I see three potential places it could be added:
- https://hamilton.dagworks.io/en/latest/concepts/function-modifiers/
- https://hamilton.dagworks.io/en/latest/how-tos/run-data-quality-checks/
- https://github.com/DAGWorks-Inc/hamilton/tree/main/examples/data_quality

Does that sound reasonable?

elijahbenizzy:
> @elijahbenizzy I rebased and added that missing docstring - looks like changes were detected correctly in CI. How deep do you think I should go with the documentation? I see three potential places it could be added:
> 
> * https://hamilton.dagworks.io/en/latest/concepts/function-modifiers/ [ I ended up adding this already]
> * https://hamilton.dagworks.io/en/latest/how-tos/run-data-quality-checks/
> * https://github.com/DAGWorks-Inc/hamilton/tree/main/examples/data_quality
> 
> Does that sound reasonable?

Sorry for the back and forth, this is ready! Only thing, re: docs:

I think what you did is good enough out of those three. If you want to add an example it would obviously be appreciated but don't want to slow you down. The place that I'd add it is here: https://hamilton.dagworks.io/en/latest/reference/decorators/. This way we have a nice reference with the docstring.

To do so you can:
1. Add a file here with a brief description (call it `pydantic.rst` or something
2. Make it reference the relevant decorator (`h_pydantic.check_output`). mention that you can get it with check_output to make that clear.
3. Add a newline between this and the next line (`.rst` is an epic PITA): https://github.com/DAGWorks-Inc/hamilton/pull/1121/files#diff-c0dc4b429a8dd227d562cd69242be34a5b19fc49a48f98b7267dc722767d34a3R30 -- otherwise it won't compile

If this is too much for you/you don't have time, let me know and I can add it in easily after merging. You've already done a ton! Otherwise I want to get this out in the next release (tuesday if we can!).

Really appreciate it 🫡

cswartzvi:
> Sorry for the back and forth, this is ready! Only thing, re: docs:
> 
> I think what you did is good enough out of those three. If you want to add an example it would obviously be appreciated but don't want to slow you down. The place that I'd add it is here: https://hamilton.dagworks.io/en/latest/reference/decorators/. This way we have a nice reference with the docstring.
> 
> To do so you can:
> 
> 1. Add a file here with a brief description (call it `pydantic.rst` or something
> 2. Make it reference the relevant decorator (`h_pydantic.check_output`). mention that you can get it with check_output to make that clear.
> 3. Add a newline between this and the next line (`.rst` is an epic PITA): https://github.com/DAGWorks-Inc/hamilton/pull/1121/files#diff-c0dc4b429a8dd227d562cd69242be34a5b19fc49a48f98b7267dc722767d34a3R30 -- otherwise it won't compile
> 
> If this is too much for you/you don't have time, let me know and I can add it in easily after merging. You've already done a ton! Otherwise I want to get this out in the next release (tuesday if we can!).
> 
> Really appreciate it 🫡

No problem, I enjoy helping! I would like to add at least one more doc - I have something queued up, just one quick question. You mentioned adding `pydantic.rst` to https://hamilton.dagworks.io/en/latest/reference/decorators/ do you think adding it directly to https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/ might be better? This was my original idea because it doesn't break the meaning of the listed decorator and it fits in nicely with `check_output.rst` (specifically I would up have to update [this](https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/#:~:text=Note%20we%20also%20have%20a%20plugin%20that%20allows%20you%20to%20use%20pandera.%20There%20are%20two%20ways%20to%20access%20it%3A%201.%20%40check_output(schema%3Dpandera_schema)%202.%20%40h_pandera.check_output()%20on%20a%20function%20that%20declares%20a%20typed%20pandera%20dataframe%20as%20an%20output) paragraph and add a reference). Either way works for me, just let me know.

cswartzvi:
> Sorry for the back and forth, this is ready! Only thing, re: docs:
> 
> I think what you did is good enough out of those three. If you want to add an example it would obviously be appreciated but don't want to slow you down. The place that I'd add it is here: https://hamilton.dagworks.io/en/latest/reference/decorators/. This way we have a nice reference with the docstring.
> 
> To do so you can:
> 
> 1. Add a file here with a brief description (call it `pydantic.rst` or something
> 2. Make it reference the relevant decorator (`h_pydantic.check_output`). mention that you can get it with check_output to make that clear.
> 3. Add a newline between this and the next line (`.rst` is an epic PITA): https://github.com/DAGWorks-Inc/hamilton/pull/1121/files#diff-c0dc4b429a8dd227d562cd69242be34a5b19fc49a48f98b7267dc722767d34a3R30 -- otherwise it won't compile
> 
> If this is too much for you/you don't have time, let me know and I can add it in easily after merging. You've already done a ton! Otherwise I want to get this out in the next release (tuesday if we can!).
> 
> Really appreciate it 🫡

No problem, I enjoy helping! I would like to add at least one more doc - I have something queued up, just one quick question. You mentioned adding `pydantic.rst` to https://hamilton.dagworks.io/en/latest/reference/decorators/ do you think adding it directly to https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/ might be better? This was my original idea because it doesn't break the meaning of the listed decoroator and it fits in nicely with `check_output.rst` (specifically I would up have to update [this](https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/#:~:text=Note%20we%20also%20have%20a%20plugin%20that%20allows%20you%20to%20use%20pandera.%20There%20are%20two%20ways%20to%20access%20it%3A%201.%20%40check_output(schema%3Dpandera_schema)%202.%20%40h_pandera.check_output()%20on%20a%20function%20that%20declares%20a%20typed%20pandera%20dataframe%20as%20an%20output) paragraph and add a reference). Either way works for me, just let me know.

elijahbenizzy:
> > Sorry for the back and forth, this is ready! Only thing, re: docs:
> > I think what you did is good enough out of those three. If you want to add an example it would obviously be appreciated but don't want to slow you down. The place that I'd add it is here: https://hamilton.dagworks.io/en/latest/reference/decorators/. This way we have a nice reference with the docstring.
> > To do so you can:
> > 
> > 1. Add a file here with a brief description (call it `pydantic.rst` or something
> > 2. Make it reference the relevant decorator (`h_pydantic.check_output`). mention that you can get it with check_output to make that clear.
> > 3. Add a newline between this and the next line (`.rst` is an epic PITA): https://github.com/DAGWorks-Inc/hamilton/pull/1121/files#diff-c0dc4b429a8dd227d562cd69242be34a5b19fc49a48f98b7267dc722767d34a3R30 -- otherwise it won't compile
> > 
> > If this is too much for you/you don't have time, let me know and I can add it in easily after merging. You've already done a ton! Otherwise I want to get this out in the next release (tuesday if we can!).
> > Really appreciate it 🫡
> 
> No problem, I enjoy helping! I would like to add at least one more doc - I have something queued up, just one quick question. You mentioned adding `pydantic.rst` to https://hamilton.dagworks.io/en/latest/reference/decorators/ do you think adding it directly to https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/ might be better? This was my original idea because it doesn't break the meaning of the listed decoroator and it fits in nicely with `check_output.rst` (specifically I would up have to update [this](https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/#:~:text=Note%20we%20also%20have%20a%20plugin%20that%20allows%20you%20to%20use%20pandera.%20There%20are%20two%20ways%20to%20access%20it%3A%201.%20%40check_output(schema%3Dpandera_schema)%202.%20%40h_pandera.check_output()%20on%20a%20function%20that%20declares%20a%20typed%20pandera%20dataframe%20as%20an%20output) paragraph and add a reference). Either way works for me, just let me know.
> 
> Edit: I pushed an example of what I am talking about. If that's not what you're looking for just let me know 😄



> directly to https://hamilton.dagworks.io/en/latest/reference/decorators/check_output/ might be better? This was my original idea because it doesn't break the meaning of the listed decoroator and it fits in nicely with `check_output.rst` (specifically I would

Yes, I think that's pretty reasonable, perhaps a better place to put it! Only thing is to add the autoclass or whatever there (and just make it clear which maps to which). All nits though, this is pretty much good to go as far as I'm concerned!

elijahbenizzy (approved):
Looks good, let's do the last docs stuff then merge! Appoving so all I have to do is click merge next.

cswartzvi:
I added the autoclass to the bottom of the page and tightened up a few of the examples. Let me know if it needs anything else!

elijahbenizzy:
> I added the autoclass to the bottom of the page and tightened up a few of the examples. Let me know if it needs anything else!

Looks great, thank you! Merging!

Issue #1124, opened 2024-09-05
parameterize decorator should add values to default_parameter_values


vograno:
**Is your feature request related to a problem? Please describe.**
When a vanilla node is created the default values of its parameters, if any, are available via `node.default_parameter_values` attribute. However, when a node is created by `@parameterize`, the values passed to the decorator are not available via the attribute. In fact they are not available from the node object at all.

The point I am making is that those values are  essentially  default values for the expanded node and should be surfaced  in the same fashion. This feature would make expanded nodes  equivalent to vanilla nodes in this regards.

**Describe the solution you'd like**
Define a node
```
x_value = {}
@parameterize(output_node=dict(x=value(x_value))
def fun(x):
    return x
```

The following should hold
```
node = driver.graph.nodes['output_node']
assert 'x' in node.default_parameter_values
assert node.default_parameter_values['x'] is x_value
```

skrawcz:
To clarify, you wouldn't expect the following to work normally though right?

```python
# this should not work
result = driver.execute(final_vars=['output_node'], inputs={"x": {"some": "value"}})
```

But I think I see where you're coming from now I think. The tricky part is how would you handle this:

```python
x_value = {}
y_value = {}
@parameterize(
  output_node=dict(x=value(x_value), 
  output_node2=dict(x=value(y_value))
)
def fun(x: dict) -> dict:
    return x
```
Or is the case that all of them have a the same default value?

In which case in the interim you could do something functionally equivalent like

```
@tag(marker="default") # could identify this as such
def x() -> dict:
   return {} # or reference some variable like you have above in the module

@tag(marker="parameterized")
@parameterize(output_node=dict(x=source(x))
def fun(x: dict) -> dict:
    return x
```
then:

1. ask the graph for all parameterized nodes via `dr.list_available_variables({"TAG_NAME": "TAG_VALUE"})`
2. ask the graph for all default nodes via `dr.list_available_variables({"TAG_NAME": "TAG_VALUE"})`
3. figure out which default values
4. default nodes provide the originating function, so you can exercise it to get the value.
5. if you wanted to change that value, you could just use `overrides=` via .execute()...

vograno:
> To clarify, you wouldn't expect the following to work normally though right?
> 
> ```python
> # this should not work
> result = driver.execute(final_vars=['output_node'], inputs={"x": {"some": "value"}})
> ```

This shouldn't work indeed.

[...]

PR #1125, opened 2024-09-07
Sketches possible fix to help #1124

skrawcz:
We're inconsistent as to when originating_functions is populated. So this tries to fix it for expander parameterize decorator at least.

## Checklist


skrawcz:
Related to #1124

PR #1126, opened 2024-09-08
Bug fix when using `HamiltonTracker` with polars dataframes 

elyase:
Fixes an error when tracking in the hamilton ui with polars dataframes:

```
    return col.str.lengths()
           ^^^^^^^^^^^^^^^
AttributeError: 'StringNameSpace' object has no attribute 'lengths'
```

## Changes

Use the [correct polars method](https://docs.pola.rs/api/python/stable/reference/series/api/polars.Series.str.len_chars.html)

## How I tested this

Run the code and it doesn't error anymore

## Checklist

### DIFF
--- a/ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
+++ b/ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
@@ -147,7 +147,7 @@ def datetime_column_stats(
 
 
 def str_len(col: pl.Series) -> pl.Series:
-    return col.str.lengths()
+    return col.str.len_chars()
 
 
 def avg_str_len(str_len: pl.Series) -> float:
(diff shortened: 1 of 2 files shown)


elijahbenizzy:
Hey! Thanks for this -- seems reasonable on first glance. 

Two thoughts:
1. Is this due a polars version? Might be that it used to work. 
2. Looks like this was not added in the tests, so my guess on (1) is it never worked: https://github.com/DAGWorks-Inc/hamilton/blob/004ac5e4efd8aefa45a1888b3c2df375f2843370/ui/sdk/tests/tracking/test_polars_col_stats.py Would be nice to have tests (if you want to add, should be easy to copy/paste/adjust those), but I don't want this to be a blocker to you getting your fix in.

So if you're up to the challenge of adding tests feel free to, otherwise we can add some a bit later.

elyase:
@elijahbenizzy yes the names changed after polars==0.18. I added a test for the fix

elijahbenizzy:
> @elijahbenizzy yes the names changed after polars==0.18. I added a test for the fix

Great -- this looks good. Mind running pre-commit changes? `pre-commit run --all` should do it. Then we'll merge! Appreciate the fix/test 🙌

elyase:
✔️ thanks to you all for hamilton!

elijahbenizzy (approved):
Thank you for your contribution! Can do a quick patch release for the SDK shortly.

PR #1131, opened 2024-09-10
Bumps version from 1.75.0 to 1.76.0

elijahbenizzy:
New release!


elijahbenizzy:
Whoops commit message, bumped from 1.75.1... OK for now, just leaving in case people are curious about the mismatch.

Issue #1120, opened 2024-09-04
Support node overriding when combining multiple modules


ytang137:
Sometimes it's convenient to compose a DAG by combining multiple modules with `dr = driver.Builder().with_modules(module_a, module_b)`. If `module_a` and `module_b` have duplicate nodes, currently the builder will throw an error. It can be useful to have an option to allow the node in the later module to override the one in an earlier module, perhaps something like `.allow_module_overrides()` - as suggested by Elijah ben Izzy on Slack.

elijahbenizzy:
Copying for persistence:

```
Yijun Tang
  Tuesday at 2:43 PM
Hi, can I ask what later calls will take precedence refers to in the docstring of [with_modules](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/driver.py#L1757)? I was wondering if it’s possible to have 2 modules, module_a and module_b , that both implement a common_node node, and calling builder.with_module(module_a) and then builder.with_module(module_b) would use the common_node from module_b to build the DAG - but this doesn’t seem to be the case.

[driver.py](https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/driver.py)
    def with_modules(self, *modules: ModuleType) -> "Builder":
<https://github.com/[DAGWorks-Inc/hamilton](https://github.com/DAGWorks-Inc/hamilton)|DAGWorks-Inc/hamilton>DAGWorks-Inc/hamilton | Added by [GitHub](https://hamilton-opensource.slack.com/services/B03M52ALBGB)
:eyes:
1



Elijah ben Izzy
  [Tuesday at 2:45 PM](https://hamilton-opensource.slack.com/archives/C03M33QB4M8/p1725399942869669?thread_ts=1725399812.696459&cid=C03M33QB4M8)
Hey — so I think that this docstring is wrong. In that it uses all of them together — it will find two definitions for common_node, and error out. Just to ensure — what happens when you try?
Otherwise “overriding” functions is something we’re thinking about but I don’t think we have the capability yet.
The common approach is to switch between a/b depending on a config, and parse out common ones.


Elijah ben Izzy
  Tuesday at 2:47 PM
To clarify — we have @config.when to help with this type of thing!
e.g.
@config.when(context="a")
def common_node__a() -> ...:
    ...
 other module:
@config.when(context="b")
def common_node__b() -> ...:
    ...
 (edited) 


Yijun Tang
[...]

PR #1134, opened 2024-09-11
Override nodes at module import - Feature Sketch

jernejfrank:
This addresses #1120 and adds the capability to override nodes from later imported modules. It also suggests how to handle the case when we import the same module twice.

## Changes
Adds ```allow_module_overrides``` as a method in the Builder and as an attribute in the Driver.  If set true, it allows the node with the same name to be overwritten.

## How I tested this
There are only two small mock scripts to have a running sketch; I will add unit tests.

## Checklist


zilto:
Sharing Slack comments for lineage

Thanks for the contribution! I see the thread of issues and Slack convo.
I think it's worth discussing the feature a bit more before you commit additional time to it!
For one, relying on ordering of arguments can be brittle. For example .with_modules(foo, bar, allow_override=True) might not equal .with_modules(bar, foo, allow_override=True). It's good to use the flag allow_override to prevent users accidentally overriding functions though
Since this change involves many core internals (FunctionGraph, Driver  , Builder) I'm wondering if there are other ways to support this feature. Hamilton couples a lot of thing to the module Python construct (for better or worse). But concretely, the issue seems to be overriding functions. It might be a valuable change to catch duplicate functions "closer to the user" (e.g., Builder.build() rather than in internal components like FunctionGraph 

An idea would be to use find_functions() , which is a public-facing API part of hamilton.graph_utils. Ideally, it would allow us to only make "surface level" changes closer to the Builder.
## Default behavior
We can keep .with_modules() as-is and avoid changes to other internals (Driver, FunctionGraph). This ensures safety for all Hamilton users
## Override behavior
When specifying .with_modules(..., allow_overrides=True),
the logic of .with_modules() could directly call find_functions() and dynamically removed duplicate functions

```python
def with_modules(*modules):
   if allow_overrides:
      functions_seen = set()
      for module in modules:
         functions = graph_utils.find_functions(module)
         for name, callabl in functions:
            if name in functions_seen
               del getattr(module, name)
            else:
               functions_seen.add(name)
   
    self.modules.extend(modules)
```

elijahbenizzy:
> This addresses #1120 and adds the capability to override nodes from later imported modules. It also suggests how to handle the case when we import the same module twice.

@zilto this specifically doesn't allow for *node* overriding, this is only function overriding. Burt also, there is no inherent problem changing the core functionality -- it shouldn't be static. In this case the graph is responsible for assembling the nodes. It also has the current error message that breaks, so it makes sense to change that. We have more flexibility if we adjust the internal API to change later than using user-facing functions, which tie us to the external API.

hamilton/driver.py
```diff
@@ -391,6 +391,7 @@ def __init__(
         adapter: Optional[
             Union[lifecycle_base.LifecycleAdapter, List[lifecycle_base.LifecycleAdapter]]
         ] = None,
+        allow_module_overrides: bool = False,
```
elijahbenizzy, line 394: This is good to have here for the old (non-builder) API/passing through -- add to docstring?

hamilton/driver.py
```diff
@@ -417,8 +418,15 @@ def __init__(
             adapter.call_all_lifecycle_hooks_sync("pre_do_anything")
         error = None
         self.graph_modules = modules
+        # Feels like this should be public class attribute.
+        self.allow_module_overrides = allow_module_overrides
```
elijahbenizzy, line 423: Start non-public IMO, then we can make it public. That said, we're not doing a good job here of differentiating, so just don't worry about it.

hamilton/driver.py
```diff
@@ -1818,6 +1826,8 @@ def __init__(self):
         # common fields
         self.config = {}
         self.modules = []
+        # Allow later modules to override nodes of the same name
+        self.module_overrides = False
```
elijahbenizzy, line 1831: nit -- I'd call this `_allow_module_overrides`

hamilton/driver.py
```diff
@@ -1979,6 +1989,17 @@ def with_grouping_strategy(self, grouping_strategy: grouping.GroupingStrategy) -
         self.grouping_strategy = grouping_strategy
         return self
 
+    # I think we need this here or similarly .with_modules(allow_overrides=True),
```
elijahbenizzy, line 1993: Agreed we want the API central

hamilton/driver.py
```diff
@@ -2010,13 +2031,25 @@ def build(self) -> Driver:
                 adapter=lifecycle_base.LifecycleAdapterSet(*adapter),
             )
 
+        if not self.module_overrides:  # if override on than this doesn't matter
```
elijahbenizzy, line 2035: Hmm, interesting... This is actually a different type of module override -- those that are the same. Can you elaborate more on the case that it broke? Feels like an upstream thing... I'm not 100% sure I trust the `module in module_set` -- identity is a little weird/what happens if you make an update but keep an old reference around?

hamilton/graph.py
```diff
@@ -169,7 +170,7 @@ def create_function_graph(
         for n in fm_base.resolve_nodes(f, config):
             if n.name in config:
                 continue  # This makes sure we overwrite things if they're in the config...
-            if n.name in nodes:
+            if n.name in nodes and not allow_module_overrides:
                 raise ValueError(
                     f"Cannot define function {n.name} more than once."
                     f" Already defined by function {f}"
```
elijahbenizzy, line 176: You can add an error message here that links to the docs in case people want to use it -- now that we have a good approach to fixing this. Just make sure to outline the caveats.

elijahbenizzy (commented):
Looking good! Left some comments on specifics. High-level:
1. You'll want docs (see slack for how to add/develop). Make sure to reference caveats (why you might/might not want to do this)
2. Wondering if we want more in the example -- a README/notebook might be nice. If you have a compelling version of why this feature is useful that woudl be nice, otherwise don't want to upscope this! You're not the one that asked for this feature, you're the one building it :)

docs/concepts/builder.rst
```diff
@@ -73,6 +73,22 @@ It encourages organizing code into logical modules (e.g., feature processing, mo
         .build()
     )
 
+.. note::
+
+    Your modules may have same named functions which will raise an error when using ``.build()`` since we cannot have two functions with the same name. You can use the method ``.allow_module_overrides()`` and Hamilton will choose the function from the later imported module.
```
elijahbenizzy, line 78: Two "nodes" with the same name -- fns -> nodes are 1:many.  Might be worth mentioning that.

examples/module_overrides/run.py
```diff
@@ -0,0 +1,22 @@
+import module_a
```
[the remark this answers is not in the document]
jernejfrank: I am on the fence about having this in the examples as I don't really have a use case for it. Maybe we remove it for now and ask Yijun Tang who opened the issue if he can give a rough sketch of how he will use it. Happy to implement it, just don't want it to be a very contrived example.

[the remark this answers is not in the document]
elijahbenizzy: Yep! So just a simple README then might be nice -- one that says what this does/a quick description of the feature. Can ignore a notebook until we have it ready.

hamilton/driver.py
jernejfrank: yes, if  you do something like
```
.with_modules(module_A, module_A)
```
hamilton will just add it to the list of modules and treat them as two different ones. If you print the attribute, you get two pointers in the list
```
[<module 'module_A' from 'path to module A'>,<module 'module_A' from 'path to module A'>]
```
Then when you try to build the driver, because they are treated as separate you run into the issue that you have same-named functions and get an error that functions have to have different names.

The ``module_set`` tries to get create an ordered set out of the list we use for storing the module references.

examples/module_overrides/z_double_import_problem.py
```diff
@@ -0,0 +1,139 @@
...
+    nodes["b"].depended_on_by.append(nodes["A"])
+    nodes["c"].depended_on_by.append(nodes["A"])
+    nodes["B"].dependencies.append(nodes["A"])
+    nodes["C"].dependencies.append(nodes["A"])
+    return nodes
+
+
+def create_testing_nodes_override_B():
+    """Helper function for creating the nodes represented in dummy_functions.py."""
+    nodes = {
```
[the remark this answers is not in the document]
jernejfrank: Removed the script since it got implemented in the unit test

jernejfrank:
I feel like the README is the most concise description of this functionality, so I am wondering if we should link to it more?

tests/test_graph.py
```diff
@@ -455,13 +456,72 @@ def create_testing_nodes():
     return nodes
 
 
+def create_testing_nodes_override_B():
```
elijahbenizzy, line 459: Same note about node.Node.from_fn

elijahbenizzy:
Looking good -- to get it out:
- [ ] Let's decide whether we want the additional clashing modules in -- if so, let's add a driver level test
- [ ] Add a driver-level test folr this feature (we have driver code)
See this file for E2E tests -- good enough for this IMO: https://github.com/DAGWorks-Inc/hamilton/blob/a7a2c8c85dbafa59680e554d9d358020af1240e7/tests/test_end_to_end.py#L436

elijahbenizzy:
> I feel like the README is the most concise description of this functionality, so I am wondering if we should link to it more?

Yep, agreed. Can link from docstring and/or from concepts section.

hamilton/driver.py
elijahbenizzy: Yeah, I mean, maybe a separate PR? Or an issue with the proposed fix?

elijahbenizzy (approved):
This looks good, thank you!

hamilton/graph.py
elijahbenizzy: We can add this after the fact, but on second thought we don't need to advertise this as the first solution as it's a slight power-user case.

PR #1141, opened 2024-09-14
Add command options to install dependency.

seyawudba:

This change was necessary to allow users install dependencies seamlessly without any problems.

## Changes
- Extra commands were added to the developer_setup.md file to allow users whose terminals wouldn't execute the dependency install command to also install dependencies without any struggle.

## How I tested this
- A preview was made in the vscode editor to see how the changes would look when pushed.
- The pre-commit checks were also run and everything passed successfully.

## Notes
- I didn't remove the earlier command because I think it might work for users.

## Checklist

### DIFF
--- a/developer_setup.md
+++ b/developer_setup.md
@@ -29,6 +29,14 @@ python -m venv ./venv
 pip install .[dev,test]
 ```
 
+Please be informed to use the command below as the square bracket might not be recognized as a pattern.
+
+```shell
+pip install .\[dev,test\]
+
+pip install '.[dev,test]'
+```
+
 Set up `pre-commit`, which will run some lightweight formatting and linting tasks on every commit.
 
 ```shell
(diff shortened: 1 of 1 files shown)


developer_setup.md
```diff
@@ -29,6 +29,14 @@ python -m venv ./venv
 pip install .[dev,test]
 ```
 
+Please be informed to use the command below as the square bracket might not be recognized as a pattern.
+
+```shell
+pip install .\[dev,test\]
+
+pip install '.[dev,test]'
```
elijahbenizzy, line 37: The second one (single quotes) should always work -- just have that without square brackets?

seyawudba: Could you please come clearer on your question?

elijahbenizzy: There are two pip install commands. Do we need both?

elijahbenizzy (approved):
Thank you!

developer_setup.md
seyawudba: Yes, both of them work but I had to remove one. So I think we are good.

Issue #749, opened 2024-03-08
Add @post_pipe


skrawcz:
**Is your feature request related to a problem? Please describe.**
`@pipe` is nice, but can be a little counter-intuitive to read. i.e. the function parameter declares the dependency, is then modified by `@pipe` and then passed into the function.

**Describe the solution you'd like**
`@post_pipe` would be the opposite of `@pipe`. The function would declare the dependency, the body of the function would run, and then the `@post_pipe` steps would run after the function.

```python
@post_pipe(# these run after the function is run.
    step(_transform_1, v=1),
    step(_transform_2, v=2),
)
def A_processed(A: pd.DataFrame) -> pd.DataFrame:
    return A # original A -- could do modifications here.
```

**Describe alternatives you've considered**
N/A

**Additional context**

elijahbenizzy:
@skrawcz this is a duplicate https://github.com/DAGWorks-Inc/hamilton/issues/701

Issue #701, opened 2024-02-15
Post-function modification (opposite/equivalent to `pipe`)


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
People often want data quality checks to modify the output, but that's really not what they're supposed to do.
`@pipe` applies a function, but it is run beforehand. So if you just want to change the output of a node, you have to do it in two steps.

**Describe the solution you'd like**
New decorator that functions like `pipe`, but runs afterwards:

```python
@mutate(
    step(_keep_range, range=(0,100)),
    step(_dropna)
)   
def data() -> pd.Series:
    return ...
```

This would form the DAG: `data.raw -> data.with_drop_between -> data` 

**Describe alternatives you've considered**
Making it more central to data quality or just integating with the node.

**Additional context**
Related to an OS question.

zilto:
I think there's a place for this along with `@pipe`. Also the `data.raw -> data` fits the `@check_output` semantic.

Although, I propose to rename it `@pipe_out` `@pipe_output` or `@pipe_post`. Maybe we'd rename `@pipe` to `@pipe_in` (input, or pre) too?

Although these decorators are not much different than writing additional nodes, I think they can greatly facilitate migration by allowing to plug in an existing codebase into new/smaller Hamilton initiatives

elijahbenizzy:
> I think there's a place for this along with `@pipe`. Also the `data.raw -> data` fits the `@check_output` semantic.
> 
> Although, I propose to rename it `@pipe_out` `@pipe_output` or `@pipe_post`. Maybe we'd rename `@pipe` to `@pipe_in` (input, or pre) too?
> 
> Although these decorators are not much different than writing additional nodes, I think they can greatly facilitate migration by allowing to plug in an existing codebase into new/smaller Hamilton initiatives

Feels like `@modify_input` and `@modify_output` might be better than `pipe`?

zilto:
I like your @pipe documentation and the fact that is echoes the pandas `.pipe()` operation. Apparently, Polars also [has a pipe operator](https://docs.pola.rs/py-polars/html/reference/dataframe/api/polars.DataFrame.pipe.html)

elijahbenizzy:
From @skrawcz on #749 (duplicate):

**Is your feature request related to a problem? Please describe.**
`@pipe` is nice, but can be a little counter-intuitive to read. i.e. the function parameter declares the dependency, is then modified by `@pipe` and then passed into the function.

[...]

PR #1145, opened 2024-09-18
post_pipe implementation

jernejfrank:
Addressing #701 and #749 .

Sketch at the moment. One thing I want to add is an identity function for the last node that dynamically takes over the signature of the last post_pipe node. I couldn't find a way to do this easily, ideas?

## Changes
- Added post_pipe to mutate the output of a function through a series of steps.

## How I tested this
- Added some unit and e2e tests, will expand

## Checklist


hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
+        :param collapse: Whether to collapse this into a single node. This is not currently supported.
+        :param _chain: Whether to chain the first parameter. This is the only mode that is supported. Furthermore, this is not externally exposed. @flow will make use of this.
+        """
+        self.transforms = transforms
+        self.collapse = collapse
+        self.chain = _chain
+        self.namespace = namespace
+
+        if self.collapse:
+            raise NotImplementedError(
```
elijahbenizzy, line 902: This, by the way, is. a nice feature to add/a good one if you're interested at some point (at least to open up an issue)

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
 #         super(flow, self).__init__(*transforms, collapse=collapse, _chain=False)
+
+
+class post_pipe(base.NodeInjector):
+    def __init__(
+        self,
+        *transforms: Applicable,
+        namespace: NamespaceType = ...,
+        collapse=False,
+        _chain=False,
```
elijahbenizzy, line 887: I'd take away `_chain` -- it doesn't look like we ever got around to implementing `@flow` 😆

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
+        :param config:
+        :param fn:
+        :return:
+        """
+
+        # We want the output of post-pipe to connect downstream, so the output needs to be named the
+        # same as the original function. We rename the first function to fn_raw and bind it as the
+        # first parameter. When the pipe nodes are wired through we rename the last node as the
+        # original function name
+        original_node = nodes[-1].copy_with(name=f"{nodes[-1].name}_raw")
```
elijahbenizzy, line 925: Oh, cool, I was confused as to why you were selecting the last, but in a topologically sorted graph it always should be, clever.

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
+
+        if self.collapse:
+            raise NotImplementedError(
+                "Collapsing step() functions as one node is not yet implemented for pipe(). Please reach out if you want this feature."
+            )
+
+        if self.chain:
+            raise NotImplementedError("@flow() is not yet supported -- this is ")
+
+    def transform_dag(
```
elijahbenizzy, line 909: So I think this indicates that a node injector is not correct -- E.G. you want some variant of a `SubdagTransformer`(which is the correct one). But you're basically doing that by overriding `transform_dag`. 

Your logic appears correct, but I think this might simplify

I think what you want is a `SingleNodeNodeTransformer`, which automatically applies to the sink node. This is akin to `extract_columns`, which applies to the last node in a chain. See this: https://github.com/DAGWorks-Inc/hamilton/blob/b818c15eb66c4d175a48d89087cbe7b62b6c448b/hamilton/function_modifiers/base.py#L557. 

From a graph algebra perspective, what we do is:

1. Take a subdag 
2. Find the "last" node in the subdag (if there are multiple we error out). (this + (1) should be handled by `SingleNodeNodeTransformer`
3. Append one node for each step (this is where we reuse the code from `@pipe` -- you do this with `self.inject_nodes`)
4. Ensure the final node has the name of the original node (you do this as well, nice).
5. Rename the first node (original node) to be `_raw` (you do this as well)

So yes, I believe what you have is correct, but I'm wondering if that class will save you some effort!

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
+    ) -> List[node.Node]:
+        """Injects nodes into the graph. This creates a node for each pipe() step,
+        then reassigns the inputs to pass it in."""
+
+        # I think the better approach would be to draw from the last node another `identity` node
+        # that has the correct name but just passes the value of the last pipe node through. This
+        # would allow us to display all pipe transforms in the DAG, otherwise the last one gets
+        # re-named and we effectively hide a node, but I am stuck at how to dynamically add the
+        # correct typehints
+        def __identity(foo: Any) -> Any:
```
elijahbenizzy, line 948: So re: typing, we have `input_types` in node (as we discussed earlier) which should help. The `copy_with` allows you to overwrite it.

That said, I'm going back and forth on whether we want an identity node, or we just want to name it with the name of the last one. 
- With `__identity` we preserve the transformations
- With just the name of the node, we simplify the DAG

Thoughts?

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +876,134 @@ def optional_config(self) -> Dict[str, Any]:
...
+            return foo
+
+        self.transforms = self.transforms + (step(__identity).named(fn.__name__),)
+
+        # since the dict is always length 1 just gets single value out
+        current_param = next(iter(params))
+        fn_count = Counter()
+        nodes = []
+        for applicable in self.transforms:
```
elijahbenizzy, line 957: This code looks correct, and I think you'll be able to share with `@pipe` a little later on (it was copies, right?)

elijahbenizzy (commented):
Really impressed you finagled this API into getting it functional, awesome job!

I left some suggestions (knowing you have plans to simplify) -- haven't tried it myself but I think it might simplify things. Kudos -- you really dove into the complexity here!

hamilton/function_modifiers/macros.py
jernejfrank: Agree, can open separate PR after we finish the core functionality.

hamilton/function_modifiers/macros.py
jernejfrank: Is `@flow` dead or still on the agenda?  If it is still good feature, shouldn't take too long to implement after we finish here and I can leave this in. Otherwise, I take it out.

hamilton/function_modifiers/macros.py
jernejfrank: I am leaning towards keeping the `__identity` to have full transparency in the DAG. I guess I am more comfortable to err on the side of having things a bit cluttered compared to hiding a full transformation of the node.

Looking a bit into the future -- if we implement `collapse`, the user will have the option to simplify the DAG if he wants to.

Let me know if I am missing something about clustering that would mess up UX?

hamilton/function_modifiers/macros.py
jernejfrank: I had a dilemma about whether to use node injector or node transformer. In this case both seem to describe the desired graph behaviour (either injecting node or transforming the last node). I went with the injector at the end to keep it the same as `pipe` and seemed to me that as long as I respect the API of `transform_subdag` I can override the inner part. 

But looking back I should have just went with the base that has the most fitting functionality. Still on my todo to refactor the common things from `pipe` and `post_pipe` out.

hamilton/function_modifiers/macros.py
elijahbenizzy: Yeah, IIRC `@flow` was specifically a procedural approach. `@pipe` without the chaining implication. Definitely a cool feature, and we should have it, but we have not focused. If you're interested then go for it, but let's get `@pipe`/`@post_pipe` working

hamilton/function_modifiers/macros.py
elijahbenizzy: So I think `NodeInjector` is problematic because it leaves the first node in. E.G. NodeInject maps node -> list[nodes], then appends. This is (I think) why you implemented `transform_subdag` -- which isn't part of the interface. IMO `SingleNodeNodeTransformer` will likely solve some of the issues for you? Should make the code simpler.

hamilton/function_modifiers/macros.py
elijahbenizzy: As long as it's namespaced, clustering will work. And yep, see the point -- happy to go with it for now!

hamilton/function_modifiers/macros.py
```diff
@@ -859,6 +859,185 @@ def optional_config(self) -> Dict[str, Any]:
...
+        @post_pipe(
+            step(_add_one),
+            step(_multiply, y=2),
+            step(_sum, y=value(3)),
+            step(_multiply, y=source("upstream_node_to_multiply")),
+        )
+        def run_function(upstream_int: int) -> int:
+            return upstream_int
+
+    .. code-block:: python
```
elijahbenizzy, line 921: Hmm now that I think about it the nested stuff is pretty confusing...

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +856,322 @@ def optional_config(self) -> Dict[str, Any]:
...
+        This is mainly for the legacy API.
+        """
+        out = {}
+        for applicable in self.transforms:
+            for resolver in applicable.resolvers:
+                out.update(resolver.optional_config)
+        return out
+
+
+# Should this become an abstract class for pipe
```
elijahbenizzy, line 1126: I think it's pretty clean to keep it as a helper function -- abstract classes are good when you have more than one implementation that share state/data. In this case it's one tool that it's built around and it would take some wrangling to get it to work with a class.

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +856,322 @@ def optional_config(self) -> Dict[str, Any]:
...
+# Or better question, when do you decide to make an abstract factory?
+def chain_nodes(
+    _first_arg: str,
```
elijahbenizzy, line 1129: Why `_` here? I think you want `_` as the prefix of the function name and the parameters to not have underscores.

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +856,322 @@ def optional_config(self) -> Dict[str, Any]:
...
+        The last node is an identity to the previous one with the original name `function_name` to
+        represent an exit point of `post_pipe`.
+        """
+
+        original_node = node_.copy_with(name=f"{node_.name}_raw")
+
+        def __identity(foo: Any) -> Any:
+            return foo
+
+        transforms = self.transforms + (step(__identity).named(fn.__name__),)
```
elijahbenizzy, line 1085: This reads *very* cleanly -- nice!

elijahbenizzy, line 1085: Edge-case -- if there are no transforms, maybe we shouldn't have identity? I guess the simplest thing would just be to return the node in that case. This would happen if none of them resolve to true...

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +856,322 @@ def optional_config(self) -> Dict[str, Any]:
...
+
+        nodes, _ = chain_nodes(
```
elijahbenizzy, line 1087: I would call this `chain_steps` or `chain_transforms` actually -- feels a little cleaner

elijahbenizzy (commented):
Looking almost there, nice work!

Re: documentation -- i think it's clear, but more illustrations will help. the core concepts are simple:
1. `pipe` adds a set of transformations that feed into a parameter of a node
2. `@post_pipe` adds a set of transformations that alter the output of a node

This is all done in an immutable manner, so you can see the intermediate results

Also, idea re: naming:

- `@pipe_in`, `@pipe_out`
- `@pipe_into`, `@pipe_outof`
- `@transform_input`, `@transform_output`

Ideas?

hamilton/function_modifiers/macros.py
jernejfrank: I am leaning towards `@transform_input` and `@transform_output`, because they clarify what it does. Reading pipe sometimes can be confusing.. Maybe others have some strong opinions?

Also, I don't like `function_raw`, but I don't have a better name for it...

jernejfrank: Re docs:I took the nested part out because the procedural reads better to me and I don't think we need both.

jernejfrank:
Should we also add depracation warning onto pipe once we decide on the name?

hamilton/function_modifiers/macros.py
elijahbenizzy: So I prefer those too, but talked internally with @skrawcz and @zilto and they both prefer `@pipe_input` and `@pipe_output`. While it's a little less clear exactly what's happening, it matches some common patterns (SKLearn, pandas, polars, etc...), and thus is a a little easier for the people using it to adopt. I'm inclined to go with that -- it's fairly clear as to what's happening too IMO.

examples/scikit-learn/species_distribution_modeling/README.md
```diff
@@ -0,0 +1,43 @@
+# Species distribution modeling
```
elijahbenizzy, line 1: Cool example! Let's add a dag.png file as well?

examples/scikit-learn/species_distribution_modeling/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+We translate the Species distribution modeling from scikit-learn into Hamilton to showcase `pipe` and`post_pipe`.
+
+# Highlights
+ Example of a simple ETL pipeline broken into modules with external couplings.
+
+1) To see how to couple external modules / source code and integrate it into a Hamilton DAG check out `grids.py` or `preprocessing.py`, where we use `@pipe` to wrap and inject external functions as Hamilton nodes.
```
elijahbenizzy, line 8: Example is really good -- I'd highlight a litlte more use of pipe in the readme (E.G. why we're using it + a small example).

ALso there are some single-step pipe calls -- might be nice to see if they should be double ones? 

Furthermore, the pipe calls end up showing up in the graph... I think starting them with underscores should fix that?

elijahbenizzy (commented):
Looking great, basically ready! Some things we should change prior to release:
1. `@pipe_outputs` + `@pipe_input`
2. Deprecate `@pipe` (just make it subclass `@pipe_input`, and move the logic up)
3. A bit of touch-up to the example (looks really good)

We can release next week!

examples/scikit-learn/species_distribution_modeling/README.md
jernejfrank: Re@ renaming: just to double check it should be `pipe_outputs` and not `pipe_output`? One is singular the other plural then

jernejfrank: Re hiding node: I added the underscore but the nodes didn't disappear for me which is strange, since they did before.. Need to see if there's a problem for me only or something else going haywire.

elijahbenizzy (approved):
Looks good! One quick change -- let's add to this: https://github.com/DAGWorks-Inc/hamilton/blob/main/docs/reference/decorators/pipe.rst. 

Then I'll merge, and we can release this week!

elijahbenizzy (approved):

examples/scikit-learn/species_distribution_modeling/README.md
```diff
@@ -0,0 +1,93 @@
...
+def foo(bar:Any)->Any:
+    ...
+```
+In the above case a node gets created from `baz()` and is appended after `foo`; thus we get
+```python
+qux(baz(foo(bar)))
+```
+
+This can be particularly useful when:
+1. Ensuring consi
```
skrawcz, line 56: Need to push a fix here :)

examples/scikit-learn/species_distribution_modeling/README.md
```diff
@@ -0,0 +1,93 @@
...
+    plot_species_distribution()
+    plt.show()
+
+```
+so that it does not run when imported into other scripts.
+
+This script can be run as is for comparison. Actually the external functions `construct_grids()` and `create_species_bunch()` we will directly import and use as external functions to showcase how you can use our `pipe` and `pipe_output` functionality to "Hamiltonise" external modules.
+
+## hamilton_notebook.ipynb
+Is a nodebook that contains all the modules and has the execution cells to run the complete code. It also gives you the ability to visualize the DAG.
```
skrawcz, line 73: typo: nodebook -> notebook

hamilton/function_modifiers/macros.py
```diff
@@ -876,3 +881,177 @@ def optional_config(self) -> Dict[str, Any]:
...
+            step(B2)
+        )
+        def B(...):
+            return ...
+
+    we obtain the new DAG **A --> B_raw --> B1 --> B2 --> B --> C**, where we can think of the **B_raw --> B1 --> B2 --> B** as a "pipe" that takes the raw output of B, applies to it
+    B1, takes the output of B1 applies to it B2 and then gets renamed to B to re-connect to the rest of the DAG.
+
+    While it is generally reasonable to contain these constructs within a node's function,
+    you should consider `pipe_output` for similar reasons as `pipe`, namely, for any of the following reasons:
```
skrawcz, line 908: `pipe` -> `pipe_input`

examples/scikit-learn/species_distribution_modeling/README.md
elijahbenizzy: addressed in another PR

Issue #1148, opened 2024-09-20
Error when trying to use data adapter that uses `dataclasses.field(default_factory=...)`


Riezebos:
When a DataSaver class has a field with a default factory, an error occurs when trying to use it in a dag.

# Current behavior
I get the following error:

```
hamilton.function_modifiers.base.InvalidDecoratorException: Missing required parameters for adapter : <class 'pyvolt.hamilton_adapters.BigQuerySaver'>: {'settings'}. Required parameters/types are: {'table_name': <class 'str'>, 'settings': <class 'pyvolt.config_utils.Settings'>}. Optional parameters/types are: {'dataset_id': <class 'str'>, 'location': <class 'str'>, '_con': <class 'ibis.backends.BaseBackend'>}. 
```

## Steps to replicate behavior
I created a DataSaver that saves ibis tables to bigquery, is uses configuration from a Settings class that I created. If I use the class by itself there are no issues.

```python
@dataclass
class BigQuerySaver(DataSaver):
    table_name: str
    dataset_id: str = "data_marts"
    location: str = "europe-west1"
    settings: Settings = field(default_factory=Settings)

    _con: ibis.BaseBackend = None

    def __post_init__(self):
        self._con = ibis.bigquery.connect(
            project_id=self.settings.gcp_project,
            dataset_id=self.dataset_id,
            location=self.location,
        )

    def save_data(self, data: ibis.Table) -> dict[str, Any]:
        """Write Ibis table to BigQuery."""
        log.info("Writing table %s to BigQuery.", self.table_name)

        self._con.create_table(self.table_name, data.to_pyarrow(), overwrite=True)

    @classmethod
    def applicable_types(cls) -> Collection[Type]:
        return [ibis.Table]

    @classmethod
    def name(cls) -> str:
        return "bigquery"

BigQueryDataSaver(table_name="test_table")
```

The class instantiation works fine:
```
BigQuerySaver(table_name='test_table', dataset_id='data_marts', location='europe-west1', settings=Settings(gcp_project='my_project'), _con=<ibis.backends.bigquery.Backend object at 0x11501a930>)
```

But when I call e.g. `dr.display_all_functions()`, I get the error shown at the top.

It seems like something checks whether the decorator has all required args, but that something doesn't recognize that this argument is optional.

## Library & System Information
sf-hamilton==1.77.0

Python 3.12

# Expected behavior
No error

zilto:
[...]

PR #1149, opened 2024-09-20
added check for default_factory

zilto:
Related to #1148 

The current check uses `cls.get_required_arguments()` but this misses cases where the `dataclass` has an attribute with `field(default_factory=...)`. 

Now, the validation adds a step to inspect the dataclass's fields and check if a `default_factory` is set or if it's set to the sentinel value `dataclass.MISSING`.

Tests:
- added two tests with mock `DataLoader` and `DataSaver` ensuring that the `Driver` can now be properly built.
- Did a repro of the issue mentioned and now the `Driver` can be built and executed successfully

### DIFF
--- a/hamilton/function_modifiers/adapters.py
+++ b/hamilton/function_modifiers/adapters.py
@@ -1,3 +1,4 @@
+import dataclasses
 import inspect
 import logging
 import typing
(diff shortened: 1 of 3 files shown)


hamilton/function_modifiers/adapters.py
```diff
@@ -49,9 +50,18 @@ def validate(self):
 
         :raises InvalidDecoratorException: If the arguments are invalid.
         """
+        # get_required_arguments won't catch dataclasses.field(default_factory=...)
         required_args = self.adapter_cls.get_required_arguments()
         optional_args = self.adapter_cls.get_optional_arguments()
         missing_params = set(required_args.keys()) - set(self.kwargs.keys())
+
+        SENTINEL = object()
+        fields = getattr(self.adapter_cls, "__dataclass_fields__", SENTINEL)
```
elijahbenizzy, line 59: You want `hasattr`, not `SENTINEL` -- that pattern isn't really needed here IMO

hamilton/function_modifiers/adapters.py
```diff
@@ -49,9 +50,18 @@ def validate(self):
 
         :raises InvalidDecoratorException: If the arguments are invalid.
         """
+        # get_required_arguments won't catch dataclasses.field(default_factory=...)
```
elijahbenizzy, line 53: Why not just change it to? Makes more sense here: https://github.com/DAGWorks-Inc/hamilton/blob/72866175bb1e5dc65cedfd2de1096cc082fa19ee/hamilton/io/data_adapters.py#L64.

elijahbenizzy (commented):
Can we move this upstream? If not I think there's a bit of a code smell somewhere else. 

E.G. it's not the job of the adapters to correct for the default argument, it's the job of the validator. The adapter just passes in the arguments. Assuming the call-pattern is correct (It can be instantiated without them), then I think that makes sense.

elijahbenizzy (approved):
Looks good, just a comment to remove

Issue #1168, opened 2024-10-06
`@ignore` decorator


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
Often times you want helper functions that don't start with `_`. This could be used internally by tools such as mutate/pipe_output/
anything that relies on helper functions.

**Describe the solution you'd like**

```python
@ignore
def helper(...) -> ...:
    """This will not be part of the DAG"""
    ...
```

**Describe alternatives you've considered**
Forcing `_`, or just makeing this internal

**Additional context**
Would be useful for #1180. Should be easy to implement.

zilto:
Closing since # 1172 was merged

PR #1172, opened 2024-10-08
@ignore

jernejfrank:
Solves #1168.

## Changes
- implemented ignore decorator
- used it internally in `mutate` to automatically hide helper functions --> resolved a TODO
- exposed it for general use

## How I tested this
- unit test
- on the abstract mutate example (for both `step` and `mutate`)

## Checklist


examples/mutate/abstract functionality blueprint/pipe_output.py
```diff
@@ -6,43 +6,43 @@
...
 # data1 and data2
-def _filter(some_data: pd.DataFrame) -> pd.DataFrame:
+def filter_(some_data: pd.DataFrame) -> pd.DataFrame:
     return some_data.dropna()
 
 
 # data 2
 # this is for value
-def _add_missing_value(some_data: pd.DataFrame, missing_row: List[Any]) -> pd.DataFrame:
+def add_missing_value_(some_data: pd.DataFrame, missing_row: List[Any]) -> pd.DataFrame:
```
elijahbenizzy, line 15: We should only have `_` at the end of of fns if they're builtins (E.G. `filter`) -- otherwise just normal names

hamilton/function_modifiers/__init__.py
```diff
@@ -29,6 +29,7 @@
 
 # The config decorator
 config = configuration.config
+ignore = configuration.ignore.ignore()
```
elijahbenizzy, line 32: I think you can skip one level of `ignore` here. E.G. just instantiate the class... It carries no state so it should be good

examples/mutate/abstract functionality blueprint/mutate.py
```diff
@@ -24,7 +24,7 @@ def data_3() -> pd.DataFrame:
 
 # data1 and data2
 @mutate(data_1, data_2)
-def _filter(some_data: pd.DataFrame) -> pd.DataFrame:
```
skrawcz, line 27: so the way this works, is that hamilton's find functions will create a node for this function, but with the decorator we inject `ignore` so that when we come to resolve nodes, they resolve to None.

Is that right?

elijahbenizzy (commented):
Looks good! Small nits. One thought -- is `@ignore` (the name I suggested) the best name?

Other thoughts (would love your ideas):
- `@hamilton_ignore`
 - `@exclude`
 - `@hamilton_exclude`


And another thought -- should we have an option in `@mutate` and `step` to not ignore them? E.G. is there any way we may want to also include it in the pipeline (instinct says no here...)

examples/mutate/abstract functionality blueprint/mutate.py
elijahbenizzy: @skrawcz nope, this is a `NodeResolver` so it will short-circuit, and not create a node for the function. Effectively Hamilton won't see a function when it attempts to parsse (and thus it's not bound to the same rules).

examples/mutate/abstract functionality blueprint/pipe_output_on_output.py
```diff
@@ -3,15 +3,15 @@
...
 
 
-def _post_step(something: int) -> int:
+def post_step(something: int) -> int:
     return something + 100
 
 
-def _something_else(something: int) -> int:
+def something_else(something: int) -> int:
     return something + 1000
```
skrawcz, line 15: so my concern here is that people can now mix functions in a module that aren't hamilton nodes; i.e. you don't know what you're looking at...

hamilton/function_modifiers/macros.py
```diff
@@ -620,6 +620,8 @@ def step(
     they will be converted to a value (a literal)
     :return: an applicable with the function applied
     """
+    # This function will be excluded from the DAG as a node since we are inserting it manually
+    fn = ignore()(fn)
```
elijahbenizzy, line 624: So, thinking about this, I'm not sure we should be injecting this with `step` -- to me this might be a parameter we want to `pipe_output`. 

The problem is we won't know whether a function is included until later (when it's referred), which feels harder to parse. So maybe remove this, or put this behind a default-off toggle?

examples/mutate/abstract functionality blueprint/pipe_output_on_output.py
elijahbenizzy: @skrawcz yep, agreed, see this comment (based on yours): https://github.com/DAGWorks-Inc/hamilton/pull/1172#discussion_r1792881453

examples/mutate/abstract functionality blueprint/mutate.py
```diff
@@ -35,7 +35,7 @@ def _filter(some_data: pd.DataFrame) -> pd.DataFrame:
 # data 2
 # this is for value
 @mutate(data_2, missing_row=value(["c", 145]))
-def _add_missing_value(some_data: pd.DataFrame, missing_row: List[Any]) -> pd.DataFrame:
```
skrawcz, line 38: I think this is okay to not have `_` from a API UX consistency view, only because we decorate it with a Hamilton thing -- so you will know that this turns into part of DAG for a node.

examples/mutate/abstract functionality blueprint/pipe_output_on_output.py
skrawcz: so my thought is that we don't enable `@ignore` here.

jernejfrank:
> Looks good! Small nits. One thought -- is `@ignore` (the name I suggested) the best name?
> 
> Other thoughts (would love your ideas):
> 
> * `@hamilton_ignore`
> * `@exclude`
> * `@hamilton_exclude`
> 
> And another thought -- should we have an option in `@mutate` and `step` to not ignore them? E.G. is there any way we may want to also include it in the pipeline (instinct says no here...)

Having a look with fresh eyes I agree, we should make it explicit that it is for hamilton to avoid confusion. How about `hamilton_skip`?

examples/mutate/abstract functionality blueprint/mutate.py
jernejfrank: I also checked and the following will not raise an error of missing type hints, so this allows you to have non-hamilton functions without the "_" pre-fix:

```python
@ignore
def foo(a,b,c):
    return a+b+c
```

zilto:
> > Looks good! Small nits. One thought -- is `@ignore` (the name I suggested) the best name?
> > Other thoughts (would love your ideas):
> > 
> > * `@hamilton_ignore`
> > * `@exclude`
> > * `@hamilton_exclude`
> > 
> > And another thought -- should we have an option in `@mutate` and `step` to not ignore them? E.G. is there any way we may want to also include it in the pipeline (instinct says no here...)
> 
> Having a look with fresh eyes I agree, we should make it explicit that it is for hamilton to avoid confusion. How about `hamilton_skip`?

Thoughts on naming:
- we have `@cache(behavior="ignore")`
- we can think of caching as "skipping a node"
- As far as I understand, using `@ignore` is equivalent to prefixing the function with `_`. Instead of the decorator name indicating "what it does" (ignore, exclude), why not indicate "what it is". For consistency with the Hamilton documentation, I would call this `@helper` or `@utility` because you're indicating that this is a utility function.

elijahbenizzy:
I think it should be very specific that Hamilton is the one doing it -- so I'm in favor of `@hamilton_exclude`.

That said, I don't think this should be the first way to do things (it should be with `_`), but there are times it'll be useful (the docs here hit it right, it should show up in the docs, but not in the "How-to" section)

docs/reference/decorators/config_when.rst
```diff
@@ -18,6 +18,7 @@ Note the following:
 
   * ``@config`` If you're feeling adventurous, you can pass in a lambda function that takes in the entire configuration and resolves to ``True`` or ``False``. You probably don't want to do this.
 
+* To always exclude a function (such as helper functions) from the DAG you can also use ``@hamilton_skip``.
```
elijahbenizzy, line 21: Mention that we can use `_` as a prefix (preferred) or `@hamilton_skip`

elijahbenizzy:
Looks good, let's just change the wording of the docs to indicate that `_` is preferred!

elijahbenizzy (approved):
Looks good! Let's just change the wording to show that `_` is preferred

elijahbenizzy (approved):

Issue #1161, opened 2024-09-30
Add `target` parameter to `pipe_input`


elijahbenizzy:
**Is your feature request related to a problem? Please describe.**
Currently it assumes the first parameter. But that's a bit silly, we should use a target parameter. That would also allow pipe_input to work with other decorators cleanly.

**Describe the solution you'd like**

```python
@pipe_input(
    step(...),
    step(...),
    target="b" # applies to parameter `b`
)
def foo(a: int, b: int) -> int:
    ...
```

**Describe alternatives you've considered**
Not doing this or just reordering parameters for the user.

**Additional context**
Tangentially related to this: https://github.com/DAGWorks-Inc/hamilton/pull/1160

zilto:
Nitpick: should it be `target` or `target_`. Both are present in the codebase (`from_.parquet(..., target="raw_data")`, `@tag(..., target_="field_b")`)

PR #1173, opened 2024-10-09
pipe_input target

jernejfrank:
Addresses #1161.

Default behaviour reverts back to first parameter chaining for backwards compatibility.

## Changes
- Corrected minor bug for Applicable (passing through target when using other methods)
- Added user facing API on_input
- Added on_input to pipe_input for global setting

## How I tested this
- Unit tests

## Notes
There are two differences to `on_output`:

1. Since we do not use NodeTransformer we can have a mixture of global and local `on_input` settings. I made it so that the local (on the level of Applicable) adds to the global one (on the level of pipe_input). The other option would be to allow the local to override the global one. I also added a safeguard that if `on_input` is used, all transforms need to have a target to have it clear.
2. The namespace convention needs slight adjustment in case we apply pipe_input to multiple parameters. Since each node is chained through the parameter name, I have now that in case the same transform is applied to two parameters, we create two nodes with the same namespace but add to them "_param_name" so that we don't have problems with same-node names in the DAG.

## Checklist


hamilton/function_modifiers/macros.py
```diff
@@ -773,65 +809,152 @@ def final_result(upstream_int: int) -> int:
...
-        consumption/output later. Setting the namespace in individual nodes as well as in ``pipe`` is not yet supported.
+        consumption/output later. Setting the namespace in individual nodes as well as in ``pipe_input`` is not yet supported.
+
     """
 
     def __init__(
         self,
         *transforms: Applicable,
         namespace: NamespaceType = ...,
+        on_input: base.TargetType = None,
```
elijahbenizzy, line 862: So I'm not sure that `on_input` makes sense for anything but a single string, and the semantics aren't quite the same. Some thoughts:
1. By default, this applies to the first parameter
2. Does `...` apply to everything?

IMO you should restrict it more -- either a string, or default (which is the first parameter). Also, `@load_from` (which is similar) has the parameter named `target`. But that's confusing, and probably not the best (maybe it should be called `on_input` there?. Thoughts on naming?

hamilton/function_modifiers/macros.py
```diff
@@ -773,65 +809,152 @@ def final_result(upstream_int: int) -> int:
...
+
         self.transforms = transforms
         self.collapse = collapse
+        self.target = on_input
         self.chain = _chain
         self.namespace = namespace
 
+        if isinstance(on_input, str):  # have to do extra since strings are collections in python
+            self.target = [on_input]
+        elif isinstance(on_input, Collection):
```
elijahbenizzy, line 883: What's the use-case for a collection here? It's interesting, but feels a bit more complex than necessary.

hamilton/function_modifiers/macros.py
```diff
@@ -773,65 +809,152 @@ def final_result(upstream_int: int) -> int:
...
-            first_arg=first_parameter,
-            transforms=self.transforms,
-            namespace=self.namespace,
-            config=config,
-            fn=fn,
-        )
+        total_nodes = []
+        total_rename_maps = {}
+
+        for param in parameters_transforms_map:
```
elijahbenizzy, line 980: This logic is getting complex -- we should probably break it out into a separate function and add a unit test or two for it...

elijahbenizzy (commented):
Nice! Left some thoughts -- I think this is valuable (E.G. the first parameter is a bit of an awkward assumption), but we might want to restrict the API a bit. Also some naming/consistency stuff.

hamilton/function_modifiers/macros.py
jernejfrank: > So I'm not sure that `on_input` makes sense for anything but a single string, and the semantics aren't quite the same. Some thoughts:
> 
> 1. By default, this applies to the first parameter

Yes, I left the default to chain the first parameter and keep the namespace as is for backwards compatibility 👍 .

> 2. Does `...` apply to everything?

I think we can add `...`, but you could also just use the global `on_input` and pass in a list with the parameters.

> 
> IMO you should restrict it more -- either a string, or default (which is the first parameter). Also, `@load_from` (which is similar) has the parameter named `target`. But that's confusing, and probably not the best (maybe it should be called `on_input` there?. Thoughts on naming?

Ah okay, so the idea I had in mind is to leave flexibility that you can run `pipe_input` on more than one parameter: 3 different cases specifically

1. You can set a global `on_input` and it will apply it to those input arguments, like if we want to run it on the second argument *w-->_add_one --> _add_two --> 2nd arg of foo*
```python
def _add_one(v: int) -> int:
    return v + 1

def _add_two(v: int) -> int:
    return v + 2

@pipe_input(
    step(_add_one).named("a"), 
    step(_add_two).named("b"),
    on_input="w",
    namespace="global",
)
def fool(v: int, w: int) -> int:
    return w
```

2. You can set a local `on_input` and choose which transform runs on which input argument, like  *v-->_add_one --> 1st arg of foo* and *w-->_add_two --> 2nd arg of foo*. 
```python
def _add_one(v: int) -> int:
    return v + 1

def _add_two(v: int) -> int:
    return v + 2

@pipe_input(
    step(_add_one).on_input("v").named("a"), 
    step(_add_two).on_input("w").named("b"),
    namespace="local",
)
def fool(v: int, w: int) -> int:
    return v+w
```

3. You can set have a mixture `on_input` for global and local. The local adds specific transforms on top of what the global has. Here the whole pipeline gets applied to `v`, but only *_add_two* to `w`
```python
def _add_one(v: int) -> int:
    return v + 1

def _add_two(v: int) -> int:
    return v + 2

@pipe_input(
    step(_add_one).named("a"), 
    step(_add_two).on_input("w").named("b"),
    on_input="v"
    namespace="local",
)
def fool(v: int, w: int) -> int:
    return v+w
```

4. The `None` defaults to first parameter and if someone sets `on_input` but not all the `steps` in `pipe_input` get hit, we error out that is unclear what the user wants and he either needs to specify locally or globally or omit.

jernejfrank:
> Nice! Left some thoughts -- I think this is valuable (E.G. the first parameter is a bit of an awkward assumption), but we might want to restrict the API a bit. Also some naming/consistency stuff.

The API expanded substantially to leave as much flexibility in there as possible. It also is a complete dual to `pipe_output` in the sense that you can target multiple parameters (like we do end nodes) and can allocate transforms individually between them -- so you could use them interchangeably / becomes a personal preference if you rather chain on output nodes or input parameter.

But I'm happy to constrain it more, like @elijahbenizzy suggested, to only allow only a single parameter and you pass it in as a string seems reasonable. In case you want more flexibility use `pipe_output`. There would be one case I don't see this working out, but that is probably then more indicative of a code smell than a lack of flexibility:

```python

def _transform():
    ...

def _other_transform():
    ...

@pipe_output(step(_transform)) # this won't work
def interesting_node():
    ...

@pipe_input(
        step(_transform).on_input="interesting_node", # this should work
        step(_other_transform).on_input("foo") # leaving flexibility to also transform other params
) 
def downstream(foo, bar, baz, interesting_node,...):
    ...

def some_other_dependency(interesting_node):
# here you don't want to have the pipe transform happening just the original `interesting_node` input
    ...
```

jernejfrank:
> > Nice! Left some thoughts -- I think this is valuable (E.G. the first parameter is a bit of an awkward assumption), but we might want to restrict the API a bit. Also some naming/consistency stuff.
> 
> The API expanded substantially to leave as much flexibility in there as possible. It also is a complete dual to `pipe_output` in the sense that you can target multiple parameters (like we do end nodes) and can allocate transforms individually between them -- so you could use them interchangeably / becomes a personal preference if you rather chain on output nodes or input parameter.
> 
> But I'm happy to constrain it more, like @elijahbenizzy suggested, to only allow only a single parameter and you pass it in as a string seems reasonable. In case you want more flexibility use `pipe_output`. There would be one case I don't see this working out, but that is probably then more indicative of a code smell than a lack of flexibility:
> 
> ```python
> def _transform():
>     ...
> 
> def _other_transform():
>     ...
> 
> @pipe_output(step(_transform)) # this won't work
> def interesting_node():
>     ...
> 
> @pipe_input(
>         step(_transform).on_input="interesting_node", # this should work
>         step(_other_transform).on_input("foo") # leaving flexibility to also transform other params
> ) 
> def downstream(foo, bar, baz, interesting_node,...):
>     ...
> 
> def some_other_dependency(interesting_node):
> # here you don't want to have the pipe transform happening just the original `interesting_node` input
>     ...
> ```

Maybe it would make sense to keep it like this because the subtle difference in the implementations between `pipe_input` and `pipe_output` can work hand in hand.

Specifically, for `pipe_output` we touch the original node by creating this chain `node_raw`--`chain_nodes`--`node`, whereby for `pipe_input` we have `param`--`chain_node`--`renamed_param_in_upstream_node` so we don't touch the original node, but  rename the argument in the next node.

hamilton/function_modifiers/macros.py
```diff
@@ -345,7 +345,7 @@ def __init__(
         :param _resolvers: Resolvers to use for the function
         :param _name: Name of the node to be created
         :param _namespace: Namespace of the node to be created -- currently only single-level namespaces are supported
-        :param _target: Selects which target nodes it will be appended onto. By default all.
+        :param _target: Selects which target nodes it will be appended onto. Default None gets resolved on decorator level.
```
elijahbenizzy, line 348: I would explain this more -- E.G. reference the decorators/how they resolve it.

hamilton/function_modifiers/macros.py
```diff
@@ -773,65 +810,234 @@ def final_result(upstream_int: int) -> int:
...
+            For extra control in case of multiple function arguments (parameters), we can also specify the target parameter that we wish to transform.
+            In case ``on_input`` is set to None (default), we apply ``pipe_input`` on the first parameter only. If ``on_input`` is set for a specific transform
+            make sure the other ones are also set either through a global setting or individually, otherwise it is unclear which transforms target which parameters.
+
+            The following applies *_add_one* to ``p1``, ``p3`` and *_add_two* to ``p2``
+
+            .. code-block:: python
+
+                @pipe_input(
+                    step(_add_one).on_input(["p1","p3"])
```
elijahbenizzy, line 840: Update docs -- we disabled this (for now), right?

hamilton/function_modifiers/macros.py
elijahbenizzy: Thanks @jernejfrank  -- this makes a lot of sense.

To sum up the offline conversation we had, we decided to restrict it to a single input (for now), knowing that we can turn it on for more. The reasoning is to reduce supported surface area until we get a request for this feature!

elijahbenizzy (approved):
Looks good! Some small docs suggestions. 

Also, let's create a ticket here. You've done so much for multiple already, we should track + link from the docs. Nice work!

hamilton/function_modifiers/macros.py
jernejfrank: Correct, I found a hack how to "comment out" things from `rst` so Sphinx doesn't generate this part of the docstring.

PR #1188, opened 2024-10-16
fix: added polars to extension list

zilto:
User reported an issue where using `pandera` with `polars` raised the error. As identified by the user, this is because `polars` wasn't added to the list of supported extension.

> Actual error: No registered subclass of BaseDefaultValidator is available for arg: schema and type <class 'polars.dataframe.frame.DataFrame'>. This either means (a) this arg-type contribution isn't supported or (b) this has not been added yet (but should be). In the case of (b), we welcome contributions. Get started at [github.com/dagworks-inc/hamilton](http://github.com/dagworks-inc/hamilton).

### DIFF
--- a/hamilton/data_quality/pandera_validators.py
+++ b/hamilton/data_quality/pandera_validators.py
@@ -6,7 +6,7 @@ from hamilton import registry
 from hamilton.data_quality import base
 from hamilton.htypes import custom_subclass_check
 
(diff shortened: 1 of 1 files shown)


elijahbenizzy (commented):
Looks good!

elijahbenizzy (approved):

PR #1198, opened 2024-10-21
FIX pipe_output / mutate naming convention and docstrings

jernejfrank:
The naming convention for the first node in the `pipe_output` chain can potentially conflict with user-defined nodes. For example:

```python

@pipe_output(...)
def features():
    ...
```

would lead to `features_raw` which is a common user naming convention. This makes it harder to have the conflict by following the same naming convention as the rest of the chained nodes.

## Changes
- first node is now named `foo.with_raw`
- corrected docstring in `mutate.__init__` to avoid escape key errors in pytest.

## How I tested this
- unit tests
- sphinx autoload

## Notes
The docstring issue is currently only tracked on slack: https://hamilton-opensource.slack.com/archives/C03M34FM058/p1729431573025409

## Checklist

### DIFF
--- a/hamilton/function_modifiers/macros.py
+++ b/hamilton/function_modifiers/macros.py
@@ -1158,7 +1158,7 @@ class pipe_output(base.NodeTransformer):
         def B(...):
             return ...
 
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
Looks good! Thank you.

Issue #1200, opened 2024-10-23
graph visualization fails because of unescaped config values


MG-MW:
Depending on given configuration, graph visualization fails

# Current behavior

```python
import hamilton.driver
import hamilton.ad_hoc_utils


def world(hello: str) -> str:
    return f"{hello} world"


temp_module = hamilton.ad_hoc_utils.create_temporary_module(world)


driver = hamilton.driver.Builder().with_modules(temp_module).with_config({"hello": "Hi"}).build()
# does not fail
driver.display_upstream_of("world")

driver = hamilton.driver.Builder().with_modules(temp_module).with_config({"hello": "<"}).build()
# throws error
driver.display_upstream_of("world")
```
## Stack Traces
<details>
<summary>Stack trace is not really helpful</summary>
<code>
---------------------------------------------------------------------------
CalledProcessError                        Traceback (most recent call last)
File ~/.envs/py312test/lib/python3.12/site-packages/graphviz/backend/execute.py:88, in run_check(cmd, input_lines, encoding, quiet, **kwargs)
     87 try:
---> 88     proc.check_returncode()
     89 except subprocess.CalledProcessError as e:

File ~/.local/share/mise/installs/python/3.12.3/lib/python3.12/subprocess.py:502, in CompletedProcess.check_returncode(self)
    501 if self.returncode:
--> 502     raise CalledProcessError(self.returncode, self.args, self.stdout,
    503                              self.stderr)

CalledProcessError: Command '[PosixPath('dot'), '-Kdot', '-Tsvg']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

CalledProcessError                        Traceback (most recent call last)
File ~/.envs/py312test/lib/python3.12/site-packages/IPython/core/formatters.py:977, in MimeBundleFormatter.__call__(self, obj, include, exclude)
    974     method = get_real_method(obj, self.print_method)
    976     if method is not None:
--> 977         return method(include=include, exclude=exclude)
    978     return None
    979 else:

File ~/.envs/py312test/lib/python3.12/site-packages/graphviz/jupyter_integration.py:98, in JupyterIntegration._repr_mimebundle_(self, include, exclude, **_)
     96 include = set(include) if include is not None else {self._jupyter_mimetype}
     97 include -= set(exclude or [])
---> 98 return {mimetype: getattr(self, method_name)()
     99         for mimetype, method_name in MIME_TYPES.items()
    100         if mimetype in include}

[...]

PR #1202, opened 2024-10-23
Fix `keep_dot` propagation in `Driver` display functions

ChronoJon:
When investigating #1200 I noticed, that `keep_dot` parameters weren't propagated to the render function. 

## Changes

The previous non-functional parameters are now passed on.

## Checklist

These are not applicable


skrawcz (approved):
thanks for the catch!

Issue #1204, opened 2024-10-24
Can only fetch first run id from cache using `SQLiteMetadataStore` object


elutins:
Cache only returns the first `run_id` that populated the cache when calling `get_run_ids()` on a `hamilton.caching.stores.sqlite.SQLiteMetadataStore` object

# Current behavior

- When calling `get_run_ids()` on a `hamilton.caching.stores.sqlite.SQLiteMetadataStore` class object, only the first `run_id` that populated the cache is returned rather than all `run_ids` of the cache. 

## Stack Traces

Believe this is caused by [this line here](https://github.com/DAGWorks-Inc/hamilton/blob/09f07129f2f7e6c06ec61914ca97e8df3dd08ed5/hamilton/caching/stores/sqlite.py#L180) in `hamilton/caching/stores/sqlite.py` in which the return `result` object is being indexed to the first item only

## Steps to replicate behavior
1. instantiate a driver object using `with_cache()` 
```
driver = (
    hamilton.driver.Builder()
    .with_config()
    .with_modules({some_module})
    .with_cache()
    .build()
)
```
2. execute the driver `driver.execute(final_vars=[some_var], inputs={some_input})`
3. execute the driver again grabbing a different `final_var`:  `driver.execute(final_vars=[some_other_var], inputs={some_input})`
4. run ` hamilton.caching.stores.sqlite.SQLiteMetadataStore(".hamilton_cache").get_run_ids()` to get the run_ids that populated cache
    - the expected/wanted behavior is that this command would return _all_ run_ids - not just the first 

## Library & System Information
- Using hamilton version `1.80.0` & python `3.12.7`

# Expected behavior

- When calling `get_run_ids()` on a `hamilton.caching.stores.sqlite.SQLiteMetadataStore` class object, I'd expect a full list of run ids that have populated the cache to be returned rather than just the first run.

# Additional context

skrawcz:
@elutins thanks for flagging. We'll get a fix out today.

skrawcz:
@elutins this has been fixed in `sf-hamilton==1.81.2`.  Thanks for raising!

PR #1205, opened 2024-10-24
fix: caching `SQLiteMetadataStore.get_run_ids()`

zilto:
This PR follows issue #1204 

It includes a few changes to the `MetadataStore` base class and sets a clearer contract for `.get_run()` and `.get_run_ids()`.

- `.get_run_ids()` should return a list of runs sorted by their start time, i.e., when `.initialize()` is called. This order may differ from the order of the first recorded node execution. An empty list will be returned if no run were initialized.

- `.get_run()` should return a list of dictionaries, where each `dict` is associated with a particular node execution. Each `dict` must minimally return the `data_version` and the `cache_key` (which can be decoded to retrieve the `node_name`, `code_version`,  and `dependencies_data_versions`), but may include more information. An `IndexError` is raised if `run_id` doesn't exist. An empty list is returned if the run was initialized, but recorded no node execution.

Tests were added to catch the bug reported in the original issue.

### DIFF
--- a/hamilton/caching/stores/base.py
+++ b/hamilton/caching/stores/base.py
@@ -95,7 +95,7 @@ class MetadataStore(abc.ABC):
         """
 
     @abc.abstractmethod
-    def get(self, cache_key: str) -> Optional[str]:
+    def get(self, cache_key: str, **kwargs) -> Optional[str]:
         """Try to retrieve ``data_version`` keyed by ``cache_key``.
         If retrieval misses return ``None``.
         """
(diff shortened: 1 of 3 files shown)


elijahbenizzy (approved):
Looks good, unclear about a few of the other API changes (`**kwargs`), but I think we should ship it! Also docs

hamilton/caching/stores/base.py
```diff
@@ -118,15 +118,19 @@ def exists(self, cache_key: str) -> bool:
...
     @abc.abstractmethod
-    def get_run(self, run_id: str) -> Any:
-        """Return all the metadata associated with a run.
-        The metadata content may differ across MetadataStore implementations
+    def get_run(self, run_id: str) -> Sequence[dict]:
+        """Return a list of node metadata associated with a run.
+        For each node, the metadata should include:
+            - ``cache_key`` (created or used)
+            - ``data_version``
+        This is to allow users to manually query the MetadataStore or ResultStore.
```
skrawcz, line 129: this is what the docs would be complaining about.

hamilton/caching/stores/sqlite.py
```diff
@@ -106,13 +107,21 @@ def set(
         self,
         *,
         cache_key: str,
-        node_name: str,
-        code_version: str,
         data_version: str,
         run_id: str,
+        node_name: str = None,
+        code_version: str = None,
```
skrawcz, line 113: this could be backwards incompatible. But if we're the only ones using this, then all good.

zilto: It's not because the `HamiltonCacheAdapter` is the only object passing values here. As shown in the updated logic of `.get()` it will resolve the values of `node_name` and `code_version` from the `cache_key` so these values will never be `None` (for instance, the SQL table still enforce `NOT NULL` constraints)

For instance, the `MetadataStore` base class doesn't expose `node_name` and `code_version` as arguments because other implementations might no provide them directly. This information is redundant with the content of `cache_key`. 

The `SQLiteMetadataStore` implementation uses `node_name` and `code_version` explicitly to allow indexing and cache management (e.g., delete all instance of this node). These values could be derived by decoding `cache_key` but that would be a wasteful operation

skrawcz: it's more that if people were not using keyword args it would be break due to the change in argument order.

skrawcz: (I didn't see any update to callers of this function hence why I asked)

PR #1206, opened 2024-10-24
Bumps hamilton version from 1.81.1 to 1.81.2

elijahbenizzy:
Version bump


elijahbenizzy:
This technically has some new features but it's a bug-fix and experimental so I'm keeping it a patch version

PR #1207, opened 2024-10-25
feat: in-memory caching

zilto:
This feature was requested by user. The goal is to enable caching without persisting results to disk. This is useful when using interactive Python sessions, notebooks, or even webservices, and don't want to use disk space.

In-memory caching requires implementing both a metadata and a result store in memory. It was also the opportunity to introduce `.load_from()` and `.persist_to()` which allows to load and save stores via other implementations. 

sidenote: This feature could exist on any implementation actually (e.g., `S3ResultStore.load_from(FileResultStore(...))`). In the mean time, a similar pattern is currently possible through the in-memory store: 

```python
s3_result_store = S3ResultStore(...)
file_result_store = FileResultStore(...)

# this populates the S3 store from the local file-based store
temp_store = InMemoryResultStore.load_from(file_result_store, data_versions=[...])
temp_store.persist_to(s3_result_store)

# the Driver will use the S3 store, now populated with data that was local
dr = driver.Builder().with_modules(...).with_cache(result_store=s3_result_store).build()
```


hamilton/caching/stores/memory.py
```diff
@@ -0,0 +1,279 @@
...
+    def delete(self, cache_key: str) -> None:
+        """Delete the ``data_version`` for ``cache_key``."""
+        del self._data_versions[cache_key]
+
+    @override
+    def delete_all(self) -> None:
+        """Delete all stored metadata."""
+        self._data_versions.clear()
+
+    def persist_to(self, metadata_store: Optional[MetadataStore] = None) -> None:
```
elijahbenizzy, line 58: Using another store is a smart way to do this, nice

hamilton/caching/stores/memory.py
```diff
@@ -0,0 +1,279 @@
...
+                for node_metadata in metadata_store.get_run(run_id):
+                    data_versions_to_retrieve.add(node_metadata["data_version"])
+
+        for data_version in data_versions_to_retrieve:
+            # TODO disambiguate "result is None" from the sentinel value when `data_version`
+            # is not found in `result_store`.
+            result = result_store.get(data_version)
+            in_memory_result_store.set(data_version, result)
+
+        # TODO could be beneficial to delete the `result_store` after loading from it,
```
elijahbenizzy, line 277: `del` would just be a reference clear unless you wipe it, python does not have pointer capabilities in the normal use. Just document how to keep things clean IMO

hamilton/caching/stores/sqlite.py
```diff
@@ -118,7 +122,15 @@ def set(
         # if the caller of ``.set()`` directly provides the ``node_name`` and ``code_version``,
         # we can skip the decoding step.
         if (node_name is None) or (code_version is None):
-            decoded_key = decode_key(cache_key)
+            try:
+                decoded_key = decode_key(cache_key)
+            except BaseException as e:
+                raise ValueError(
+                    f"Failed decoding the cache_key: {cache_key}.\n",
+                    "Was it manually created? Do `code_version` and `data_version` found in ",
```
elijahbenizzy, line 130: What's the error case that this corresponds to? Is a user ever going to hit this? If so, point to docs on how to use it (it's a bit unclear from the error message)

tests/caching/metadata_store/test_base.py
```diff
@@ -0,0 +1,194 @@
...
+    metadata_store = _instantiate_metadata_store(metdata_store_cls, tmp_path)
+
+    yield metadata_store
+
+    metadata_store.delete_all()
+
+
+@pytest.mark.parametrize(
+    "metadata_store",
+    [SQLiteMetadataStore, InMemoryMetadataStore],
```
elijahbenizzy, line 54: Nice parameterizatio -- we should have a quick guide on "implementing your own" (noting it's a bit experimental...)

elijahbenizzy: Not a blocker

tests/caching/metadata_store/test_memory.py
```diff
@@ -0,0 +1,65 @@
...
+from hamilton.caching.stores.sqlite import SQLiteMetadataStore
+
+# `metadata_store` is imported but not directly used because it's
+# a pytest fixture automatically provided to tests
+from .test_base import _mock_cache_key, metadata_store  # noqa: F401
+
+
+@pytest.mark.parametrize(
+    "metadata_store",
+    [SQLiteMetadataStore],
```
elijahbenizzy, line 13: Why parameterize?

zilto: I was thinking that it would be relevant to test `.persist_to()` and `.load_from()` for all combinations `InMemory x Other`. This could be redundant with unit tests for `Other` implementation and integration testing though

tests/caching/metadata_store/test_base.py
zilto: The idea of `test_base.py` is that all implementations should be passing. Currently, all test have a parameterization with the list of implementations.

I think it would makes sense to have a constant at the top of the module `IMPLEMENTATIONS = [...]` containing the list of classes and using this list throughout.

hamilton/caching/stores/sqlite.py
zilto: This would happen if the user is meeting these conditions:
- using `SQLiteMetadataStore`
- passes a manually crafted key that isn't the result of `hamilton.caching.cache_key.create_key()` (i.e., it can't be parsed)
- they are not passing explicitly a `node_name` AND `code_version` so the metadata store needs to derive it.

e.g.,
```python
sqlite_metadata_store.set(
   cache_key="foo"   # not a properly formatted key
   data_version="any-string"  # this can be anything
)
```

This will never happen when using the adapter, only if the user is manually interacting with the metadata store.

## Why this odd behavior?
Metadata stores `.set()` only required kwargs `cache_key` and `data_version`. This provides limited information for cache management. Storing the `node_name` and `code_version` in the metadata store allows to delete entries by node name or version (not implemented yet). On the other, forcing this information to be available can add constraints

hamilton/caching/stores/memory.py
```diff
@@ -0,0 +1,279 @@
...
+    override = lambda x: x  # noqa E731
+
+from hamilton.caching.cache_key import decode_key
+
+from .base import MetadataStore, ResultStore, StoredResult
+from .file import FileResultStore
+from .sqlite import SQLiteMetadataStore
+
+
+class InMemoryMetadataStore(MetadataStore):
```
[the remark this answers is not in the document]
zilto: Following our internal discussion, this is not currently possible for `Parallelizable/Collect` scenarios. It would required the [sqlite shared-cache mode](https://www.sqlite.org/sharedcache.html). However, for `python < 3.11`, the standard library `sqlite3` is compiled with the constant `THREADSAFE=1` which makes this feature unusable (or requires the user to recompile SQLite) ([reference](https://ricardoanderegg.com/posts/python-sqlite-thread-safety/))

hamilton/caching/stores/sqlite.py
zilto: For later, I think it could be a `__init__()` kwarg for `MetadataStore` to toggle indexing `node_name` and `code_version`. This mostly affects SQL-based metadata store, but `node_name` indexing can also be possible in key-value stores via tags

hamilton/caching/stores/memory.py
elijahbenizzy: `del result_store` doesn't do much -- so yeah, remove it

tests/caching/metadata_store/test_memory.py
elijahbenizzy: OK, given that it's a parameterization of 1, add a comment

elijahbenizzy (approved):
Looks good, a few more small nits

hamilton/caching/stores/sqlite.py
elijahbenizzy: Yep, so add comments/more to the error message to clarify

zilto: yup, I improved it in the latest code version

Issue #1158, opened 2024-09-25
Generic @with_columns dataframe support


skrawcz:
**Is your feature request related to a problem? Please describe.**
We have @with_columns for pyspark, but not for other dataframes.

**Describe the solution you'd like**
`@with_columns` works for any dataframe type.

**Describe alternatives you've considered**
N/A

**Additional context**
This would help with reusing transforms across dataframes without getting into naming issues.

zilto:
If we implement `@with_columns` for Polars, we can leverage [Narwhals](https://github.com/narwhals-dev/narwhals) to support cuDF, Modin, pandas, Polars, Pyarrow, Dask, Ibis, Duckdb, Vaex, and more. We have to distinguish eager and lazy execution though.

Narwhals is an ok dependency to add because `@with_columns` is already an extension feature and Narwhals itself only depends on the standard library.

zilto:
several libraries are now supported

PR #1209, opened 2024-10-30
with_columns for Pandas

jernejfrank:
Creating the option to use `with_columns` on Pandas data frames.

Partially addressing #1158.

## Changes
- added new decorator

## How I tested this
- unit tests
- e2e
- example

## Notes
Known issues:
- config.when does not work for individual nodes within the `with_columns` subtag.

The other thing is, at the moment, this extracts relevant columns from the df and then appends them. This would then be an eager execution.

Polars and PySpark can do `with_columns` lazy and use their optimizers under the hood. Pandas to my knowledge doesn't have that, but there are two ways how we could mimic laziness:

1. using `eval`
```python
df.eval('new_column = f(previous_columns)'))
```
2. using`assign`
```python
df.assign(new_column = lambda x: x(previous_columns))
```
My reasoning so far is, that anybody who is concerned with latency will not be using Pandas in the first place so this seems like an overkill.

## Checklist


tests/function_modifiers/test_recursive.py
```diff
@@ -3,22 +3,25 @@
 import random
 from typing import Tuple
 
+import pandas as pd
 import pytest
 
-from hamilton import ad_hoc_utils, graph
+from hamilton import ad_hoc_utils, driver, graph, node
```
zilto, line 9: Could use

```python
from hamilton import ..., node as hamilton_node
```

because the name `node` is overriden elsewhere (e.g., line 187)

tests/resources/with_columns.py
```diff
@@ -0,0 +1,39 @@
+import pandas as pd
+
+from hamilton.function_modifiers import with_columns
+
+
+def initial_df() -> pd.DataFrame:
+    return pd.DataFrame({"col_1": [1, 2, 3, 4], "col_2": [11, 12, 13, 14], "col_3": [1, 1, 1, 1]})
+
+
+def substract_1_from_2(col_1: pd.Series, col_2: pd.Series) -> pd.Series:
```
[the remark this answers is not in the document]
jernejfrank: Learned something new today: 
[Substract was formerly used in analogy with abstract. But in modern usage, it is written according to the Latin, subtract.](https://english.stackexchange.com/questions/3640/is-substract-versus-subtract-a-proper-word)

hamilton/function_modifiers/__init__.py
```diff
@@ -88,6 +88,7 @@
 
 subdag = recursive.subdag
 parameterized_subdag = recursive.parameterized_subdag
+with_columns = recursive.with_columns
```
elijahbenizzy, line 91: This should probably live in a pandas extension -- while pandas has, historically, been the dependency for Hamilton, we want to move it out (and this is very pandas-specific logic)

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
         :return: Any required config items.
         """
         return None
+
+
+# TODO: Copied here from h_spark, needs refactoring
+def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> List[node.Node]:
```
elijahbenizzy, line 609: Yep, would make this live here and have pandas/spark refer to it

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+                # process, or just return unprocessed
+                ...
+
+    the above would output a dataframe where the two columns ``a_from_df`` and ``b_from_df`` get
+    overwritten.
+    """
+
+    @staticmethod
+    def _check_for_duplicates(nodes_: List[node.Node]) -> bool:
+        "Ensures that we don't run into name clashing of columns and group operations."
```
elijahbenizzy, line 715: Might be worth expanding in the docstrings a bit more about what this means -- I'm getting a little lost int he details (E.G. is this a user error? What's the remediation?) More for internal docs.

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+            if node_counter[node_.name] > 1:
+                return True
+        return False
+
+    def __init__(
+        self,
+        *load_from: Union[Callable, ModuleType],
+        columns_to_pass: Union[str, List[str]] = None,
+        pass_dataframe_as: str = None,
+        select: Union[str, List[str]] = None,
```
elijahbenizzy, line 728: So I think we should keep this consistenct with the `h_spark` `with_columns` -- maybe we add a single string ability for that? Or we remove it for this?

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+            upstream inputs should be taken from the dataframe, and which shouldn't. Note that, if this is
+            left empty (and external_inputs is as well), we will assume that all dependencies come
+            from the dataframe. This cannot be used in conjunction with pass_dataframe_as.
+        :param pass_dataframe_as: The name of the dataframe that we're modifying, as known to the subdag.
+            If you pass this in, you are responsible for extracting columns out. If not provided, you have
+            to pass columns_to_pass in, and we will extract the columns out for you.
+        :param namespace: The namespace of the nodes, so they don't clash with the global namespace
+            and so this can be reused. If its left out, there will be no namespace (in which case you'll want
+            to be careful about repeating it/reusing the nodes in other parts of the DAG.)
+        :param config_required: the list of config keys that are required to resolve any functions. Pass in None\
```
elijahbenizzy, line 744: This *should* solve the problem of config wiring through...

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+        if isinstance(columns_to_pass, str):
+            self.initial_schema = [columns_to_pass]
+        else:
+            self.initial_schema = columns_to_pass
+
+        self.dataframe_subdag_param = pass_dataframe_as
+        self.namespace = namespace
+        # This never gets used within the class, but pyspark had it so keeping it here in case we
+        # need to access it from somewhere outside
+        self.upstream_dependency = pd.DataFrame
```
elijahbenizzy, line 784: Let's remote this for now? Should also be able to remove from h_spark -- it's an artifact and was left in erroneously.

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+        self, inject_parameter: str, params: Dict[str, Type[Type]]
+    ) -> List[node.Node]:
+        """Helper function to extract columns for the user."""
+        output_type = params[inject_parameter]
+        series_type = registry.get_column_type_from_df_type(output_type)
+
+        out = []
+        for column in self.initial_schema:
+
+            def extractor_fn(
```
elijahbenizzy, line 798: Ooh we should be able to share logic with `extract_columns`... Might be good to leave this as a TODO note

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+        self, fn: Callable, params: Dict[str, Type[Type]]
+    ) -> Tuple[str, Collection[node.Node]]:
+        """Selects the correct dataframe and optionally extracts out columns."""
+        initial_nodes = []
+        if self.dataframe_subdag_param is not None:
+            inject_parameter = self.dataframe_subdag_param
+        else:
+            # If we don't have a specified dataframe we assume it's the first argument
+            sig = inspect.signature(fn)
+            inject_parameter = list(sig.parameters.values())[0].name
```
elijahbenizzy, line 830: SHould be able to validate the type for the parameter as well?

hamilton/function_modifiers/recursive.py
```diff
@@ -600,3 +603,292 @@ def required_config(self) -> Optional[List[str]]:
...
+
+        return inject_parameter, initial_nodes
+
+    def _create_merge_node(self, upstream_node: str, node_name: str) -> node.Node:
+        "Node that adds to / overrides columns for the original dataframe based on selected output."
+
+        def new_callable(**kwargs) -> Any:
+            df = kwargs[upstream_node]
+            for column in self.select:
+                df[column] = kwargs[column]
```
elijahbenizzy, line 850: Hmm, we're mutating this... We should be careful as that breaks assumptions. 

The nice thing about pandas is that the dataframes shouldn't mess with memory if you create new ones, the series are what takes up the memory. So we should be able to continually modify it by creating new ones (I think...)

elijahbenizzy: Might be worth using `assign` or something like that... Looking here (you mentioned similar things in your comments on the code): https://stackoverflow.com/questions/72291290/how-to-create-new-column-dynamically-in-pandas-like-we-do-in-pyspark-withcolumn

tests/function_modifiers/test_recursive.py
```diff
@@ -539,3 +543,261 @@ def test_recursive_validate_config_inputs_happy(config, inputs):
...
+            "col_2": [11, 12, 13, 14],
+            "col_3": [1, 1, 1, 1],
+            "subtract_1_from_2": [10, 10, 10, 10],
+            "multiply_3_by_5": [5, 5, 5, 5],
+        }
+    )
+    pd.testing.assert_frame_equal(result, expected_df)
+
+
+def test_end_to_end_with_columns_pass_dataframe():
```
elijahbenizzy, line 786: Other tests you may want to consider
1. Inputs from outside the subdag
2. Inputs from external stuff
3. Config

elijahbenizzy (commented):
Looking good! Main comments about how to do the assign-style operations. Definitely on the right track!

hamilton/function_modifiers/recursive.py
zilto: Looking at the way [Narwhals implements](https://github.com/narwhals-dev/narwhals/blob/5c3db5b149cbf64b97de49bc4ad726f019f9f578/narwhals/_pandas_like/dataframe.py#L388) the Polars `.with_columns()` for other backends can also be informative.

jernejfrank: I looked a bit deeper:

1. `assign` does a copy of the data frame. (It should be shallow unless the flag is set to `copy="deep"`. From [this thread](https://stackoverflow.com/questions/48177914/why-use-pandas-assign-rather-than-simply-initialize-new-column) my impression is that assign was created for chaining operations together and leaving the original data frame intact.
2. Narwhals uses `concat(dfs, axis=1, **extra_kwargs)` under the hood and this does have a warning: "[concat()](https://pandas.pydata.org/docs/reference/api/pandas.concat.html#pandas.concat) makes a full copy of the data, and iteratively reusing [concat()](https://pandas.pydata.org/docs/reference/api/pandas.concat.html#pandas.concat) can create unnecessary copies. Collect all [DataFrame](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html#pandas.DataFrame) or [Series](https://pandas.pydata.org/docs/reference/api/pandas.Series.html#pandas.Series) objects in a list before using [concat()](https://pandas.pydata.org/docs/reference/api/pandas.concat.html#pandas.concat)." However, for our use case I guess we can do this only once so should also be ok. 


And tested it -- seems like `concat` is the way to go in terms of performance:

```
%timeit pd.concat([df:(10000, 5000), cols:2000], axis=1)
280 ms ± 22.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit df:(10000, 5000).assign(**cols:2000)
4.44 s ± 765 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%memit pd.concat([df:(10000, 5000), cols:2000], axis=1)
peak memory: 2107.36 MiB, increment: 0.20 MiB

%memit df:(10000, 5000).assign(**cols:2000)
peak memory: 2525.92 MiB, increment: 37.08 MiB
```

```python
%load_ext memory_profiler
import numpy as np
import pandas as pd
import warnings  
warnings.simplefilter(action='ignore', category=pd.errors.PerformanceWarning)

pd_size = 10000
df = pd.DataFrame(np.random.randint(0,pd_size,size=(pd_size, pd_size//2)))#, columns=list('ABCD'))

cols = []
for i in range(pd_size//10):
    cols.append(np.random.rand(pd_size))

concat_cols = [pd.Series(col) for col in cols]
assign_cols = {f"{pd_size+i+1}":col for i,col in enumerate(cols)}


print(f'\n%timeit pd.concat([df:{df.shape}, cols:{len(concat_cols)}], axis=1)')
temp_df = [df.copy()]
temp_df.extend(concat_cols)
%timeit l=pd.concat(temp_df, axis=1)

print(f'\n%timeit df:{df.shape}.assign(**cols:{len(assign_cols)})')
temp_df = df.copy()
%timeit temp_df.assign(**assign_cols)


print(f'\n%memit pd.concat([df:{df.shape}, cols:{len(concat_cols)}], axis=1)')
temp_df = [df.copy()]
temp_df.extend(concat_cols)
%memit l=pd.concat(temp_df, axis=1)

print(f'\n%memit df:{df.shape}.assign(**cols:{len(assign_cols)})')
temp_df = df.copy()
%memit temp_df.assign(**assign_cols)
```

hamilton/function_modifiers/recursive.py
jernejfrank: Will make it consistent with spark -- in case this becomes a pain point can always add it later on.

elijahbenizzy (commented):
THis looks good -- let's rebase, I'll take one more look tomorrow morning, then we can merge and release!

elijahbenizzy (approved):

PR #1217, opened 2024-11-05
Add allow_module_overrides to AsyncDriver

### LINKED ISSUE #1216: allow_module_overrides doesn't work with AsyncBuilder
Tried using the `allow_module_overrides` with an AsyncBuilder. It sets the param on the builder, but it does not get passed in when the AsyncDriver (and thus the Driver) is initialized: 
1. https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/async_driver.py#L490-L495
2. https://github.com/DAGWorks-Inc/hamilton/blob/main/hamilton/async_driver.py#L232-L246

I think the arg just needs to be passed down from AsyncBuilder -> AsyncDriver -> Driver. 

Example: 
```python
builder = (
     AsyncBuilder()
    .with_config(config)
    .with_modules(*self.included_modules)
    .allow_module_overrides()
    .with_adapters(self.result_builder)
)
driver = await builder.build()
```

rwhitten577:
AsyncDriver got left out of the `.allow_module_overrides` feature. Adding support for it . 
Fixes #1216 

## Changes
* Pass `allow_module_overrides` from AsyncBuilder -> AsyncDriver -> Driver

## How I tested this
* Added new unit test to cover functionality
* Verified in my own project & DAG that it works

## Checklist

### DIFF
--- a/hamilton/async_driver.py
+++ b/hamilton/async_driver.py
@@ -198,6 +198,7 @@ class AsyncDriver(driver.Driver):
         *modules,
         result_builder: Optional[base.ResultMixin] = None,
         adapters: typing.List[lifecycle.LifecycleAdapter] = None,
+        allow_module_overrides: bool = False,
     ):
         """Instantiates an asynchronous driver.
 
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
Looks good! Would be nice to have a test for `build_without_init` (could just add this to the one you have, or create another), but that's not necessary

rwhitten577:
> Looks good! Would be nice to have a test for `build_without_init` (could just add this to the one you have, or create another), but that's not necessary

Thanks @elijahbenizzy , just added!

Issue #1218, opened 2024-11-05
fix: Better error for `@pipe` without `step()`


zilto:
This problem was shared by a user. 

## Problem
The issue is caused by this node where `@pipe_output` does contain `step()` but the use of `.when()` (which interacts with the config) may mean no `step` is provided (relates to #1213)

```python
@pipe_output(
    step(_foo).when(key="foo"),
    step(_bar).when(key="bar"),
)
def filtered_data(raw_data: pd.DataFrame) -> pd.DataFrame:
    return ...
```

This leads to this unhelpful error (trimmed). It indicates `IndexError: list index out of range` because the list of `step` is empty.

```python
IndexError                                Traceback (most recent call last)
    354 builder = base_builder.copy()
--> 355 dr = builder.with_config(config).with_modules(cell_module).build()
    357 # determine final vars
 
File ..., in Builder.build(self)
-> 2153 return Driver(
   ...
   2161 )

File ...\hamilton\driver.py:466, in Driver.__init__(self, config, adapter, allow_module_overrides, _materializers, _graph_executor, _use_legacy_adapter, *modules)
    464     error = telemetry.sanitize_error(*sys.exc_info())
    465     logger.error(SLACK_ERROR_MESSAGE)
--> 466     raise e
    467 finally:
    468     # TODO -- update this to use the lifecycle methods
    469     self.capture_constructor_telemetry(error, modules, config, adapter)

File ..., in Driver.__init__(self, config, adapter, allow_module_overrides, _materializers, _graph_executor, _use_legacy_adapter, *modules)
    438 self.graph_modules = modules
    439 try:
--> 440     self.graph = graph.FunctionGraph.from_modules(
   ...
    445     )

File ...\hamilton\graph.py:726, in FunctionGraph.from_modules(config, adapter, allow_module_overrides, *modules)
    707 @staticmethod
    708 def from_modules(
    ...
    713 ):
   ...
--> 726     nodes = create_function_graph(
    727         *modules, config=config, adapter=adapter, allow_module_overrides=allow_module_overrides
    728     )

File ...\hamilton\graph.py:170, in create_function_graph(config, adapter, fg, allow_module_overrides, *modules)
    168 # create non-input nodes -- easier to just create this in one loop
    169 for _func_name, f in functions:
--> 170     for n in fm_base.resolve_nodes(f, config):
    171         if n.name in config:
    172             continue  # This makes sure we overwrite things if they're in the config...

[...]

PR #1221, opened 2024-11-07
Bugfix pipe_output after config.when resolution

jernejfrank:
Addressing #1218

For `pipe_output`

```python
@pipe_output(
    step(_foo).when(key="foo"),
    step(_bar).when(key="bar"),
)
def filtered_data(raw_data: pd.DataFrame) -> pd.DataFrame:
    return ...
```
in case no conditions are met, e.g. `config={"key":"skip"}, it returns `filtered_data` as if `pipe_output` isn't there.

I think this is better than raising an error since it leaves the ability to choose transforms at runtime (which is related to using `config.when` on any other function in the DAG).

I also amended tests to capture this.

### DIFF
--- a/hamilton/function_modifiers/macros.py
+++ b/hamilton/function_modifiers/macros.py
@@ -1314,6 +1314,10 @@ class pipe_output(base.NodeTransformer):
             fn=fn,
         )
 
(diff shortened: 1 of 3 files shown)


skrawcz (approved):
Thanks @jernejfrank !

Issue #1228, opened 2024-11-13
Hamilton UI - Configuration menu for clean up and tracked execution infos


legout:
**Is your feature request related to a problem? Please describe.**
Currently, the Hamilton Tracker/UI saves a lot of information for every dataflow run, like success info, node parameters (inputs), node results(outputs), execution times,... 

For ETL dataflows, that runs often and processes a lot of data during every run, this results in a lot of data stored in the postgres/sqlite db (might be a problem) and the RUNS section in the UI becomes unresponsive.

Personally, I am most interested in the success info, execution times and logs. 

**Describe the solution you'd like**
1. Configuration of regular cleanups for the tracker data of each run. Ideally, this can be done individual for every data type (success info, inputs, outputs, logs,...)
2. Configuration of which kind of data should be stored by the tracker.

zilto:
Thank you for opening the issue. Following the discussion on Slack, I know @skrawcz reproduced the issue yesterday. We're working on a fix!

Tagging this related issue: #921

skrawcz:
CC @legout https://github.com/DAGWorks-Inc/hamilton/pull/1232 is a start. Let me know if you have comments. Should be able to build off of this and add more over time...

zilto:
(closing for now; feel free to reopen!)

skrawcz:
I'm going to reopen and modify the issue to focus on exposing endpoints in the server and UI to delete data.

legout:
@skrawcz 
This slack thread is related

https://hamilton-opensource.slack.com/archives/C03M33QB4M8/p1739280812916689

elijahbenizzy:
Yeah -- should be easy enough to delete all db items below a certain date. Extra requirements:
1. Make idempotent -- if it fails, we should be able to run again
2. (ideally) -- task-based -- likely going to be blocking for now as we don't have task-based capabilities so it's a bit more complex, but will see if django offers something out of the box
3. (ideally) -- track prior calls/cleanups -- again, not for the first release but we can easily have a UI page + an endpoint for prior cleanup jobs

elijahbenizzy:
[...]

Issue #921, opened 2024-05-27
Enable configuration/tags to turn off data introspection in SDK 


skrawcz:
**Is your feature request related to a problem? Please describe.**
Currently the SDK introspects all outputs. Sometimes, especially in the case of Parallel, this could be a really long list of things; it'd be nice to turn it off selectively.

**Describe the solution you'd like**
Option 1: inspect tag to skip
```python
@tag(sdk=["skip.introspection"])
def my_func(...) -> ...:
```

Option 2: skip via configuration
```
sdk_config = {"my_func": ["skip.introspection"]}
HamiltonTracker(..., tracker_config=sdk_config) # or some config file
```

---
We'd then want these to both output some type of result that the UI could then render.
Minimally based on the type returned it could be something basic about it -- so when you got to the UI you can see that this was skipped.

**Describe alternatives you've considered**
N/A

**Additional context**
This is an optimization.

Functionality to hide sensitive information from being logged could follow a similar path -- but likely a more explicit decorator in the code would be best in this situation.

PR #1232, opened 2024-11-17
Add SDK config

skrawcz:
This adds a few things to enable one to more
easily customize what is captured via the SDK.

1. Should you capture any data statistics at all?
2. Max lengths of dicts and lists -> so they don't get too big.

You can configure this via three means:
1. python constants
2. config file variables
3. environment variables

There is a prescedence order. So there are default values in the module.
You can then override them via config file variables.
Which then can in turn be overriden by environment variables.
Lastly the user can always modify the constants by directly
changing the module variables.

Note: we also skip logging metadata from datasavers and loaders if
the CAPTURE_DATA_STATISTICS = False. We can fix this by special
casing it, but for now I don't think people would complain with
the current functionality.

We should still enable #921 but I think this is a simpler route for now.

## Changes
- telemetry.py
- sdk related files

## How I tested this
- locally

## Notes
This is related to:
 - #921 this still leaves that on the table, but a different implementation
 - #1228 this would give people an option to not log too much

## Checklist

### DIFF
--- a/ui/sdk/src/hamilton_sdk/tracking/constants.py
+++ b/ui/sdk/src/hamilton_sdk/tracking/constants.py
@@ -1,0 +1,87 @@
+"""This module contains constants for tracking.
+
+We then override these by:
+1. Looking for a configuration file and taking the section under `SDK_CONSTANTS`.
+2. Via environment variables. They should be prefixed with `HAMILTON_`.
+3. Lastly folks can manually adjust these values directly by importing the module and changing the value.
+
+Note: This module cannot import other Hamilton modules.
+"""
+
+import configparser
+import logging
+import os
+from typing import Any
+
+logger = logging.getLogger(__name__)
+
+# The following are the default values for the tracking client
+CAPTURE_DATA_STATISTICS = True
+MAX_LIST_LENGTH_CAPTURE = 50
+MAX_DICT_LENGTH_CAPTURE = 100
+
+# Check for configuration file
+# TODO -- add configuration file support
+DEFAULT_CONFIG_URI = os.environ.get("HAMILTON_CONFIG_URI", "~/.hamilton.conf")
+DEFAULT_CONFIG_LOCATION = os.path.expanduser(DEFAULT_CONFIG_URI)
+
+
+def _load_config(config_location: str) -> configparser.ConfigParser:
+    """Pulls config if it exists.
+
+    :param config_location: location of the config file.
+    """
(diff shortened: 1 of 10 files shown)


tests/plugins/test_plotly_extensions.py
```diff
@@ -15,6 +16,7 @@ def figure():
 def test_plotly_static_writer(figure: go.Figure, tmp_path: pathlib.Path) -> None:
     file_path = tmp_path / "figure.png"
 
+    os.environ["BROWSER_PATH"] = "/path/to/browser"  # required for PlotlyStaticWriter to work
```
elijahbenizzy, line 19: Mind adding to comments as to why this is required?

ui/sdk/src/hamilton_sdk/tracking/constants.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        with open(config_location) as f:
+            config.read_file(f)
+    except Exception:
+        pass
+
+    return config
+
+
+_constant_values = globals()
```
elijahbenizzy, line 44: I feel like you should just hardcode this to a dict then we don't have to deal with globals. E.G.

`constants_values={"CAPTURE_DATA_STATISTICS" : True, ...}`

Then you just load up the file and modify that. Should simplify the code.

elijahbenizzy (commented):
Looking reasonable, let's:
1. SImplify the way we use globals
2. Add docs

Will take another quick look after docs then let's merge.

tests/plugins/test_plotly_extensions.py
skrawcz: this will be removed.

ui/sdk/src/hamilton_sdk/tracking/constants.py
skrawcz: No I don't like that  because that would then encumber directly accessing the module values; or I could do something like this at the end to push them all into the module space... either way playing with globals.

elijahbenizzy: This has a single global versus multiple? Don't see why we want to directly access module values.

elijahbenizzy (approved):
This is fine but I don't see the need to change module-level variables for configuration? Just have a `.set()` function that takes in the values? I guess it's nice for auto-complete, but it should work with a dict anyway.

PR #1233, opened 2024-11-17
CI fixes because other packages break other packages

skrawcz:
Fixes for CI.

Contains 1 temporary fix around plotly.

## Checklist


elijahbenizzy (approved):
This will get easier if we drop 3.8 support, right? It's EOL officially now.

PR #1234, opened 2024-11-18
Feat/polars with columns , async with_columns pandas

jernejfrank:
Please let me know if scope creep too big and I can cut some things out into a new PR.

## Changes
- extract_columns for polars LazyFrame (converts columns into Expressions that just need to be collected at the end)
- with_columns for polars (supported eager and lazy execution -- that's why the above change)
- refactored with_columns by using `registry` single dispatch method and enabled it for Pandas / Polars for now.
- async with_columns for pandas
- target=None selects all viable sink nodes and appends to data frame

## How I tested this
- unit
- e2e
- examples

## Notes
The one thing I didn't touch is the spark extension (zero experience with pyspark), because the implementation so far is different / not sure if we can use extract_columns as straightforward as with pandas/polars, but happy to tackle that as well in case its possible.

## Checklist


zilto (commented):
Overall every excited about this feature! All of my comments are related to the plugins system and I didn't check the specific logic of the function modifier.

I think for now, you can use the `registry` or abstract backend approach. A larger cleanup and unification of the plugin system is planned for Hamilton 2.0. 

An open question is whether we should support `narwhals`. I would be in favor given `polars` **is already** a Hamilton plugin.

hamilton/function_modifiers/recursive.py
```diff
@@ -11,17 +14,21 @@
 else:
     from typing import NotRequired
 
+from pandas import DataFrame as PandasDataFrame
+from polars import DataFrame as PolarsDataFrame
+from polars import LazyFrame as PolarsLazyFrame
```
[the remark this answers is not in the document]
jernejfrank: Thanks for all the feedback @zilto!

Happy to go ahead with `registry`-approach since it seems a straightforward change / keeps it consistent with `extract_columns` that is used under the hood.

Could you confirm the following just to double-check that I understand correctly:
The change necessary is that we move away from a factory pattern with subclassing the different implementations from and changing the now abstract method `create_merge_node` into a `@singledispatch` method with the plugin extensions `h_pandas` and `h_polars` implementing it, the same as in `extract_columns`.

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +633,234 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
...
+        initial_nodes = (
+            []
+            if self.dataframe_subdag_param is not None
+            else self._create_column_nodes(inject_parameter=inject_parameter, params=params)
+        )
+
+        return inject_parameter, initial_nodes
+
+    @abc.abstractmethod
+    def create_merge_node(self, upstream_node: str, node_name: str) -> node.Node:
```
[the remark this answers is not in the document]
jernejfrank: Not sure this is appropriate here, the `with_columns` call could be somewhere in the middle of the DAG and the user doesn't want to collect a LazyFrame just yet for example. 

This node is mostly meant to implement the different `with_columns` flavours -- for example in Polars its straightforward API matching since a `with_columns` method already exists, Pandas we needed to manually implement it ourselves.

plugin_tests/h_polars/resources/with_columns_end_to_end.py
```diff
@@ -0,0 +1,66 @@
+import polars as pl
```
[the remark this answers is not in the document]
jernejfrank: Me neither, what I did for now is kept the general functions agnostic to the dataframe type under `tests/` and methods that need a data frame input under `plugin_tests/` where some of them are duplicate with just changing to the correct dataframe type. 

Would be good to get a general idea what the testing convention is @elijahbenizzy ?

jernejfrank:
I left `with_columns` in `h_pandas` for backwards compatibility and have something similar in `h_polars` to be consistent, but IMO we should deprecate `h_pandas` (super short lived plugin lol), remove it from `h_polars` and keep it central in `recursive.py`.

hamilton/function_modifiers/recursive.py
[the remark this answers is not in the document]
zilto: There's also a [hamilton.plugins.h_polars_lazyframe](https://github.com/DAGWorks-Inc/hamilton/blob/3d98be3e2cf8b8a0e399be77b2d2a7199fe63105/hamilton/plugins/h_polars_lazyframe.py) with a result builder for lazy frames.

From my understanding, the `create_merge_node()` serves the same purpose as `ResultBuilder.build_result()`, which we have implementations for pandas, polars, polars lazy frame, vaex, ibis, and more (IIRC). Wanted to point it out to save you development time. I think there are two paths:

- duplicate code: copy-paste logic from ResultBuilder into this `create_merge_node` method, but make sure to leave a `# TODO` noting that this is the same logic as result builders
- refactor to extract logic: the `.build_result()` or `create_merge_node()` should eventually end up as a method under `hamilton.registry` with `@singledispatch`. Then, the same implementation could be called by `@with_columns` and `ResultBuilder`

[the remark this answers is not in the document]
jernejfrank: Not quite. The `ResultBuilder` will take all the end nodes (`final_vars`) and try to build a dataframe out of it. There are two cases and an edge case:

1. If the end node is a dataframe, it adopts that.
2. If the end nodes are a list of series, it combines them into a dataframe.
3. Edge case: In the case the end nodes are mixed the result is funky and definetly not something useful. 

In our case we want to keep the original dataframe and append columns to it, so we need to insert all original columns + new columns into the `ResultsBuilder`. In the case a selected column already exists in the dataframe we need to prefer the selected over the existing. 

Now, when the user initiates the `with_columns` decorator he passes in a subset of the existing columns / will most likely extract himself a subset of existing columns and not all. So, for us to generate the whole frame we will need to essentially extract all the columns in the background to be able to use the `ResultBuilder` logic. 
 
It can be done, but I find the amount of gluing necessary to get what we need a bit much; I find we need a better reason to justify it than just code reusability. Mind you, we will still need to register the `merge_node` in the `registry` anyways and there we can levarage  library-specific `with_columns` implementations / implementing it for libraries that don't have it (e.g. Pandas -- literally one line of code) in a more succinct way.

So all in all, I am not a fan of using `ResultBuilder` here unless there is a good argument about the overall design structure.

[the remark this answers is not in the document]
zilto: Per my initial comment, code duplication is totally reasonable and desirable here, but we should add clear comments in `@with_columns` source code AND in the docstrings, docs, and user guides

Code reuse is not a goal in itself. Rather, it's about limiting the number of "concepts/abstraction" we introduce in Hamilton, which impacts downstream maintenance, documentation, user onboarding, etc. A frequent challenge mentioned with Hamilton is that 5 ways to do a single thing are presented to users, which can lead to heterogeneous coding practices within a team.

You highlighted good points that the Reference docs highlight:

- `ResultBuilder().build_result()`:  merge columns and scalars into a dataframe
- `@with_columns` merge operation: subdag that assigns new columns to a dataframe, e.g., merge a dataframe and columns

But then likely scenarios:
> User A has code using `ResultBuilder`. Workflow gets complex, they want to use `@subdag`. I tell them you can't pass a result builder to a `@subdag`. Then,  `@with_columns` should solve the problem, but it doesn't have the same behavior for handling scalars (I'm not fond of it personally)

The main challenge to unify both is that `ResultBuilder` need backwards compatibility for handling scalar. However, as you're writing logic to handle merging of dataframes + columns, it seems a reasonable TODO to update ResultBuilder to handle dataframes + columns + scalars in a backwards compatible (or non-backwards in a future major release).

Then, we have a much simpler story for users "result builder merges X, Y, Z, and `@with_columns` is `@subdag + ResultBuilder`" without other caveats. Part of the documentation effort is also orienting users, making informative suggestion and comparisons, and writing in red bold letters what problem it solves / when it works / when it won't work.

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +629,314 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
...
+                "extract + modify the columns and afterwards select them."
+            )
+
+        pruned_nodes = prune_nodes(subdag_nodes, self.select)
+        if len(pruned_nodes) == 0:
+            raise ValueError(
+                f"No nodes found upstream from select columns: {self.select} for function: "
+                f"{fn.__qualname__}"
+            )
+        # In case no node is selected we append all possible nodes that have a column type matching
```
jernejfrank, line 924: Adding here another comment for future reference:

At the moment `ResultBuilder` will create a column out of a scalar, native Polars `with_columns` does the same.

In our `with_columns` implementation this results in an error since we enforce only matching series type can be appended into the dataframe.

I am torn how we want to handle this: on the one hand we should try to mimick the same behaviour  as external libraries (since that is what users are used to / easier adoption with legacy code), on the other hand enabling this can lead to nasty bugs because it will fail silently.

hamilton/function_modifiers/recursive.py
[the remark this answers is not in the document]
elijahbenizzy: `create_merge_node` I don't think is the right way to do this, but close. 

The problem here is that we have not really defined an algebra for doing the atomic operations for `with_columns` cross-dataframe and it's not actually clear it's the same:

1. Spark does it differently
2. Pandas + polars are similar actually, but lazy stuff gets in the way

So, while long-term it's a bit cleaner to come up with a set of atomic operations and delegate to each one, I think that it's a little more than we need right now.

We should be able to register `with_columns` -- we can merge stuff later, but it'll be unecessarily restrictive for now.

Narwals can help, but really this is fairly simple -- we have three implementations and they don't use the same code (yet), so unless we confidently believe that they'll all look the same, I'd prefer to duplicate code and reimplement the entirity of `with_columns`. Perhaps we can group this to use the same interface, but again, it's not clear that all dataframe implementations *will* use the same interface...

hamilton/function_modifiers/recursive.py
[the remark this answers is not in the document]
elijahbenizzy: See comment above -- I think we don't have the right general pattern yet, and unifying is more trouble than it's worth. Bringing in the `ResultBuilder` construct is, IMO, a little more complexity than needed.

Looking at the bigger picture, we *might* want the next push to be a more generic `with_columns` -- @jernejfrank and I initially discussed that then downscoped, in favor of individual versions that we can merge.

So I think we implment pandas (done), and polars (done), then maybe one more (narwhals, which also gives other stuff for free) or maybe we have enough information to then unify/delete code.

hamilton/plugins/vaex_extensions.py
```diff
@@ -26,6 +26,15 @@ def fill_with_scalar_vaex(
     return df
 
 
+@registry.with_columns.register(vaex.dataframe.DataFrame)
```
elijahbenizzy, line 29: Should we be registering it as `NotImplemented`? Or should the fact that we *don't* register it mean it's not implemented...

hamilton/plugins/h_polars.py
```diff
@@ -54,3 +67,152 @@ def build_result(
 
     def output_type(self) -> Type:
         return pl.DataFrame
+
+
+# Do we need this here?
+class with_columns(with_columns_factory):
```
elijahbenizzy, line 73: So yeah, I think this should live individually -- the contract shouldb e per-dataframe-type (for now). Then we can add a central one, but this enables separate implementations and the fact that we're using the same class to be an implementation detail.

hamilton/plugins/h_polars.py
```diff
@@ -54,3 +67,152 @@ def build_result(
...
+            via with_columns. The length of each column has to match the original dataframe length.
+            Existing columns will be overridden.
+        :param namespace: The namespace of the nodes, so they don't clash with the global namespace
+            and so this can be reused. If its left out, there will be no namespace (in which case you'll want
+            to be careful about repeating it/reusing the nodes in other parts of the DAG.)
+        :param config_required: the list of config keys that are required to resolve any functions. Pass in None\
+            if you want the functions/modules to have access to all possible config.
+        """
+
+        super().__init__(
```
elijahbenizzy, line 211: Curious, why are we delegating to the superclass here? Why not just pass through?

hamilton/registry.py
```diff
@@ -79,6 +79,9 @@ def load_extension(plugin_module: ExtensionName):
         assert hasattr(
             mod, f"fill_with_scalar_{plugin_module}"
         ), f"Error extension missing fill_with_scalar_{plugin_module}"
+        assert hasattr(
```
elijahbenizzy, line 82: AHh, I see this is the answer to the above.Makes sense leaving it unimplemented in a few!

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +629,314 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
                 stack.append(dep_node)
             seen_nodes.add(dep)
     return output
+
+
+class with_columns(base.NodeInjector, abc.ABC):
```
elijahbenizzy, line 634: This is abstract -- where are the functions that are left as abstract? E.G. what does a subclass have to implement?

elijahbenizzy (commented):
OK, what you did here is super cool. This took me a while to wrap my head around.

I think I'm on board with the general approach, but have the following concerns (that I'm still grokikng). Thanks @zilto for pointing these out:
1. The level of commonality between implementations -- should the goal be to make it easy to add a new implementation (E.G. by using the registry), or is that premature optimization? 
2. How it's packaged -- what's the import/usage? Should the user care that there's a generic `with_columns`? Or not?
3. Related to (2) registry + import structure

Let's hop on a call at some point?  Or discuss asynchronously?

hamilton/function_modifiers/recursive.py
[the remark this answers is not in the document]
zilto: I had a call with Jernej in the meantime yesterday. The main notes:
- my intent with this particular thread was to point that out that currently `@with_columns != @subdag + ResultBuilder`. My impression from the docs and the code is that it *should* be the end goal
- the current implementations work, are useful, and should be merged. We just have to add mentions in docs and `# TODO` where appropriate
- the current `@with_columns` doesn't fill scalars like the `ResultBuilder`. But also, Polars `.with_columns()` does fill scalars, but PySpark `.withColumns()` doesn't (see below)

```python
import polars as pl

df = pl.DataFrame({"a": [10, 11], "b": ["aa", "bb"]})
df.with_columns(1)
```

</small><table border="1" class="dataframe"><thead><tr><th>a</th><th>b</th><th>literal</th></tr><tr><td>i64</td><td>str</td><td>i32</td></tr></thead><tbody><tr><td>10</td><td>&quot;aa&quot;</td><td>1</td></tr><tr><td>11</td><td>&quot;bb&quot;</td><td>1</td></tr></tbody></table></div>

```python
import pyspark
spark = pyspark.sql.SparkSession.builder.getOrCreate()
df = spark.createDataFrame([(10, 11), ("aa", "bb")], schema=["a", "b"])
df.withColumns({"c": 1}).show()
```
>[63](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:63)     jcol = _create_column_from_name(col)
     [64](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:64) else:
---> [65](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:65)     raise PySparkTypeError(
     [66](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:66)         error_class="NOT_COLUMN_OR_STR",
     [67](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:67)         message_parameters={"arg_name": "col", "arg_type": type(col).__name__},
     [68](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:68)     )
     [69](https://vscode-remote+wsl-002bubuntu-002d20-002e04.vscode-resource.vscode-cdn.net/home/tjean/projects/dagworks/hamilton/~/projects/dagworks/hamilton/.venv/lib/python3.11/site-packages/pyspark/sql/column.py:69) return jcol
> PySparkTypeError: [NOT_COLUMN_OR_STR] Argument `col` should be a Column or str, got int.

jernejfrank:
Ok, so here is my line of thinking with regards to the changes:

The `with_columns` consists of three parts:

1. We need the input node. This can be a full dataframe if `pass_datafame_as` is used or we need to extract columns into nodes if `columns_to_pass` is used. Given that some data frames are supported in Hamilton's `extract_columns` and some are not, this should be implemented on a per library basis.
2. We need the subtag nodes. Again, we can re-use Hamilton's `subdag` functionality, but some libraries again will need more (see h_spark) and is therefore again to be implemented on a per library basis.
3. Last is combining eveything into a single dataframe again to be implemented on a per library basis.

So what I decided is to leave three abstract methods:
1. `get_initial_nodes`
2. `get_subdag_nodes`
3. `create_merge_node`

that should create enough flexibility to implement any dataframe library, but is also concrete enough to wire together everything in `inject_nodes` from `NodeInjector`.

Now, every plugin library, `h_pandas`, `h_polars`, and `h_polars_lazyframe` inherits from this class and in the their initialisation calls out to the parent factory init, but passes in the required dataframe type (e.g. `pd.DataFrame`,  `pl.DataFrame`, or `pl.LazyFrame`) which is in turn derived from the extension modules. So in effect we use the `registry` approach without hard-binding us to needing to implementat any functionality in there.

Since that part of the API is private, should we want to switch to registry, the refactoring is straightforward and shouldn't get us into trouble down the road.

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +628,268 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
                 stack.append(dep_node)
             seen_nodes.add(dep)
     return output
+
+
+class with_columns_factory(base.NodeInjector, abc.ABC):
```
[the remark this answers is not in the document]
elijahbenizzy: Also adjust docs to point out that this is not user-facing.

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +628,268 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
...
+
+    In case you need more flexibility you can alternatively use ``pass_dataframe_as``, for example,
+
+    .. code-block:: python
+
+            # with_columns_module.py
+            def a_from_df(initial_df: pd.Series) -> pd.Series:
+                return initial_df["a_from_df"] / 100
+
+        def b_from_df(initial_df: pd.Series) -> pd.Series:
```
elijahbenizzy, line 700: spacing?

hamilton/plugins/h_pandas.py
```diff
@@ -144,52 +121,38 @@ def __init__(
...
-        self, inject_parameter: str, params: Dict[str, Type[Type]]
+        self, fn: Callable, inject_parameter: str, params: Dict[str, Type[Type]]
     ) -> List[node.Node]:
         output_type = params[inject_parameter]
 
-        def temp_fn(**kwargs) -> pd.DataFrame:
-            return kwargs[inject_parameter]
+        if inspect.iscoroutinefunction(fn):
+
+            async def temp_fn(**kwargs) -> Any:
```
elijahbenizzy, line 150: Curious why this is necessary -- we should be able to have it work with both sync and async? E.G. this function doesn't need to be awaited? Or is it that we end up awaiting this so we have to use an async function (even though its the same)? Comment would help.

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +628,268 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
...
+        input_types = typing.get_type_hints(fn)
+        if inject_parameter not in params:
+            raise InvalidDecoratorException(
+                f"Function: {fn.__name__} does not have the parameter {inject_parameter} as a dependency. "
+                f"@with_columns requires the parameter names to match the function parameters. "
+                f"If you wish do not wish to use the first argument, please use `pass_dataframe_as` option. "
+                f"It might not be compatible with some other decorators."
+            )
+
+        if input_types[inject_parameter] != required_type:
```
elijahbenizzy, line 751: Might want to use a subclass check instead of equality

hamilton/function_modifiers/recursive.py
```diff
@@ -626,3 +628,268 @@ def prune_nodes(nodes: List[node.Node], select: Optional[List[str]] = None) -> L
...
+                ...
+
+    the above would output a dataframe where the two columns ``a_from_df`` and ``b_from_df`` get
+    overwritten.
+    """
+
+    # TODO: if we rename the column nodes into something smarter this can be avoided and
+    # can also modify columns in place
+    @staticmethod
+    def _check_for_duplicates(nodes_: List[node.Node]) -> bool:
```
elijahbenizzy, line 722: nit -- `_contains_duplicates` might be better. Also, probably best not to have an underscore as we use it externally.

hamilton/function_modifiers/recursive.py
elijahbenizzy: To circle back, I thnk we're removing result builder to reduce complexity, right?

elijahbenizzy (commented):
Looking good! Happy with the architecture choices. One thing that would be helpful -- let's add a page on the docs (E.G. a `README` in `function_modifiers`) that addresses architectural decisions. A lot here on how we did it, and it'll be really helpful for picking up later.

Left a few comments, can approve and ship soon!

hamilton/plugins/h_pandas.py
jernejfrank: I just checked, it is not.. I thought since `extract_columns` is checking for a co-routine, we need to wire it through, but turns out it works without it as well. Same for the `create_merge_node`.

elijahbenizzy:
> Ok, so here is my line of thinking with regards to the changes:
> 
> The `with_columns` consists of three parts:
> 
> 1. We need the input node. This can be a full dataframe if `pass_datafame_as` is used or we need to extract columns into nodes if `columns_to_pass` is used. Given that some data frames are supported in Hamilton's `extract_columns` and some are not, this should be implemented on a per library basis.
> 2. We need the subtag nodes. Again, we can re-use Hamilton's `subdag` functionality, but some libraries again will need more (see h_spark) and is therefore again to be implemented on a per library basis.
> 3. Last is combining eveything into a single dataframe again to be implemented on a per library basis.
> 
> So what I decided is to leave three abstract methods:
> 
> 1. `get_initial_nodes`
> 2. `get_subdag_nodes`
> 3. `create_merge_node`
> 
> that should create enough flexibility to implement any dataframe library, but is also concrete enough to wire together everything in `inject_nodes` from `NodeInjector`.
> 
> Now, every plugin library, `h_pandas`, `h_polars`, and `h_polars_lazyframe` inherits from this class and in the their initialisation calls out to the parent factory init, but passes in the required dataframe type (e.g. `pd.DataFrame`, `pl.DataFrame`, or `pl.LazyFrame`) which is in turn derived from the extension modules. So in effect we use the `registry` approach without hard-binding us to needing to implementat any functionality in there.
> 
> Since that part of the API is private, should we want to switch to registry, the refactoring is straightforward and shouldn't get us into trouble down the road.

Nice, I think this is a good overview. Note there might *still* be shared stuff between the implementations, in which case you have two options to reduce duplicated code (should you want):

1. Joint/helper functions
2. Additional subclasses, E.G. for column-specific ones (polars/pandas)

But I think these will be options for later, and quite possibly over-engineered.

To be clear, this doesn't look like it works with spark, yet? Do you think that's a possibility?

elijahbenizzy (commented):
This code is way cleaner -- nice work! And more powerful. Will take another look later tonight but just left a few comments -- this is very inline with how we discussed it.

hamilton/plugins/h_polars.py
```diff
@@ -54,3 +73,216 @@ def build_result(
...
+            for column in self.select:
+                columns_to_append[column] = kwargs[column]
+
+            return df.with_columns(**columns_to_append)
+
+        column_type = registry.get_column_type_from_df_type(self.dataframe_type)
+        input_map = {column: column_type for column in self.select}
+        input_map[inject_parameter] = self.dataframe_type
+        merge_node = node.Node(
+            name="__append",
```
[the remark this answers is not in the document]
jernejfrank: Forgot to change that, `spark` has the last node as `_select` -- I think in our case we should use `_append` or `_with_columns` to avoid misunderstanding with the `select` functionality of polars.

elijahbenizzy (approved):

PR #1253, opened 2024-12-07
Typed dict extract fields

### LINKED ISSUE #1252: @extract_fields - support TypedDict as return type

I would like to be able to return an instance of `TypedDict` from a node to make my IDE warn me if I forgot some keys, for example:

```python
import dataclasses
from typing import TypedDict

import hamilton.function_modifiers

from .api.kittens import Kitten, KittensApi

@dataclasses.dataclass
class T_memo:
  last_kitten_timestamp: float

T_kittens = list[Kitten]

class T_kittens_and_updated_memo(TypedDict):
    kittens: T_kittens
    updated_memo: T_memo

@hamilton.function_modifiers.extract_fields(dict(kittens=T_kittens, updated_memo=T_memo))
def kittens_and_updated_memo(KittensApi: KittensApi, memo: T_memo) -> T_kittens_and_updated_memo:
  # TODO: fetch some kittens
  kittens: list[Kitten] = []
  return T_kittens_and_updated_memo(kittens=kittens, updated_memo=memo)
```

What I currently get is:

```
hamilton.function_modifiers.base.InvalidDecoratorException: For extracting fields, output type must be a dict or typing.Dict, not: <class 'kittens_project.pipeline.T_kittens_and_updated_memo'>
```

In the implementation of `@extract_fields`, treat `TypedDict` as a normal dict.

Write this code instead:

```python
from typing import Any
# ...

@hamilton.function_modifiers.extract_fields(dict(kittens=T_kittens, updated_memo=T_memo))
def kittens_and_updated_memo(KittensApi: KittensApi, memo: T_memo) -> dict[str, Any]:  # 1️⃣ declare return type as just dict
  # TODO: fetch some kittens
  kittens: list[Kitten] = []
  # ❌ bad idea - mypy / pyright won't warn about missing `updated_memo`
  #return dict(kittens=kittens)
  # 2️⃣ transform the result into a `dict` instance
  return dict(T_kittens_and_updated_memo(kittens=kittens, updated_memo=memo))
```

skrawcz:
Adds typeddict support for extract_fields

```python
from typing import TypedDict
from hamilton.function_modifiers import extract_fields

class MyDict(TypedDict):
    foo: str
    bar: int

@extract_fields()
def some_function()->MyDict:
    return MyDict(foo="s", bar=1)
```
The above will automatically extract the fields `foo` and `bar`.

You can also do:

```python
from typing import TypedDict
from hamilton.function_modifiers import extract_fields

class MyDict(TypedDict):
    foo: str
    bar: int

@extract_fields({"foo": str})
def some_function()->MyDict:
    return MyDict(foo="s", bar=1)
```
To only expose a subset of the fields.

## Changes
 - extract_fields decorator behavior -> it can now handle TypedDict

## How I tested this
 - locally
 - unit tests

## Notes
 - fixes #1252 

## Checklist


hamilton/function_modifiers/expanders.py
```diff
@@ -759,13 +759,25 @@ def validate(self, fn: Callable):
...
-            pass
+            _validate_extract_fields(self.fields)
+        elif typing_extensions.is_typeddict(output_type):
+            if self.fields is None:
+                self.fields = typing.get_type_hints(output_type)
+            else:
+                # check that fields is a subset of TypedDict that is defined
+                typed_dict_fields = typing.get_type_hints(output_type)
+                for k, v in self.fields.items():
+                    if typed_dict_fields.get(k, None) != v:
```
elijahbenizzy, line 776: Subtle issue here --  if the `typed_dict_fields.get(k)` is a subclass of `self.fields[k]` it should work (I think, although typing can be confusing). Use `custom_subclass_check`?

elijahbenizzy (commented):
One minor point

hamilton/function_modifiers/expanders.py
skrawcz: What would the use case be? I think someone using a subclass type of the field type in a typeddict is stretching things... can always extend, but for now I don't think we need to support.

zilto:
Could we loosen the constraints to enable `Mapping` type?

hamilton/function_modifiers/expanders.py
elijahbenizzy: The use-case is if the fields declare a `Sequence` and the typed dict is a list. It's valid. This is an issue that we've had with the rest of our codebase (not accounting for subclasses) that shows up in weird ways. 

- pandas versus geopandas (I think)
- Sequence versus list
- Mapping versus dict, etc...

As to why you'd want this, this will happen if you're using a typed dict from somewhere else. While it's unlikely, we do this everywhere we can in the codebase.

We have a utility function that has been tuned to get this right: https://github.com/DAGWorks-Inc/hamilton/blob/4e8dc43ca802b9fcdf9059dfdb67f68d5cc38e41/hamilton/htypes.py#L38.

skrawcz: Because we already have that function I'm happy to delegate to it.

skrawcz:
> Could we loosen the constraints to enable `Mapping` type?

if it's quick to do. Otherwise can punt for another PR.

hamilton/function_modifiers/expanders.py
elijahbenizzy: Yeah would not be worth it if we hadn't commonly be doing it elsewhere.

Issue #1263, opened 2024-12-30
Caching for async driver


nemilentsau:
I am getting a value error when trying to use caching with an async_driver 

# Current behavior
dr = await (
    async_driver.Builder().with_modules(pdf_processor_async).with_cache().build()
)

ValueError                                Traceback (most recent call last)
[/home/sagemaker-user/ai-experiment/llm-frameworks/hamilton/test_hamilton_async.ipynb](https://vscode-remote+cjivqy5cbq7ohcr-002estudio-002eus-002dwest-002d2-002esagemaker-002eaws.vscode-resource.vscode-cdn.net/home/sagemaker-user/ai-experiment/llm-frameworks/hamilton/test_hamilton_async.ipynb) Cell 6 line 1
----> 1 dr = await (
      2     async_driver.Builder().with_modules(pdf_processor_async).with_cache().build()
      3 )

ValueError: Multiple adapters cannot (currently) implement the same lifecycle method. Sync methods: ['do_node_execute']. Async methods: []

## Stack Traces

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
[/home/sagemaker-user/ai-experiment/llm-frameworks/hamilton/test_hamilton_async.ipynb](https://vscode-remote+cjivqy5cbq7ohcr-002estudio-002eus-002dwest-002d2-002esagemaker-002eaws.vscode-resource.vscode-cdn.net/home/sagemaker-user/ai-experiment/llm-frameworks/hamilton/test_hamilton_async.ipynb) Cell 6 line 1
----> 1 dr = await (
      2     async_driver.Builder().with_modules(pdf_processor_async).with_cache().build()
      3 )

File [/opt/conda/lib/python3.10/site-packages/hamilton/async_driver.py:512](https://vscode-remote+cjivqy5cbq7ohcr-002estudio-002eus-002dwest-002d2-002esagemaker-002eaws.vscode-resource.vscode-cdn.net/opt/conda/lib/python3.10/site-packages/hamilton/async_driver.py:512), in Builder.build(self)
    504 async def build(self):
    505     """Builds the async driver. This also initializes it, hence the async definition.
    506     If you don't want to use async, you can use `build_without_init` and call `ainit` later,
    507     but we recommend using this in an asynchronous lifespan management function (E.G. in fastAPI),
   (...)
    510     :return: The fully
    511     """
--> 512     dr = self.build_without_init()
    513     return await dr.ainit()

[...]

PR #1264, opened 2025-01-01
Adds FutureAdapter that delegates executions to a threadpool for parallelization

skrawcz:
Taking inspiration from #1263, I implemented
a similar adapter to how async works.

We get away with this because we don't encounter
SERDE boundaries.

If you run the example DAG you'll see that:

1. it is parallelized as it should be
2. you can use caching and the tracking adapter

Rough edges:
 - doesn't support DAGs with parallelizable / collect. 
 - haven't tested it with other adapters

## Changes
 - adds adapter that lazily enables parallelism via threadpool executor
 - docs
 - example

## How I tested this
 - runs locally using the example

See this image from the tracker showing that things are running in parallel:
Screen Shot 2024-12-31 at 10 25 22 PM

## Notes - TODO:

- [x] unit tests
- [x] example against async & comparison.
~- [ ] example uses materializers & caching to exercise things.~
- [x] check error experience

## Checklist


hamilton/plugins/h_threadpool.py
```diff
@@ -0,0 +1,58 @@
...
+        return self.executor.submit(_new_fn, execute_lifecycle_for_node, **kwargs)
+
+    def build_result(self, **outputs: Any) -> Any:
+        """Given a set of outputs, build the result.
+
+        :param outputs: the outputs from the execution of the graph.
+        :return: the result of the execution of the graph.
+        """
+        for k, v in outputs.items():
+            if isinstance(v, Future):
```
elijahbenizzy, line 56: I think you want `concurrent.futures.as_completed(...)`. Can't convince myselfe that we're not deadlocking in certain cases, but I think the fact that we're doing topological order should be good enough...

elijahbenizzy (approved):
Mark as experimental IMO

hamilton/plugins/h_threadpool.py
skrawcz: I don't think this matters here. Since this wont block anything executing.

Issue #1268, opened 2025-01-06
Build DAG from functions


zilto:
# Summary
Add first-class way to build a DAG from function objects (in contrast to DIY/ hacky). The user API could be either:
- `Builder().with_functions()`
- or an intermediary `my_module = module_from_functions(fns*)` then passed to `.with_modules()`

Goals:
- simplify codebase
- foundation for better hierarchical / nested  graph structures (i.e., subdags)

# Current
At the core of Hamilton, users:
1. write functions in a Python module ("dataflow code")
2. load that module in the "driver code" to build a DAG
3. execute the DAG via the Driver

## Problem
- It is currently possible to build Hamilton DAGs from functions, but we have no official "here's how you do it" that we guarantee we'll support.
  - `ad_hoc_utils` means exactly the opposite
- In a notebook, there's no good reason to create a module from a notebook function before passing to Hamilton (except our how constraints)
- Python module machinery is complex and adds indirection to the codebase (tests, notebook extension, LSP)

## Benefits
- **greatly** simply **many** unit tests
- facilitate marimo integration 

## Hamilton 2.0 / Broader perspective

There's no well-defined structure or purpose to Hamilton top-level modules (e.g., `nodes`, `graph_types`, `graph_utils`, `graph`, `ad_hoc_utils`, `base`, `hamilton.common`, `models`). I propose a structure that matches the Hamilton lifecycle:
- `hamilton.parser`: everything that deals with source code: how functions are written, if type annotations are present (not type matching), collecting functions from modules, converting a notebook cell string to a module, remove comments and docstring before hashing source code
- `hamilton.compiler`: converting code to DAG: structuring the DAG from functions, applying function modifiers, validating types, etc.
- remove `ad_hoc_utils`

Dev-iL:
Could you please clarify what would be the difference in terms of user-facing code? Are we not currently building DAGs exclusively out of functions?

skrawcz:
Yes there could be a big refactor for Hamilton 2.0. 

Otherwise the reason trying to make things look like a module was done, was so that we wouldn't have to rewire the internals of Hamilton which assumed a module would be passed in. `ad_hoc_utils` was named that on purpose because in our opinion we wanted people curating their code into modules as part of a SDLC; coupling functions with where you execute them makes them less usable and modular since you likely couple execution imports with the ones for the functions... Now there's been a lot of lessons and learnings since then, so there could be improvements here.

Question:

[...]

PR #1271, opened 2025-01-13
[WIP] feat: build FunctionGraph from functions

zilto:
- Relates to https://github.com/DAGWorks-Inc/hamilton/issues/1268

## Changes

- The changes affect core code path, but is generally safe. Only outdated tests using `graph.create_function_graph()` instead of `graph.FunctionGraph.from_modules()` needed to be updated.
- The function signature of `graph.create_function_graph()` was modified, but it was never advertised as a public API. Further rewiring could provide full backwards compatibility (could do after validating the general solution)

## TODO
- update documentation
- change `create_function_graph()` for fully backwards-compatible APi

## How I tested this
- all tests pass after slight edits

## Notes
- `graph.create_function_graph()` should really return a `FunctionGraph` instead of a `dict`

### DIFF
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -143,7 +143,7 @@ def update_dependencies(
 
 
 def create_function_graph(
-    *modules: ModuleType,
+    *functions: List[Tuple[str, Callable]],
     config: Dict[str, Any],
     adapter: lifecycle_base.LifecycleAdapterSet = None,
     fg: Optional["FunctionGraph"] = None,
@@ -164,7 +164,6 @@ def create_function_graph(
         nodes = {}  # name -> Node
     else:
         nodes = fg.nodes
-    functions = sum([find_functions(module) for module in modules], [])
 
     # create non-input nodes -- easier to just create this in one loop
     for _func_name, f in functions:
@@ -733,8 +732,18 @@ class FunctionGraph:
         :return: a function graph.
         """
 
+        functions = sum([find_functions(module) for module in modules], [])
+        return FunctionGraph.from_functions(*functions, config=config, adapter=adapter, allow_module_overrides=allow_module_overrides)
+
+    @staticmethod
+    def from_functions(
+        *functions,
+        config: Dict[str, Any],
+        adapter: lifecycle_base.LifecycleAdapterSet = None,
+        allow_module_overrides: bool = False,
+    ) -> "FunctionGraph":
         nodes = create_function_graph(
-            *modules, config=config, adapter=adapter, allow_module_overrides=allow_module_overrides
+            *functions, config=config, adapter=adapter, allow_module_overrides=allow_module_overrides
         )
         return FunctionGraph(nodes, config, adapter)
 
--- a/hamilton/graph.py
+++ b/hamilton/graph.py
@@ -733,7 +733,12 @@ class FunctionGraph:
         """
 
         functions = sum([find_functions(module) for module in modules], [])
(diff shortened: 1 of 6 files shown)


elijahbenizzy:
Hey -- sorry, it's been a bit, will be looking over soon!

tests/function_modifiers/test_adapters.py
```diff
@@ -686,12 +681,9 @@ def test_load_from_with_multiple_inputs():
     def fn(data1: dict, data2: dict) -> dict:
         return {**data1, **data2}
 
-    fg = graph.create_function_graph(
-        ad_hoc_utils.create_temporary_module(fn),
-        config={},
-    )
+    fn_graph = graph.FunctionGraph.from_modules(ad_hoc_utils.create_temporary_module(fn), config={})
     # One filter, one loader for each and the transform function
-    assert len(fg) == 5
```
skrawcz, line 694: why does this need to change?

zilto: because the previous `graph.create_function_graph()` returns a `dict` not a `FunctionGraph`.

In the edited test, I changed it to `graph.FunctionGraph.from_modules()` which actually returns a `FunctionGraph`. The `.nodes` attribute returns the `dict` that's tested in the old test

hamilton/graph.py
```diff
@@ -143,7 +143,7 @@ def update_dependencies(
 
 
 def create_function_graph(
```
skrawcz, line 145: The type signature of this should be updated.

skrawcz: also 🤦 (at myself I think) that the name of this function and what it returns - a dict, not a function graph...

tests/function_modifiers/test_adapters.py
skrawcz: got it. thanks

skrawcz:
Yeah so what's left? This ?

* add in ability to pass functions and update code to handle that?
* add tests for new functionality
* add example

skrawcz:
My expectation

```python

# module Z

# module 1
def func1( ...) -> ...:
  ...

# module 1, or perhaps another module 2
def func2(...)->...:
  ...

# driver code somewhere else or in one of the modules - importing things appropriately
funcs = [func1, func2] 

driver.Builder().with_modules(module_z, *funcs) # this should work
```

leblancfg:
I saw the note in #1271:

>Benefits
>  facilitate marimo integration

FYI I attempted this with this branch, but it returns `TypeError: unhashable type: 'Cell'`.

zilto:
> I saw the note in #1271:
> 
> > Benefits
> > facilitate marimo integration
> 
> FYI I attempted this with this branch, but it returns `TypeError: unhashable type: 'Cell'`.

@leblancfg You're ahead of us! Will open a separate PR and share my marimo development pattern once this is merged. Feel free to discuss here #1274

elijahbenizzy:
OK, `with_modules(...)` shouldn't accept functions. `with_function(...)` should, IMO. 

That said -- in order to have function overrides, we should probably want to combine `from_fucntions` and `from_modules` into `construct(...)` or `compile(...)` that takes in a list of functions/modules. Happy to take a stab! Should be straightforward -- we can even leave around the old one for backwards compatibility.

skrawcz:
> OK, with_modules(...) shouldn't accept functions. with_function(...) should, IMO.

Disagree. I think this will make things more confusing, as it will:

1. create two ways to create nodes in the graph.
2. provide no clear precedence when say someone wants to override a function in a module.

zilto:
closing in favor of #1275

PR #1276, opened 2025-02-08
Resolve `Annotated` types (PEP 593) on dataflow visualizations

cswartzvi:
Recently, I was attempting to use `Annotated` types (PEP 593) for the output of some nodes (see `A`):

```python
from typing import Annotated

def A() -> Annotated[int, "metadata"]:
    return 42

def B(A: int) -> float:
    """Divide A by 3"""
    return A / 3

def C(A: int, B: float) -> float:
    """Square A and multiply by B"""
    return A**2 * B

if __name__ == "__main__":
    import __main__
    from hamilton import driver

    result = dr.execute(["C"])
    print(result)
```

The execution works as excepted, but when I run `dr.display_all_functions("dag.png")` the type of `A` is masked as `Annotated` on the resulting graphic:

dag1

Ideally, IMHO, we would want to see the underlying type such as:

dag2

## Changes

I updated `hamilton.htypes.get_type_as_string` so that it will consider `Annotated` types. Note that it appears that `typing.get_origin` and `typing_inspect.get_origin` do *not* produce the same result for `Annotated` types (see https://github.com/ilevkivskyi/typing_inspect/issues/109).

## How I tested this

I updated the current test for `get_type_as_string` in `test_type_utils.py` and added an additional check focusing on `Annotated` types.

## Checklist

### DIFF
--- a/tests/test_type_utils.py
+++ b/tests/test_type_utils.py
@@ -178,6 +178,7 @@ def test__safe_subclass(candidate, type_, expected):
         (pd.Series),
         (htypes.column[pd.Series, int]),
         (pd.DataFrame),
+        (typing.Annotated[int, "metadata"]),
     ],
 )
 def test_get_type_as_string(type_):
@@ -186,6 +187,12 @@ def test_get_type_as_string(type_):
         type_string = htypes.get_type_as_string(type_)  # noqa: F841
     except Exception as e:
         pytest.fail(f"test get_type_as_string raised: {e}")
+
+
+def test_type_as_string_with_annotated_type():
+    """Tests the custom_subclass_check"""
+    type_string = htypes.get_type_as_string(typing.Annotated[int, "metadata"])  # type: ignore
+    assert type_string == "int"
 
 
 @pytest.mark.parametrize(
--- a/tests/test_type_utils.py
+++ b/tests/test_type_utils.py
@@ -9,6 +9,11 @@ import pytest
 from hamilton import htypes
 from hamilton.htypes import check_instance
 
+if sys.version_info >= (3, 9):
+    from typing import Annotated
+else:
+    from typing_extensions import Annotated
+
 
 class X:
     pass
@@ -178,7 +183,7 @@ def test__safe_subclass(candidate, type_, expected):
         (pd.Series),
         (htypes.column[pd.Series, int]),
         (pd.DataFrame),
-        (typing.Annotated[int, "metadata"]),
+        (Annotated[int, "metadata"]),
     ],
 )
 def test_get_type_as_string(type_):
@@ -191,7 +196,7 @@ def test_get_type_as_string(type_):
 
 def test_type_as_string_with_annotated_type():
     """Tests the custom_subclass_check"""
-    type_string = htypes.get_type_as_string(typing.Annotated[int, "metadata"])  # type: ignore
+    type_string = htypes.get_type_as_string(Annotated[int, "metadata"])  # type: ignore
     assert type_string == "int"
 
 
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -99,7 +99,7 @@ def get_type_as_string(type_: Type) -> Optional[str]:
     :return: string representation of the type. An empty string if everything fails.
     """
 
-    if typing.get_origin(type_) is Annotated:  # differs from typing_inspect.get_origin
+    if _is_annotated_type(type_):
         type_string = get_type_as_string(typing.get_args(type_)[0])
     elif getattr(type_, "__name__", None):
         type_string = type_.__name__
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -99,7 +99,9 @@ def get_type_as_string(type_: Type) -> Optional[str]:
     :return: string representation of the type. An empty string if everything fails.
     """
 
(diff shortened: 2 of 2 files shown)


skrawcz (commented):
I think this looks good. @elijahbenizzy any thoughts? 

Tests seem to be failing due to dask releasing a new version -- will try to see if I can make a fix on a separate branch for them.

zilto:
The changes seem very clean. It's reasonable to think that `Annotated` is never the type of interest to display. 

Regarding `typing_inspect`, there's an open ticket. Would be easiest to remove it as a dependency once Python 3.9 reaches end of life # 472

elijahbenizzy:
This looks good! Technically this *could* be construed as backwards incompatible (E.G. if you were using the [node -> dict representation](https://github.com/DAGWorks-Inc/hamilton/blob/7f4bc6633bc683b511cf95dbab1521d01bc894af/hamilton/graph_types.py#L114)), but I'm not particularly concerned about that. Specifically, annotated is metadata that may not be serialized, so I'm OK with it. We also don't make guarantees on the stability of the node to dict function. But if you wanted to be a stickleer about backwards compatibility, you could add a parameter (hide_annotated) that defaults to false.

Up to you -- happy to merge either way!

cswartzvi:
Ah, wow, I didn't even think about that incompatibility. Thinking out loud here... I see that the result of `HamiltonNode.to_dict` is used in both the mlflow and openlineage plugins as a way (I believe) to recreate a hamilton run. I unfortunately do not use either mlflow or openlineage (yet), but wouldn't including `Annotated` in the serialized result prevent it from recreating the run (in a type checked manner)? 

I completely understand that it is a breaking change that someone out there could be relying on, perhaps it's better to introduce something like as temporary provision:

```python
def get_type_as_string(type_: Type, resolve_annotated: bool = True) -> Optional[str]:
    """Get a string representation of a type.

    The logic supports the evolution of the type system between 3.8 and 3.10.
    :param type_: Any Type object. Typically the node type found at Node.type.
    :param resolve_annotated: Determines if Annotated types should be resolved to their underlying type.
    :return: string representation of the type. An empty string if everything fails.
    """

    if resolve_annotated and _is_annotated_type(type_):
        type_string = get_type_as_string(typing.get_args(type_)[0])
    elif getattr(type_, "__name__", None):
        type_string = type_.__name__
    elif typing_inspect.get_origin(type_):
        base_type = typing_inspect.get_origin(type_)
        type_string = get_type_as_string(base_type)
    elif getattr(type_, "__repr__", None):
        type_string = type_.__repr__()
    else:
        type_string = None

    return type_string
```

With all that said, if you think that mlflow or openlineage would benefit from having the actual type (vice `Annotated`), then my vote is to merge as is.

elijahbenizzy:
> @elijahbenizzy Ah, wow, I didn't even think about that incompatibility. Thinking out loud here... I see that the result of `HamiltonNode.to_dict` is used in both the mlflow and openlineage plugins as a way (I believe) to recreate a hamilton run. I unfortunately do not use either mlflow or openlineage (yet), but wouldn't including `Annotated` in the serialized result prevent it from recreating the run (in a type checked manner)?
> 
> I completely understand that it is a breaking change that someone out there could be relying on, perhaps it's better to introduce something like as temporary provision:
> 
> ```python
> def get_type_as_string(type_: Type, resolve_annotated: bool = True) -> Optional[str]:
>     """Get a string representation of a type.
> 
>     The logic supports the evolution of the type system between 3.8 and 3.10.
>     :param type_: Any Type object. Typically the node type found at Node.type.
>     :param resolve_annotated: Determines if Annotated types should be resolved to their underlying type.
>     :return: string representation of the type. An empty string if everything fails.
>     """
> 
>     if resolve_annotated and _is_annotated_type(type_):
>         type_string = get_type_as_string(typing.get_args(type_)[0])
>     elif getattr(type_, "__name__", None):
>         type_string = type_.__name__
>     elif typing_inspect.get_origin(type_):
>         base_type = typing_inspect.get_origin(type_)
>         type_string = get_type_as_string(base_type)
>     elif getattr(type_, "__repr__", None):
>         type_string = type_.__repr__()
>     else:
>         type_string = None
> 
>     return type_string
> ```
> 
> With all that said, if you think that mlflow or openlineage would benefit from having the actual type (vice `Annotated`), then my vote is to merge as is.

Yeah, so thinking about it, `Annotated` (without the right information, as it was in the diagram) is a relatively useless type (the prior string), so I think this is strictly better -- we can call it a bug that we fixed?

cswartzvi:
@elijahbenizzy That sounds good to me! I also had an idea (that I mention only for the sake of posterity), that we could separate the type from the metadata and return both (in a more formalized result) if it was ever needed by a consumer.

skrawcz:
okay if you rebase from main then dask issues should go away.

tests/test_type_utils.py
```diff
@@ -188,6 +194,12 @@ def test_get_type_as_string(type_):
         pytest.fail(f"test get_type_as_string raised: {e}")
 
 
+def test_type_as_string_with_annotated_type():
```
elijahbenizzy, line 197: If you feel like doing more testing work we should really test the string output of these, not just that they fail. But that's out of scope.

elijahbenizzy (approved):

PR #1282, opened 2025-02-26
WIP: decorators should wrap functions with vanilla wrapper

skrawcz:
This is a prototype sketch.

1. each decorator wraps the function -- and returns that pointer.
2. we only wrap a function 1 level deep.
3. we validate on the original function.

This should then enable:
1. reusing base functions easily.
2. not mutating the original function.
3. also work in SERDE contexts -- but something to properly test.

## Checklist

### DIFF
--- a/hamilton/function_modifiers/base.py
+++ b/hamilton/function_modifiers/base.py
@@ -4,6 +4,7 @@ import functools
 import itertools
 import logging
 from abc import ABC
+from inspect import unwrap
 
 try:
     from types import EllipsisType
@@ -99,18 +100,30 @@ class NodeTransformLifecycle(abc.ABC):
         :param fn: Function to decorate
         :return: The function again, with the desired properties.
         """
-        self.validate(fn)
+        # stop unwrapping if not a hamilton function
+        # should only be one level of "hamilton wrapping" - and that's what we attach things to.
+        self.validate(unwrap(fn, stop=lambda f: not hasattr(f, "__hamilton__")))
+        if not hasattr(fn, "__hamilton__"):
+
+            @functools.wraps(fn)
+            def wrapper(*args, **kwargs):
+                return fn(*args, **kwargs)
+
+            wrapper.__hamilton__ = True
+        else:
+            wrapper = fn
+
         lifecycle_name = self.__class__.get_lifecycle_name()
-        if hasattr(fn, self.get_lifecycle_name()):
+        if hasattr(wrapper, self.get_lifecycle_name()):
             if not self.allows_multiple():
                 raise ValueError(
                     f"Got multiple decorators for decorator @{self.__class__}. Only one allowed."
                 )
-            curr_value = getattr(fn, lifecycle_name)
-            setattr(fn, lifecycle_name, curr_value + [self])
+            curr_value = getattr(wrapper, lifecycle_name)
+            setattr(wrapper, lifecycle_name, curr_value + [self])
         else:
-            setattr(fn, lifecycle_name, [self])
-        return fn
+            setattr(wrapper, lifecycle_name, [self])
+        return wrapper
 
     def required_config(self) -> Optional[List[str]]:
         """Declares the required configuration keys for this decorator.


elijahbenizzy:
Copying the function might be cleaner -- if this works then great but I think there are a few assumptions that you'll have to check for (async, signature stuff), which `functools.wraps` doesn't cover.

skrawcz:
> Copying the function might be cleaner -- if this works then great but I think there are a few assumptions that you'll have to check for (async, signature stuff), which `functools.wraps` doesn't cover.
> 
> Might be easier to use this utility function:
> 
> https://github.com/DAGWorks-Inc/hamilton/blob/1057a40db370d235fce647d65872221365da6bfa/hamilton/ad_hoc_utils.py#L15

yep some testing to be done for sure.

skrawcz:
@jernejfrank any thoughts on what assumption I broke with `@mutate`?

skrawcz:
> @jernejfrank any thoughts on what assumption I broke with `@mutate`?

figured it out. I have a hack that works I think. will need to refactor a few things in this PR for a proper change

jernejfrank:
> > @jernejfrank any thoughts on what assumption I broke with `@mutate`?
> 
> figured it out. I have a hack that works I think. will need to refactor a few things in this PR for a proper change

Great, just to write down why I think it failed for easier lookup: 

`mutate` looks at the functions you want to mutate and figures out if there's already a `pipe_output` decorator on it. If so, it appends the mutating function (the one decorated with `mutate`) as the next step in the already existing `pipe_output` sequence of steps. Otherwise, it will create a `pipe_output` decorator on the target function with the mutating function as the step.

I guess with the new wrapper for the decorators, this check if `pipe_output` exists doesn't work, because it looked at the wrong level?

skrawcz:
Okay there is one bug with parallel & collect on ray using the tracker -- the parallelized nodes aren't being reflected in the Hamilton UI -- I suspect it could be something could be messing with the "source code" pointers the SDK captures or something.

hamilton/function_modifiers/base.py
```diff
@@ -99,18 +101,40 @@ def __call__(self, fn: Callable):
         :param fn: Function to decorate
         :return: The function again, with the desired properties.
         """
-        self.validate(fn)
+        # # stop unwrapping if not a hamilton function
+        # # should only be one level of "hamilton wrapping" - and that's what we attach things to.
+        self.validate(unwrap(fn, stop=lambda f: not hasattr(f, "__hamilton__")))
```
elijahbenizzy, line 106: Why do we only do one level deep? Aren't there some weird cases with multiple imports, imports of imports, etc...?

hamilton/function_modifiers/base.py
```diff
@@ -805,6 +829,9 @@ def resolve_nodes(fn: Callable, config: Dict[str, Any]) -> Collection[node.Node]
     which configuration they need.
     :return: A list of nodes into which this function transforms.
     """
+    # check for mutate...
+    fn = handle_mutate_hack(fn)
```
elijahbenizzy, line 833: Take away the name "hack"

hamilton/function_modifiers/base.py
```diff
@@ -833,6 +860,36 @@ def resolve_nodes(fn: Callable, config: Dict[str, Any]) -> Collection[node.Node]
...
+    directly wrap the function, but mutate does not. It adds
+    a pointer to the function in question that we follow here.
+
+    This code depends on what the mutate class does
+    and what the pipe_output decorator does.
+
+    :param fn: Function to check
+    :return: Function or wrapped function to use if applicable.
+    """
+    if hasattr(fn, "__hamilton_wrappers__"):
```
elijahbenizzy, line 881: Is this worth the complexity it adds? Is there a simpler way to approach it? Need to grok better.

hamilton/function_modifiers/expanders.py
```diff
@@ -231,6 +231,65 @@ def replacement_function(
                     # now the error will be clear enough
                 return node_.callable(*args, **new_kwargs)
 
+            async def async_replacement_function(
```
elijahbenizzy, line 234: note -- merge with the above when you get to polishing

elijahbenizzy (commented):
If it works then great, but wow there's a lot of complexity here. I think this is a decent approach, but curious if there's a way to simplify accounting for the wrappers?

hamilton/function_modifiers/base.py
skrawcz: The problem is that with mutate the current design was the attach to the function and it woudl be resolved properly.

I think the better approach would be to register mutates via a registry that is applied to functions that are found ... and that way we wouldn't need to attach it to the function directly...

jernejfrank: Curious, if we use a registry for mutate, would it be sensible to just have one global registry for all decorators instead of the additional 1-layer wrapper?

hamilton/function_modifiers/base.py
elijahbenizzy: Yes, I got that.  Why *just* one layer was the question?

skrawcz: > Why just one layer 

1. Doesn't seem like we need more? 
2. This mirrors the prior behavior of directly attaching to the function - there were no layers before.

PR #1287, opened 2025-03-02
Add Task Submission / Return Hooks

cswartzvi:
Current pre/post task execution hooks may run on out-of-process executors potentially causing issues with certain logging and stateful adapters. This PR adds additional hooks that are run *before* a task is submitted to an executor and *after* a task future is resolved - both on the main process.

## Changes

I added task submission / resolution hooks (`BasePreTaskSubmission` and `BasePostTaskResolution`) and adapters (`TaskSubmissionHook` and `TaskResolutionHook`) to the codebase as well as updated the main task execution function `run_graph_to_completion` to call the new hooks. In support of this I made `TaskImplementation` hashable via it's `task_id` in order to avoid a parallel structure - @elijahbenizzy perhaps in the future this could be used to a queue-based task processor?

## How I tested this

I added individual tests for both `TaskSubmissionHook` and `TaskResolutionHook` and added them to `test_multi_hook`.

## Notes
1. The docs for `TaskSubmissionHook` and `TaskResolutionHook`, as well some other task execution hooks, have been updated. I think we could add some more details - maybe a flow chart of when the individual hooks are called? I can add that if the current changes seems reasonable.
2. `TaskResolutionHook` currently has an optional  `error` attribute. I am on the fence as to whether or not this is needed. My thinking is that since `TaskFuture` essentially wraps `concurrent.futures.Future` we could forward the `exception` call. Thoughts? Currently `error`  is always `None`.

## Checklist

### DIFF
--- a/hamilton/execution/executors.py
+++ b/hamilton/execution/executors.py
@@ -416,7 +416,7 @@ def run_graph_to_completion(
     execution_manager.init()
     try:
         while not GraphState.is_terminal(execution_state.get_graph_state()):
-            # get the next task from the queue
+            # Get the next task from the queue
             next_task = execution_state.release_next_task()
             if next_task is not None:
                 task_executor = execution_manager.get_executor_for_task(next_task)
@@ -442,30 +442,37 @@ def run_graph_to_completion(
                         raise e
                     task_futures[next_task] = submitted
                 else:
-                    # Whoops, back on the queue
-                    # We should probably wait a bit here, but for now we're going to keep
-                    # burning through
+                    # TODO: Investigate a backoff strategy, now for add back on the queue
                     execution_state.reject_task(task_to_reject=next_task)
-            # update all the tasks in flight
-            # copy so we can modify
+
+            # Update all the tasks in flight (copy so we can modify)
             for task, task_future in task_futures.copy().items():
+                result, error = None, None
                 state = task_future.get_state()
-                result = task_future.get_result()
-                execution_state.update_task_state(task.task_id, state, result)
-                if TaskState.is_terminal(state):
-                    if task.adapter.does_hook("post_task_resolution", is_async=False):
-                        task.adapter.call_all_lifecycle_hooks_sync(
-                            "post_task_resolution",
-                            run_id=task.run_id,
-                            task_id=task.task_id,
-                            nodes=task.nodes,
-                            success=state == TaskState.SUCCESSFUL,
-                            error=None,  # TODO -- we could get the error from the task future
-                            result=result,
-                            spawning_task_id=task.spawning_task_id,
-                            purpose=task.purpose,
-                        )
-                    del task_futures[task]
+                try:
+                    result = task_future.get_result()
+                except Exception as e:
+                    logger.exception(
+                        f"Exception resolving task {task.task_id}, with nodes: "
+                        f"{[item.name for item in task.nodes]}"
+                    )
+                    raise e
+                finally:
+                    execution_state.update_task_state(task.task_id, state, result)
+                    if TaskState.is_terminal(state):
+                        if task.adapter.does_hook("post_task_resolution", is_async=False):
+                            task.adapter.call_all_lifecycle_hooks_sync(
+                                "post_task_resolution",
+                                run_id=task.run_id,
+                                task_id=task.task_id,
+                                nodes=task.nodes,
+                                success=state == TaskState.SUCCESSFUL,
+                                error=error,
+                                result=result,
+                                spawning_task_id=task.spawning_task_id,
+                                purpose=task.purpose,
+                            )
+                        del task_futures[task]
         logger.info(f"Graph is done, graph state is {execution_state.get_graph_state()}")
     finally:
         execution_manager.finalize()
--- a/hamilton/execution/executors.py
+++ b/hamilton/execution/executors.py
@@ -464,9 +464,9 @@ def run_graph_to_completion(
                 finally:
                     execution_state.update_task_state(task.task_id, state, result)
                     if TaskState.is_terminal(state):
-                        if task.adapter.does_hook("post_task_resolution", is_async=False):
+                        if task.adapter.does_hook("post_task_return", is_async=False):
                             task.adapter.call_all_lifecycle_hooks_sync(
-                                "post_task_resolution",
+                                "post_task_return",
                                 run_id=task.run_id,
                                 task_id=task.task_id,
                                 nodes=task.nodes,
--- a/hamilton/execution/executors.py
+++ b/hamilton/execution/executors.py
@@ -10,7 +10,7 @@ else:
     from concurrent.futures.process import ProcessPoolExecutor
 
 from concurrent.futures import Executor, Future, ThreadPoolExecutor
-from typing import Any, Callable, Dict, List
+from typing import Any, Callable, Dict, List, Protocol
 
 from hamilton import node
 from hamilton.execution.graph_functions import execute_subdag
@@ -20,13 +20,17 @@ from hamilton.execution.state import ExecutionState, GraphState, TaskState
 logger = logging.getLogger(__name__)
 
 
-@dataclasses.dataclass
-class TaskFuture:
+class TaskFuture(Protocol):
     """Simple representation of a future. TODO -- add cancel().
     This a clean wrapper over a python future, and we may end up just using that at some point."""
 
-    get_state: Callable[[], TaskState]
-    get_result: Callable[[], Any]
+    def get_state(self) -> TaskState:
+        """Returns the state of the task."""
+        ...
+
+    def get_result(self) -> Any:
+        """Returns the result of the task."""
+        ...
 
 
 class TaskExecutor(abc.ABC):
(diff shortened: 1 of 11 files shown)


cswartzvi:
Ahh - used `str | None`

cswartzvi:
Regarding the failing doc test - possibly related to `ddtrace` 3.1.0 update, in particular https://github.com/DataDog/dd-trace-py/pull/12186

skrawcz:
What's the impact here to the Hamilton Tracker? Should we implement these new functions for it?

cswartzvi:
@skrawcz,

I am not that familiar with the `HamiltonTracker`, but it appears to implement, mainly, node-based hooks. This change should not alter their function - node-based hooks should run as before. 

Offhand, do you know if the UI has any issues tracking out-of-process execution? I see that `HamiltonTracker` is stateful and I don't see any evident `__getstate__` or `__setstate__` methods, but I could be missing something. If there are issues, these hooks might be of some use - I would need to experiment.

skrawcz:
> I am not that familiar with the `HamiltonTracker`, but it appears to implement, mainly, node-based hooks. This change should not alter their function - node-based hooks should run as before.

👍 

> 
> Offhand, do you know if the UI has any issues tracking out-of-process execution? I see that `HamiltonTracker` is stateful and I don't see any evident `__getstate__` or `__setstate__` methods, but I could be missing something. If there are issues, these hooks might be of some use - I would need to experiment.

Yes, that's in the client that sends the events.

elijahbenizzy:
> > I am not that familiar with the `HamiltonTracker`, but it appears to implement, mainly, node-based hooks. This change should not alter their function - node-based hooks should run as before.
> 
> 👍
> 
> > Offhand, do you know if the UI has any issues tracking out-of-process execution? I see that `HamiltonTracker` is stateful and I don't see any evident `__getstate__` or `__setstate__` methods, but I could be missing something. If there are issues, these hooks might be of some use - I would need to experiment.
> 
> Yes, that's in the client that sends the events - need to go one level deeper in the tracker to the client it uses.

To chime in -- I don't think this should ipmact the hamilton tracker. That said, we may want to consider adding more to it to leverage these -- we could get a more fine-grained view/get expose more descriptive states.

skrawcz:
Cool. Does it makes sense to provide some implementation of these hooks in this PR too? e.g. some example using it?

cswartzvi:
@skrawcz - I was actually planning to create a second PR with a dedicated logging adapter (based on a side conversation with @elijahbenizzy) using these new hooks. I can add that here if you'd like.

skrawcz:
> @skrawcz - I was actually planning to create a second PR with a dedicated logging adapter (based on a side conversation with @elijahbenizzy) using these new hooks. I can add that here if you'd like.

Cool. Yeah I say having an implementation would be good because having something implemented helps prove / ground the API :) It can be in a PR off of this branch or together.

cswartzvi:
So, the logging adapter is a bit more expansive than I originally thought. We will probably want to have a discussion. I will open a separate PR based off this branch - hopefully this evening.

hamilton/execution/executors.py
```diff
@@ -396,20 +407,32 @@ def run_graph_to_completion(
                             f"{[item.name for item in next_task.nodes]}"
                         )
                         raise e
-                    task_futures[next_task.task_id] = submitted
```
elijahbenizzy, line 399: Why the change here? Assuming `Task` is hashable? Don't see it in the spec...

elijahbenizzy: Oh, see you changed it... Got it, this allows us to have the extra data available...

tests/lifecycle/test_lifecycle_adapters_end_to_end_task_based.py
```diff
@@ -242,6 +246,72 @@ def test_individual_post_task_expand_hook():
...
+    hook_name = "pre_task_submission"
+    hook = TrackingPreTaskSubmissionHook(name=hook_name)
+    dr = _sample_driver(hook)
+    dr.execute(["output"], inputs={"n_iters_input": 5})
+    relevant_calls = [item for item in hook.calls if item.name == hook_name]
+    assert len(relevant_calls) == 10
+    spawning_task_ids = Counter([item.bound_kwargs["spawning_task_id"] for item in relevant_calls])
+    assert spawning_task_ids == {"expand-parallel_over": 5, None: 5}
+    purposes = Counter([item.bound_kwargs["purpose"] for item in relevant_calls])
+    assert purposes == {
```
elijahbenizzy, line 259: Comments on this might be nice -- E.G. why these are the call counts

elijahbenizzy: (although I realize that you probably copied this approach from others which don't document...)

elijahbenizzy (commented):
Looking good -- a few minor comments

hamilton/execution/executors.py
cswartzvi: Yup that was what I was going for! I needed to access the `adapater` for the task and thought adding a parallel structure was a little redundent. Do you see any problems with making `TaskImplementation` hashable?

tests/lifecycle/test_lifecycle_adapters_end_to_end_task_based.py
cswartzvi: Good idea! It took me a few minutes to re-figure out what the counts were looking at them now - a sure sign they are need. Adding now ...

cswartzvi:
Summary of the latest changes (I have been sitting on them)...

- Made `TaskFuture` a protocol, added a `TaskFutureWrappingFunction` implementation - this was done so that errors could be handled correctly from synchronous tasks
- Cleaned up some of the exception handling at submission and resolution calling site
- Various other quality of life improvements

cswartzvi:
Hmm ... tests are failing because of `xgboost`. Not sure if this was something I broke.

hamilton/execution/executors.py
```diff
@@ -162,6 +166,36 @@ def base_execute_task(task: TaskImplementation) -> Dict[str, Any]:
...
+
+    def __init__(self, function: Callable[[], Any]):
+        self.function = function
+        self._results = None
+        self._done = False
+        self._exception = None
+
+    def get_state(self):
+        try:
+            self._results = self.function()
```
elijahbenizzy, line 181: Shouldn't this only call the function once? E.G. it should be cached?

hamilton/execution/executors.py
```diff
@@ -172,9 +206,7 @@ def submit_task(self, task: TaskImplementation) -> TaskFuture:
         :param task: Task to submit
         :return: Future associated with this task
         """
-        # No error management for now
-        result = base_execute_task(task)
-        return TaskFuture(get_state=lambda: TaskState.SUCCESSFUL, get_result=lambda: result)
+        return TaskFutureWrappingFunction(functools.partial(base_execute_task, task))
```
elijahbenizzy, line 209: Does this add error management?

hamilton/execution/executors.py
```diff
@@ -383,11 +416,22 @@ def run_graph_to_completion(
...
+                            run_id=next_task.run_id,
+                            task_id=next_task.task_id,
+                            nodes=next_task.nodes,
+                            inputs=next_task.dynamic_inputs,
+                            overrides=next_task.overrides,
+                            spawning_task_id=next_task.spawning_task_id,
+                            purpose=next_task.purpose,
+                        )
                     try:
                         submitted = task_executor.submit_task(next_task)
```
elijahbenizzy, line 436: Not now but at some point a post_task_submission hook might make sense.

hamilton/lifecycle/api.py
```diff
@@ -371,10 +373,121 @@ def run_after_graph_execution(
...
+        :param overrides: Overrides passed to the task
+        :param spawning_task_id: ID of the task that spawned this task
+        :param purpose: Purpose of the current task group
+        :param future_kwargs: Reserved for backwards compatibility.
+        """
+        pass
+
+
+class TaskResolutionHook(BasePostTaskResolution, abc.ABC):
+    """Implement this to hook into the task resolution process. Tasks are submitted to an executor,
```
elijahbenizzy, line 431: Let's clarify a little more what task "resolution" is for users -- it's nuanced and a bit confusing.

IIUC this is what happens when we have a task "template", which gets realized/ "bound" to tasks. E.G. when all the dependencies are satisfied and we can enqueue it. Right?

elijahbenizzy (commented):
Looks good! A few small changes but I think it's almost there.

hamilton/execution/executors.py
cswartzvi: Yup! This was an oversight. Commit inbound.

hamilton/lifecycle/api.py
cswartzvi: I struggled with the name of this hook/adapter. Currently the `post_task_resolution` hook gets called in `run_graph_to_completion` after the task future gets resolved. Maybe it's better to describe it as 'a hook that fires after a task returns from an executor'? Ha - that is actually a better complement to `pre_task_submission`! What do you think - I would happy to rename the hook/adapter - maybe `BasePostTaskReturn` / `TaskReturnnHook`?

hamilton/execution/executors.py
cswartzvi: Yeah that is a good idea 💯

hamilton/execution/executors.py
cswartzvi: I believe that I removed that comment prematurely. There is no direct error management in `TaskFutureWrappingFunction`. However, there is now error management around the `post_task_resolution` hook in `run_graph_to_completion` (it gets sent to the hook before propagating upward). I would be happy to add the comment back if you like!

hamilton/lifecycle/api.py
elijahbenizzy: Yeah I like `post_task_return` and `TaskReturnHooks` -- I think that's a bit clearer.

elijahbenizzy (approved):
Thanks! Sorry this took a bit to review. Looks good, just name change the hook and ping me, I'll merge.

cswartzvi:
@elijahbenizzy I changed the name of the hook and merged in main (probably should have rebased).

Issue #1297, opened 2025-03-25
hamilton.function_modifiers.datasaver does not work with __future__.annotations


kreczko:
I am not sure if this is expected, I could not find a related issue (apologies if it exists).
At the moment it is not possible to use `from __future__ import annotations` and `hamilton.function_modifiers.datasaver` in conjunction. Since having the annotations included is a default in e.g. `ruff` (see https://docs.astral.sh/ruff/rules/future-required-type-annotation/), it can lead to friction.

# Current behavior
Using `from __future__ import annotations` breaks datasaver usage.
The internal test `if return_annotation not in (dict, Dict)` fails, since the `return_annotation` becomes a string in this scenario

## Stack Traces
```
Traceback (most recent call last):
  File "~/playground/hamilton_saver.py", line 6, in <module>
    @datasaver()  # you need ()
     ^^^^^^^^^^^
  File "~/.venv/lib/python3.11/site-packages/hamilton/function_modifiers/base.py", line 60, in replace__call__
    return call_fn(self, fn)
           ^^^^^^^^^^^^^^^^^
  File "~/.venv/lib/python3.11/site-packages/hamilton/function_modifiers/base.py", line 102, in __call__
    self.validate(fn)
  File "~/.venv/lib/python3.11/site-packages/hamilton/function_modifiers/adapters.py", line 864, in validate
    raise InvalidDecoratorException(f"Function: {fn.__qualname__} must return a dict.")
hamilton.function_modifiers.base.InvalidDecoratorException: Function: save_json_data must return a dict.
```

## Steps to replicate behavior
1. Take datasaver example from https://hamilton.dagworks.io/en/latest/reference/decorators/datasaver/
2. Add `from __future__ import annotations` at the top of the file

## Library & System Information
Ubuntu 24.04
Python 3.11.11
hamilton 1.86.1

# Expected behavior
datasaver should work with `__future__.annotations` ?

skrawcz:
@kreczko thanks for flagging. Please feel free to submit a PR with the fix!

kreczko:
@skrawcz done. Not quite sure if this will work for you, but happy to adjust if needed

skrawcz:
Closed — fixed in PR #1566.

PR #1298, opened 2025-03-28
Fix: issue #1297: datasaver does not work with __future__.annotations

kreczko:
Fixes issue #1297 by changing return_annotation retrieval `function_modifiers.adapters.datasaver.validate`.

## Changes
- modified `datasaver.validate` to check against `__future__.annotations` compatible type hints

## How I tested this
- added `from __future__ import annotations` to test_adapters
- ran `pytest tests/function_modifiers` &rarr; test failed
- made above changes
- ran `pytest tests/function_modifiers` &rarr; test succeeded

## Notes
<s>
`correct_ds_function` in `test_adapters.py` uses an alias for `dict`: `dict_ = dict`.
However, for the string comparison this is not resolved to `dict`; added `dict_` to the validation for now.

I do not understand why this is done in the tests and should, in theory, not impact "normal" use.
Feedback on this is welcome.
</s>

## Checklist

### DIFF
--- a/hamilton/function_modifiers/adapters.py
+++ b/hamilton/function_modifiers/adapters.py
@@ -878,7 +878,7 @@ class datasaver(NodeCreator):
                 f"Function: {fn.__qualname__} must have a return annotation."
             )
         origin = typing.get_origin(return_annotation)
-        if return_annotation is not dict and origin is not dict:
+        if return_annotation not in (dict, Dict, "dict", "dict_") and origin not in (dict, Dict, "dict", "dict_"):
             raise InvalidDecoratorException(f"Function: {fn.__qualname__} must return a dict.")
 
     def generate_nodes(self, fn: Callable, config) -> list[node.Node]:
--- a/tests/function_modifiers/test_adapters.py
+++ b/tests/function_modifiers/test_adapters.py
@@ -14,6 +14,7 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
+from __future__ import annotations
 
 import dataclasses
 from collections import Counter


kreczko:
The bot makes a good suggestion: instead of my change, it is better to change
`# return_annotation = inspect.signature(fn).return_annotation` to `return_annotation = typing.get_type_hints(fn)["return"]`
which does resolve `"dict_"` and `"dict"` to `dict`.
As a result, the block `if return_annotation is inspect.Signature.empty` can be replaced with `if return_annotation is None`.
I assume that's preferable and make the change

elijahbenizzy:
Thank you! This is great -- I've been meaning to make this change to everywhere in the codebase and new spots keep popping up in which it's needed. 

Just checking -- does the change to `test_adapters` mean that this won't work if we don't use `from __future__ import annotations`?

kreczko:
> Just checking -- does the change to `test_adapters` mean that this won't work if we don't use `from __future__ import annotations`?

Good point, did not test this.

```diff
--- a/tests/function_modifiers/test_adapters.py
+++ b/tests/function_modifiers/test_adapters.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
```

```bash
pytest -s tests/function_modifiers/
======================================= 383 passed, 13 warnings in 3.21s =======================================
```

luckily same result.

Out of curiosity, I tried this example:
```python
class B(dict):

    pass

def func() -> B:
    return B()

def func2() -> dict:
    return dict()

import inspect
import typing


print(inspect.signature(func).return_annotation)
print(inspect.signature(func2).return_annotation)

print(typing.get_type_hints(func)["return"])
print(typing.get_type_hints(func2)["return"])
```

without annotations they are identical
```
inspect: <class '__main__.B'>
inspect: <class 'dict'>
typing: <class '__main__.B'>
typing: <class 'dict'>
```
with `from __future__ import annotations` the inspect ones turn to string:
```
inspect: B
inspect: dict
typing: <class '__main__.B'>
typing: <class 'dict'>
```

Hope this helps

skrawcz:
@elijahbenizzy  **bump**

elijahbenizzy (approved):
Thanks!

skrawcz:
ugh, we need to rebase this and resolve conflicts. @kreczko do you have time?

kreczko:
@skrawcz done.
Note that in between creating this PR and now `return_annotation = typing.get_type_hints(fn).get("return")` has been introduced in 3050c12a8ec055a703cc375069dcd91bfb1b3c36, which is the fix for the issue reported.
I don't think this PR is needed anymore:
```diff
diff --git a/hamilton/function_modifiers/adapters.py b/hamilton/function_modifiers/adapters.py
index daa45a24..53f8c4d1 100644
--- a/hamilton/function_modifiers/adapters.py
+++ b/hamilton/function_modifiers/adapters.py
@@ -878,7 +878,7 @@ class datasaver(NodeCreator):
                 f"Function: {fn.__qualname__} must have a return annotation."
             )
         origin = typing.get_origin(return_annotation)
-        if return_annotation not in (dict, Dict, "dict", "dict_") and origin not in (dict, Dict, "dict", "dict_"):
+        if return_annotation is not dict and origin is not dict:
             raise InvalidDecoratorException(f"Function: {fn.__qualname__} must return a dict.")
 
     def generate_nodes(self, fn: Callable, config) -> list[node.Node]:
diff --git a/tests/function_modifiers/test_adapters.py b/tests/function_modifiers/test_adapters.py
index bb07fbcc..fd4c0029 100644
--- a/tests/function_modifiers/test_adapters.py
+++ b/tests/function_modifiers/test_adapters.py
@@ -14,7 +14,6 @@
 # KIND, either express or implied.  See the License for the
 # specific language governing permissions and limitations
 # under the License.
-from __future__ import annotations
 
 import dataclasses
 from collections import Counter

```
Feel free to close

skrawcz:
> @skrawcz done. Note that in between creating this PR and now `return_annotation = typing.get_type_hints(fn).get("return")` has been introduced in [3050c12](https://github.com/apache/hamilton/commit/3050c12a8ec055a703cc375069dcd91bfb1b3c36), which is the fix for the issue reported. I don't think this PR is needed anymore:
> 
> ```diff
> diff --git a/hamilton/function_modifiers/adapters.py b/hamilton/function_modifiers/adapters.py
> index daa45a24..53f8c4d1 100644
> --- a/hamilton/function_modifiers/adapters.py
> +++ b/hamilton/function_modifiers/adapters.py
> @@ -878,7 +878,7 @@ class datasaver(NodeCreator):
>                  f"Function: {fn.__qualname__} must have a return annotation."
>              )
>          origin = typing.get_origin(return_annotation)
> -        if return_annotation not in (dict, Dict, "dict", "dict_") and origin not in (dict, Dict, "dict", "dict_"):
> +        if return_annotation is not dict and origin is not dict:
>              raise InvalidDecoratorException(f"Function: {fn.__qualname__} must return a dict.")
>  
>      def generate_nodes(self, fn: Callable, config) -> list[node.Node]:
> diff --git a/tests/function_modifiers/test_adapters.py b/tests/function_modifiers/test_adapters.py
> index bb07fbcc..fd4c0029 100644
> --- a/tests/function_modifiers/test_adapters.py
> +++ b/tests/function_modifiers/test_adapters.py
> @@ -14,7 +14,6 @@
>  # KIND, either express or implied.  See the License for the
>  # specific language governing permissions and limitations
>  # under the License.
> -from __future__ import annotations
>  
>  import dataclasses
>  from collections import Counter
> ```
> 
> Feel free to close

thank you for updating and checking!

PR #1303, opened 2025-04-06
Add new function modifier `upack_fields`

cswartzvi:
Added a new modifier called `unpack_fields` which allows for the extraction of fields from a tuple output (a cross between `extract_columns` and `extract_fields`).

## Changes

* Added `unpack_fields` decorator to `hamilton/function_modifiers/__init__.py` and implemented its logic in `hamilton/function_modifiers/expanders.py`. This decorator enables the extraction of fields from a tuple output, expanding a single function into multiple nodes.

## How I tested this

* Added multiple test cases in `tests/function_modifiers/test_expanders.py` to validate the functionality of the `unpack_fields` decorator, including tests for valid and invalid type annotations and different tuple configurations.

## Notes

- Example usage with an fixed size tuple:
  ```python
    @unpack_fields("X_train" "X_validation", "X_test")
    def dataset_splits(X: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
        """Randomly split data into train, validation, test"""
        X_train, X_validation, X_test = random_split(X)
        return X_train, X_validation, X_test
  ```

- Example usage with a subset of fixed length tuples:
  ```python
    @unpack_fields("X_train")
    def dataset_splits(X: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
        """Randomly split data into train, validation, test"""
        X_train, X_validation, X_test = random_split(X)
        return X_train, X_validation, X_test
  ```

- Example usage with an indeterminate length tuple:
  ```python
    @unpack_fields("X_train" "X_validation", "X_test")
    def dataset_splits(X: np.ndarray) -> Tuple[np.ndarray, ...]:
        """Randomly split data into train, validation, test"""
        X_train, X_validation, X_test = random_split(X)
        return X_train, X_validation, X_test
  ```

## Checklist


hamilton/function_modifiers/__init__.py
```diff
@@ -59,6 +59,7 @@
 # Extract decorators
 extract_columns = expanders.extract_columns
 extract_fields = expanders.extract_fields
+unpack_fields = expanders.unpack_fields
```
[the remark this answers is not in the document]
cswartzvi: This has been done - added to both the "concepts" and "reference" sections

tests/function_modifiers/test_expanders.py
```diff
@@ -468,6 +471,135 @@ def dummy_dict() -> dict:
...
+        callabl=dummy,
+        tags={"module": "tests.function_modifiers.test_expanders"},
+    )
+    assert nodes[1].name == "A"
+    assert nodes[1].type == int
+    assert nodes[1].documentation == "dummy doc"
+    assert nodes[1].input_types == {dummy.__name__: (Tuple[int, str, int], DependencyType.REQUIRED)}
+
+
+def test_upack_fields_valid_indeterminate_tuple():
```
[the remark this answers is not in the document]
cswartzvi: Good bot - embarrassing human

cswartzvi:
Geez - apparently I have a problem remembering to include backward compatible type hints 🙃. Regardless, this is ready to be looked at!

docs/concepts/function-modifiers.rst
```diff
@@ -191,21 +191,44 @@ Sometimes, your node outputs multiple values that you would like to name and mak
 
     To add metadata to extracted nodes, use ``@tag_output``, which works just like ``@tag``.
 
+@unpack_fields
```
elijahbenizzy, line 194: This makes for very elegant code. Nice work!

hamilton/function_modifiers/expanders.py
```diff
@@ -869,6 +874,156 @@ def extractor_fn(
...
+    if base_type != tuple and base_type != Tuple:
+        return
+
+    output_args = typing_inspect.get_args(output_type)
+    num_ellipsis = output_args.count(Ellipsis)
+    if num_ellipsis > 1:
+        raise base.InvalidDecoratorException(
+            f"Invalid tuple: Found more than one ellipsis ('...'): {output_type}"
+        )
+    elif num_ellipsis == 1:
```
elijahbenizzy, line 895: What happens if it is `tuple[..., int]` (which, addmittedly, isn't really logical...)

hamilton/function_modifiers/expanders.py
```diff
@@ -869,6 +874,156 @@ def extractor_fn(
...
+
+    @override
+    def validate(self, fn: Callable):
+        output_type = typing.get_type_hints(fn).get("return")
+        field_types = _process_unpack_fields(self.fields, output_type)
+        if field_types:
+            self.field_types = field_types
+            self.output_type = output_type
+        else:
+            message = (
```
elijahbenizzy, line 960: Feel like `_process_unpack_fields` should raise an error in this case? Is there a valid reason to have an empty return?

hamilton/function_modifiers/expanders.py
```diff
@@ -869,6 +874,156 @@ def extractor_fn(
...
+                tuple_generated = fn(*args, **kwargs)
+                return tuple_generated
+
+        output_nodes = [node_.copy_with(callabl=tuple_generator)]
+
+        for idx, (field_name, field_type) in enumerate(zip(self.fields, self.field_types)):
+            # NOTE: The extractors, as defined below, are constructed to avoid closure issues.
+            if inspect.iscoroutinefunction(fn):
+
+                async def extractor(field_index: int = idx, **kwargs) -> field_type:  # type: ignore
```
elijahbenizzy, line 994: Is there any reason to have an async extractor? It'll be the same either way as `await` is never called...

elijahbenizzy (commented):
This is awesome! Thank you! I'm really looking forward to shipping this. Some small comments.

hamilton/function_modifiers/expanders.py
skrawcz: To handle this case, yeah?
```python

@unpack_fields
async def my_function(...) -> Tuple[..., int]:
   ...
 ```
 Otherwise all async functions should have an `await` in them though right?

Comment (elijahbenizzy) [reply hamilton/function_modifiers/expanders.py:994]: The extractor function goes after it, it shouldn't need to be async, as it doesn't do anything asynchronous. The function itself is async (which forms a node), but the extractor doesn't need to be.

File: hamilton/function_modifiers/expanders.py (hunk 4 above)
Comment (cswartzvi) [reply hamilton/function_modifiers/expanders.py:895]: Ah yes, good catch, I will add that!

I was actually trying to include some more tests for cases like `tuple[...]`, `tuple[int, int, ...]` but I was having some issues with backward compatibility. I will take another crack at it.

File: hamilton/function_modifiers/expanders.py (hunk 5 above)
Comment (cswartzvi) [reply hamilton/function_modifiers/expanders.py:960]: You know what, originally I didn't want to pass `output_type` to `_process_unpack_fields` and this was my way of still producing a nice error message. However, now that I changed my mind about `output_type` it's definitely better to raise the error in `_process_unpack_fields` - it's cleaner that way.

File: hamilton/function_modifiers/expanders.py (hunk 6 above)
Comment (cswartzvi) [reply hamilton/function_modifiers/expanders.py:994]: Ah, ya got me, I was copying parts of `extract_fields` and didn't notice that there was no `await` in the extractor. I don't think we need it considering (as @elijahbenizzy said) it's only indexing the result of `tuple_generator` (either sync or async). Commit incoming.

**Comment by cswartzvi** [comment]
Wooh boy, TIL that improper use of an Ellipsis in `typing.Tuple` (`Tuple[..., int]`, `Tuple[int, int, ...]`, etc.) is an **error** prior to Python 3.11, even though it's perfectly fine in `tuple` after python 3.8 (`tuple[..., int]`, `tuple[int, int, ...]`, etc.) 🤷🏻‍♂️🤷🏻‍♂️. I have greatly expanded tests of `tuple` vs `typing.Tuple` in both valid and invalid cases using a little trick with `eval`. Note that this was the only way I could come up with where earlier python versions didn't complain about the use of `tuple[...]`. Please let me know if it's too ... crazy.

**Comment by cswartzvi** [comment]
Just wanted to mention one other thing ... in [slack](https://hamilton-opensource.slack.com/archives/C03MANME6G5/p1744055165809069?thread_ts=1743736065.053229&cid=C03MANME6G5) @zilto mentioned that

> ' .. with `tuple[str, ...]` the `@unpack_fields` should be authoritative for the number of expected tuple values ...'. 

Currently, `@unpack_fields` is extracting the first $n$ fields of an $m$ length tuple with only a check for $0 \leq n < m$ in the local `extractor` function. I could make the (sync/async) `tuple_generator` check that the requested fields match the ouput, but I wanted to get everyone's take first.

Sorry @zilto, I didn't see your response until after I started pushing commits!

**Comment by zilto** [comment]
@cswartzvi Great job on the PR! It makes me very happy to see docs pages included :smile: 

Regarding tuple length, I see the merit of your approach and implementation. For example, in this snippet:
```python
  @unpack_fields("X_train")
  def dataset_splits(X: np.ndarray) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
      """Randomly split data into train, validation, test"""
      X_train, X_validation, X_test = random_split(X)
      return X_train, X_validation, X_test
```
It seems reasonable to allow `@unpack_fields("X_train")` to only unpack `X_train`, the first element of the tuple. This matches how `@extract_fields` can selectively extract keys of a dict. It avoids creating unnecessary nodes and the node `dataset_splits` allows to get the full tuple.

elijahbenizzy (approved):
Looks good!

PR #1312, opened 2025-05-19
Create .asf.yaml

pjfanning:
Used to control the GitHub project because the 'Settings' page is not accessible for ASF projects

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,65 @@
+# https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features
+
+github:
+  description: "Apache Hamilton helps data scientists and engineers define testable, modular, self-documenting dataflows, that encode lineage/tracing and metadata. Runs and scales everywhere python does."
+  homepage: https://hamilton.apache.org/
+  labels:
+    - python
+    - data-science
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -0,0 +1,65 @@
...
+        # contexts are the names of checks that must pass
+        # contexts:
+      required_pull_request_reviews:
+        dismiss_stale_reviews: false
+        require_code_owner_reviews: false
+        required_approving_review_count: 1
+
+notifications:
+  commits:              commits@hamilton.apache.org
+  issues:               dev@hamilton.apache.org
```
pjfanning, line 62: I prefer to have a dedicated notifications mailing list instead of having every GitHub event mailed ti the dev list - it overwhelms all the other traffic on the dev list.

skrawcz: sure - so `notifications@` ?

pjfanning: That's what we use in Apache Pekko. I'll start a discussion on the dev mailing list.

pjfanning: I created the notifications mailing list and have changed this PR accordingly.

skrawcz (approved):

Issue #1313, opened 2025-05-19
Set up CI on GitHub


pjfanning:
Most new ASF projects use GitHub for CI. It is especially useful to have CI jobs to validate PRs.

zilto:
To add, it would be a good time to adopt `uv`. It should reduce time to install Python dependencies, offering a noticeable speedup for plugin tests

skrawcz:
yeah we should be able to migrate off of circleci relatively easily. LLMs should be able to help convert the circleci config to github ones pretty easily.

PR #1342, opened 2025-06-14
Migrate to GitHub acitions

jernejfrank:
Addressing #1313 and uses uv.

I don't have access to CircleCI to see the settings there (I assume the config with jobs and worklof is only part of it, since there's no mention of OS inside), would be good to let me know if anything else needs to be configured for the github actions workflow.

## Changes
- replaced circleci config with github actions
- minor linting and bugfixes

## How I tested this
- Need to invoke CI from main repo

## Notes
- Fail-fast is disabled for now, so that I can see all of the tests in the matrix failing
- Not sure if we need to test on all platforms
- Adds the option to run the tests manually through the actions tab
- some test (mostly plugins)  were running only on specific python versions and run now on all supported python versions of the plugin package:
  - Ray, Dask, PySpark 3.9-3.12
  - pre-commit 3.8-3.12 (sometimes important to spot things like needing `from __future__ import annotations`)
  - pandas, polars, narwhals 3.8-3.12
  - Vaex 3.8-3.10

## Checklist


skrawcz:
> I don't have access to CircleCI to see the settings there (I assume the config with jobs and worklof is only part of it, since there's no mention of OS inside), would be good to let me know if anything else needs to be configured for the github actions workflow.

@jernejfrank you can see everything in the .circleci yaml and what it references. It's similar to github workflows, but we have more flexibility on when a workflow is triggered with the github way. So shouldn't be hard to replicate.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
 ]
 dependencies = [
-    "numpy",
+    "numpy<2",
```
skrawcz: this isn't good since 2.0 has been out for a while.

pyproject.toml
```diff
@@ -58,7 +57,8 @@ docs = [
   # required for all the plugins
   "dlt",
   # furo -- install from main for now until the next release is out:
-  "furo @ git+https://github.com/pradyunsg/furo@main",
+  "furo",
```
skrawcz, line 59: +1

skrawcz:
The SDK tests aren't passing because spark changed things. I think we just need to update the tests...

jernejfrank:
> > I don't have access to CircleCI to see the settings there (I assume the config with jobs and worklof is only part of it, since there's no mention of OS inside), would be good to let me know if anything else needs to be configured for the github actions workflow.
> 
> @jernejfrank you can see everything in the .circleci yaml and what it references. It's similar to github workflows, but we have more flexibility on when a workflow is triggered with the github way. So shouldn't be hard to replicate.

@skrawcz Ah ok, then all is migrated! I thought there might be something more since the .circleci yaml doesn't mention anything about OS to test on, etc..

Also added this to the notes of the pr:
- some test (mostly plugins)  were running only on specific python versions and run now on all supported python versions of the plugin package:
  - Ray, Dask, PySpark 3.9-3.12
  - pre-commit 3.8-3.12 (sometimes important to spot things like needing `from __future__ import annotations`)
  - pandas, polars, narwhals 3.8-3.12
  - Vaex 3.8-3.10


Will be fixing broken tests, but may take a while!

pyproject.toml
jernejfrank: Forgot to remove when debugging some other dependencies issues. Removed now.

jernejfrank:
Right, so SDK was passing but now the CI dissapeared. Any thoughts @skrawcz?

I am also thinking of limiting it to Linux (for now) since I saw that windows has some issues with package installation when using `uv` since it does a better job at finding problems than `pip`.

skrawcz:
> Right, so SDK was passing but now the CI dissapeared. Any thoughts @skrawcz?

Can you rebase against main? (you might want to squash your commits to make that easier to do if it makes sense to squash some of them)

> I am also thinking of limiting it to Linux (for now) since I saw that windows has some issues with package installation when using uv since it does a better job at finding problems than pip.

That's fine I think.

jernejfrank:
@skrawcz ok, got the tests sorted out and did a rebase. Let me know if you're happy with squash merge or need me to organize the commits :)

.github/workflows/hamilton-main.yml
```diff
@@ -0,0 +1,163 @@
...
+        - 'plugin_tests/**'
+        - 'tests/**'
+        - 'pyproject.toml'
+
+jobs:
+  test:
+    name: "Unit Tests"
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # will change this to true at the end, but want to see tests failing on all use cases
```
skrawcz: need to flip this?

skrawcz: actually I think we want this to be false always.

skrawcz (approved):
Let's squash merge this into a single commit. This will make things simpler to rebase that way ; I don't think we need to keep the individual commits, right?

I usually craft a commit message, and then leave the prior squashed commits below it!

PR #1354, opened 2025-07-06
add proper license

pjfanning:
this is the correct license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,13 +1,201 @@
-Copyright (c) 2025-present Apache Software Foundation
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
-   http://www.apache.org/licenses/LICENSE-2.0
+   1. Definitions.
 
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,13 +1,201 @@
-Copyright (c) 2025-present Apache Software Foundation
```
pjfanning, line 1: the copyright is in the notice file

LICENSE
```diff
@@ -1,13 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
skrawcz, line 189:
```suggestion
   Copyright 2025 Apache Software Foundation
```

pjfanning: we do not set this - please look at ASF projects and see we never set this - it is in the notice file

pjfanning: If you read the license - this is in the Appendix and is documentation not a copyright stmt.

skrawcz: Really? Did legal review that? It seems like the license is applied incorrectly otherwise -- but sure there would be duplication then.

skrawcz: Or it should say see NOTICE file ... or something...

skrawcz (approved):

Issue #1336, opened 2025-05-27
ASF First Release checklist


skrawcz:
- [ ] Migrate license to 2.0 (https://github.com/apache/hamilton/pull/1335)
- [ ] Are release files in correct location?

- [ ] Do release files have the word incubating in their name? (Note: only the release artifacts require "incubating" in their names, jar files (included in the binary artifact or deployed onto maven repository) are suggested but not required to have "incubating" as name suffix)

- [ ] Are the digital signature and hashes correct?

- [x] Does DISCLAIMER file exist?  (https://github.com/apache/hamilton/pull/1335)

- [x] Do LICENSE and NOTICE files exists? (https://github.com/apache/hamilton/pull/1335)

- [ ]  Is the LICENSE and NOTICE text correct?

- [x] Is the NOTICE year correct? (https://github.com/apache/hamilton/pull/1335)

- [ ] Un-included software dependencies are not mentioned in LICENSE or NOTICE?

- [ ]  License information is not mentioned in NOTICE?

- [ ] Do the contents of the release match with what's tagged in version control?

- [ ]  Are there any unexpected binary files in the release?

- [ ] Can you compile from source? Are the instruction clear?

## Is there any 3rd party code contained inside the release? If so:

- [ ] Does the software have a compatible license?

- [ ] Are all software licenses mentioned in LICENSE?

- [ ]  Is the full text of the licenses (or pointers to it) in LICENSE?

### Is any of this code Apache licensed? Do they have NOTICE files? If so:

- [ ] Have relevant parts of those NOTICE files been added to this NOTICE file?

- [ ] Do all source files have ASF headers?

pjfanning:
The binary artifacts listed in https://github.com/apache/hamilton/issues/1337 cannot be released until they have their contents reviewed (the `sf-*` artifacts listed there). They need:
* LICENSE, NOTICE and DISCLAIMER files that describe the contents of the binary artifacts (tailored to each binary artifact)
* Any 3rd party code or binaries included in the binary artifact needs to have a license compatible with the Apache one
* The licenses for these 3rd party code or binaries need to be included
* hopefully binary artifact names are renamed to include apache and possibly, incubating in their names
* all binary artifacts that Hamilton ships should be based on a source release - we can also do snapshot releases as long as they are badged as snapshot releases

skrawcz:
> LICENSE, NOTICE and DISCLAIMER files that describe the contents of the binary artifacts (tailored to each binary artifact)

[...]

PR #1374, opened 2025-09-03
Clean up repo

cswartzvi:
Hi all - I was looking through the codebase to help create the first ASF release (#1336) and thought we might benefit from a little clean-up. Please have a look and let me know if I went too far  with anything - happy to roll things back!

## Changes

- [x] Moved top-level images to new `assets` folder. Do we need to keep these?
- [x] Moved top-level md files to either `docs` or `writeups`. Maybe these should be fully integrated with the docs?
- [x] Deleted `setup.py`, removing the legacy Python packaging script and related logic
- [x] Remove unused configuration files (`.style.yapf`, `.flake8`, ...)
- [x] Update manifest (remove non-existent requirements files)
- [x] Update min required python to `3.8.1`. This allows contributors to use `uv sync` without allowing prereleases (as per the last meetup, I would like bump the min required python to `3.9` next)
- [x] Update `uv` configuration. Since `pip` now has support, I moved dev dependencies to `dependency-groups` (PEP 735)
- [x] Fix broken tests ~(maybe out of scope for this PR)~

## How I tested this

No changes to codebase

## Notes

N/A

## Checklist


cswartzvi:
@skrawcz - using a branch on the main repo (and not from my fork) seems to have fixed the issue with the failing docs action.

pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "sf-hamilton"
 dynamic = ["version"]
 description = "Hamilton, the micro-framework for creating dataframes."
 readme = "README.md"
-requires-python = ">=3.8, <4"
+requires-python = ">=3.8.1, <4"
```
jernejfrank, line 10: Just a heads up, that this PR requires using pip 25.1 which added installing groups, but also dropped support for python 3.8, so we should work towards droppng support for py38.

jernejfrank (commented):
Looks great, just a minor comment from my side.

pyproject.toml
cswartzvi: @jernejfrank, thanks for taking a look. I completely agree with you regarding dropping py38 support. Originally, I was going to include that in this PR, but it started to feel like mission creep. Changing the minimum to `3.8.1` was just enough to get `uv sync` working reliably without using prereleases (which was also causing issues).

I am going to open another PR after this one that drops py38 support - provided there are no objection from the other members of the PMC team.

jernejfrank: That sounds good to me, also feel it is better to separate it out into another PR!

skrawcz:
@cswartzvi please add a nice squash commit message to mention and relevant decisions you made here :)

PR #1376, opened 2025-09-06
feat: dynamically create `sf-hamilton-core` package

zilto:
Adds the directory `hamilton-core/` with mechanism to dynamically patch `hamilton/` and bundle a library named `sf-hamilton-core` which could be pushed to pypi.

It makes targeted 2 changes:
- disable plugin autoloading
- make `pandas` and `numpy` optional dependencies; and remove `networkx` dependency (currently unused).

This makes the Hamilton package a much lighter install and solves long library loading time. See the file `hamilton-core/README.md` for details

## Changes
- Add `hamilton-core/` directory
- `hamilton-core/setup.py` copies the code of `hamilton/` to `hamilton-core/hamilton/_hamilton`
- `hamilton-core/hamilton/__init__.py` is the entry point to `sf-hamilton-core`, which proxies everything directly to the source code of `hamilton` stored in `hamilton-core/hamilton/_hamilton`
- modify `hamilton/base.py` to lazily import `pandas` and `numpy`. This shouldn't affect users in any way

## How I tested this
- successfully ran all core unit tests locally
- added CI workflow that installs `sf-hamilton-core` and runs Hamilton's unit tests

## TODO
- remove networkx dependency
- add info to README about how it works
- clean up `hamilton-core/setup.py` and add linting + formatting
- potentially add docs pages

## Checklist

### DIFF
--- a/hamilton-core/README.md
+++ b/hamilton-core/README.md
@@ -1,0 +1,41 @@
+# Read carefully
+
+> Use at your own risk
+
+This directory contains code for the package `sf-hamilton-core`. It is a drop-in replacement of `sf-hamilton`, with two changes:
+- disable plugin autoloading
+- make `pandas` and `numpy` optional dependencies; and remove `networkx` dependency (currently unused).
+
+This makes the Hamilton package a much lighter install and solves long library loading time.
+
+## As a user
+If you want to try `sf-hamilton-core`, you need to:
+1. Remove your current Hamilton installation: `pip uninstall sf-hamilton`
+2. Install Hamilton core `pip install sf-hamilton-core`
+3. Check installation `pip list` should only include `sf-hamilton-core`.
+
+This will install a different Python package with the name `hamilton` with the smaller dependencies and plugin autoloading disabled.
+
+It should be a drop-in replacement and your existing Hamilton code should just work. Though, if you're relying on plugins (e.g., parquet materializers, dataframe result builders), you will need to manually load them.
+
+
+## How does it work
+
+
+## Why is another package `sf-hamilton` necessary
+This exists to prevent backwards incompatible changes for people who `pip install sf-hamilton` and use it in production. It is a temporary solution until a major release `sf-hamilton==2.0.0` could allow breaking changes and a more robust solution.
+
+### Disable plugin autoloading
+Hamilton has generous number of plugins (`pandas`, `polars`, `mlflow`, `spark`). To give a good user experience, Hamilton autoloads plugins based on the available Python libraries in the current Python environment. For example, `to.mlflow()` becomes available if `mlflow` is installed. Autoloaded features notably include materializers like `from_.parquet` and `to.parquet` and data validators (pydantic, pandera, etc.)
+
+The issue with this approach is that Python environment with a lot of dependencies, common in data science, can be very slow to start because of all the imports. Currently, Hamilton allows to disable autoloading via a user config or Python code. This require manual setups and is not the best default for some users.
+
+### `pandas` and `numpy` dependencies
+Hamilton was initially created for workflows that used `pandas` and `numpy` heavily. For this reason, `numpy` and `pandas` are imported at the top-level of module `hamilton.base`. Because of the package structure, as a Hamilton user, you're importing `pandas` and `numpy` every time you import `hamilton`.
+
+A reasonable change would be to move `numpy` and `pandas` to a "lazy" location. Then, dependencies would only be imported when features requiring them are used and they could be removed from `pyproject.toml`. Unfortunately, plugin autoloading defaults make this solution a significant breaking change and insatisfactory.
+
+Since plugins are loaded based on the Python package available, removing `pandas` and `numpy` would allow disable the loading of these plugins. This would break popular CSV and parquet materializers.
+
+### `networkx` dependency
+The `sf-hamilton[visualization]` extra currently includes `networkx` as a dependency, though it is never actually used. There's a single function requiring it and it could be implemented in pure Python. This has been made even easier with the addition of `graphlib` in the standard library in Python 3.9.
--- a/hamilton/base.py
+++ b/hamilton/base.py
@@ -19,15 +19,13 @@
 It should only import hamilton.node, numpy, pandas.
 It cannot import hamilton.graph, or hamilton.driver.
 """
+from __future__ import annotations
 
 import abc
 import collections
 import logging
-from typing import Any, Dict, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union, TYPE_CHECKING
 
-import numpy as np
-import pandas as pd
-from pandas.core.indexes import extension as pd_extension
 
 from hamilton.lifecycle import api as lifecycle_api
 
@@ -35,6 +33,11 @@ try:
     from . import htypes, node
 except ImportError:
     import node
+
+if TYPE_CHECKING:
+    import numpy as np
+    import pandas as pd
+
 
 logger = logging.getLogger(__name__)
 
@@ -120,6 +123,8 @@ class PandasDataFrameResult(ResultMixin):
         :param outputs: the dict we're trying to create a result from.
         :return: dict of all index types, dict of time series/categorical index types, dict if there is no index
         """
+        import pandas as pd
+
         all_index_types = collections.defaultdict(list)
         time_indexes = collections.defaultdict(list)
         no_indexes = collections.defaultdict(list)
@@ -131,6 +136,8 @@ class PandasDataFrameResult(ResultMixin):
 
         def get_parent_time_index_type():
             """Helper to pull the right time index parent class."""
+            from pandas.core.indexes import extension as pd_extension
+
             if hasattr(pd_extension, "NDArrayBackedExtensionIndex"):
                 index_type = pd_extension.NDArrayBackedExtensionIndex
             else:
@@ -220,6 +227,8 @@ class PandasDataFrameResult(ResultMixin):
 
         :param outputs: the outputs to build a dataframe from.
         """
+        import pandas as pd
+
         # TODO check inputs are pd.Series, arrays, or scalars -- else error
         output_index_type_tuple = PandasDataFrameResult.pandas_index_types(outputs)
         # this next line just log warnings
@@ -255,6 +264,7 @@ class PandasDataFrameResult(ResultMixin):
         :param outputs: The outputs to build the dataframe from.
         :return: A dataframe with the outputs.
         """
+        import pandas as pd
 
         def get_output_name(output_name: str, column_name: str) -> str:
             """Add function prefix to columns.
@@ -300,6 +310,7 @@ class PandasDataFrameResult(ResultMixin):
         return [Any]
 
     def output_type(self) -> Type:
+        import pandas as pd
         return pd.DataFrame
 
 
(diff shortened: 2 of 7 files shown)


hamilton-core/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The issue with this approach is that Python environment with a lot of dependencies, common in data science, can be very slow to start because of all the imports. Currently, Hamilton allows to disable autoloading via a user config or Python code. This require manual setups and is not the best default for some users.
+
+### `pandas` and `numpy` dependencies
+Hamilton was initially created for workflows that used `pandas` and `numpy` heavily. For this reason, `numpy` and `pandas` are imported at the top-level of module `hamilton.base`. Because of the package structure, as a Hamilton user, you're importing `pandas` and `numpy` every time you import `hamilton`.
+
+A reasonable change would be to move `numpy` and `pandas` to a "lazy" location. Then, dependencies would only be imported when features requiring them are used and they could be removed from `pyproject.toml`. Unfortunately, plugin autoloading defaults make this solution a significant breaking change and insatisfactory.
```
skrawcz, line 36:
```suggestion
A reasonable change would be to move `numpy` and `pandas` to a "lazy" location. Then, dependencies would only be imported when features requiring them are used and they could be removed from `pyproject.toml`. Unfortunately, plugin autoloading defaults make this solution a significant breaking change and unsatisfactory.
```

hamilton/base.py
```diff
@@ -20,21 +20,22 @@
...
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
 
+from hamilton import htypes
 from hamilton.lifecycle import api as lifecycle_api
 
-try:
-    from . import htypes, node
-except ImportError:
-    import node
+if TYPE_CHECKING:
```
skrawcz, line 33: comment as to importance of this

skrawcz (approved):
So does this copy everything in sf-hamilton? Including the `LICENSE` file and `NOTICE` files? We need a few things to always be there for Apache purposes. If so, we're good I think. If not you'll need to add them.

pjfanning:
I don't know much about releases to pypi, can I ask if it possible to sign the releases using something like https://www.python.org/downloads/metadata/sigstore/ or something similar?

If you sign the pypi release, could you sign it with the same key that you sign the ASF compliant source releases?

The signing keys for source releases need to be maintained on secure hardware and generally, most ASF releases are done by running jobs on a private laptop and signed with GPG (or equivalent).
https://infra.apache.org/release-signing.html

hamilton/base.py
zilto: Changes:
- I moved the imports `pandas`, `numpy` and `pandas.core.indexes.extension` from the top-level to the code path that actually use these dependencies. There should be no behavior change, but it allows to `import hamilton.base` without loading pandas each time.
- [if TYPE_CHECKING](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING) is the standard Python approach to import package that are only relevant for annotating signatures. For example, `mypy` will import `pandas` when doing type checking, but doing `from hamilton import base` won't import `pandas` 
- `pandas` and `numpy` are in the type checking block because they are used in some function signatures. `pandas.core.indexes.extension` is not because it isn't used in type annotations.
- moved `hamilton.node` to `TYPE_CHECKING` block since it's only used for annotations
- moved  `htypes` to top-level import; it should not have been in the a `try/except` in the first place because a code path of `SimplePythonDataFrameGraphAdapter` depends on it and will fail error if `htypes` isn't imported


I don't have the "why" for this code:

```python
try:
  from . import htypes, node
except ImportError:
  import node
```


- The `try/except` was [introduced in 2022](https://github.com/apache/hamilton/commit/0dcd52a71c32c3b4b43245c2603efa93b645ae29#diff-a60b4bde51922e7c95719941221e63dc0b991a8495b530c1997967f5294e73eb), but no clear indications why.
- Looking at the source code of [the file at the time](https://github.com/apache/hamilton/blob/0dcd52a71c32c3b4b43245c2603efa93b645ae29/hamilton/base.py), it was probably a brute force solution to avoid circular imports.
- The code could have been in a `TYPE_CHECKING` block (introduced in Python 3.5) since it was only ever used for annotations

zilto:
@pjfanning I have little experience with pypi, but I want to highlight that we have multiple options for distribution:
- pip allows you to install from github `pip install git+https://github.com/apache/hamilton.git`
- `@` allows to specify tag or brancehs. We could have `pip install git+https://github.com/apache/hamilton.git@core`
- you can also specify subpaths. We could have `pip install git+https://github.com/apache/hamilton.git#subdirectory=hamilton-core`
- the GitHub release workflow could include `core` versions in the release `Assets`
- These install instructions are valid in `requirements.txt` or `pyproject.toml`

Actually, you can even try `hamilton-core` right now from this PR via
```
pip install git+https://github.com/apache/hamilton.git@feat/hamilton-core#subdirectory=hamilton-core
```

I believe "hamilton-core" PR is a temporary solution to something we should fix on a major release. Users that care about this issue (already a few reacted positively on Slack) are probably ok with following a few extra steps for installation.

pjfanning:
@zilto the ASF frowns on encouraging users to use latest code in git. We aim to do official releases and have reviews and votes to improve the confidence about the release being stable. I think the pypi releases should be done with the source release and be based on the exact git commit that was accepted for the release.

hamilton/base.py
skrawcz: Yeah sorry I meant in the code leave a note/comment as to the importance :)

skrawcz: E.g. this is for hamilton-core to work...

zilto:
> @zilto the ASF frowns on encouraging users to use latest code in git. We aim to do official releases and have reviews and votes to improve the confidence about the release being stable. I think the pypi releases should be done with the source release and be based on the exact git commit that was accepted for the release.

Makes sense. Though, introducing `sf-hamilton-core` and having people depend on it means we'll have to maintain that pypi target "forever". If the fixes included in `sf-hamilton-core` are solved in Hamilton 2.0.0, doing `pip install sf-hamilton` or `pip install sf-hamilton-core` would do exactly the same thing, which could be confusing

skrawcz:
> we'll have to maintain that pypi target "forever" 

I'm not too concerned. Tooling should make this simpler. Once we hit 2.0. we have options to stop I think. I'm not too worried -- and if we always set that expectation I think we'll be good.

On that note, in the `__init__.py` can you log a warning that sf-hamilton-core will go away in 2.0?

Dev-iL:
Hey guys, what is the plan for this PR? I would very much like to see a Hamilton version that doesn't require pandas/numpy. BTW, now that we have prek, perhaps this open up some possibilities related to hooks/CI.

Dev-iL:
Some suggestions by Claude:

# Dependency & Structure Improvements for Hamilton

## 1. Remove `typing_inspect` dependency

`typing_inspect` is used in 6 files but all its functions have stdlib equivalents (Python 3.8+):

| `typing_inspect` function | Replacement |
|---|---|
| `get_origin()` / `get_args()` | Already handled in `hamilton/htypes.py` via `typing.get_origin` / `typing_extensions.get_origin` |
| `is_optional_type(tp)` | `get_origin(tp) is Union and type(None) in get_args(tp)` |
| `is_union_type(tp)` | `get_origin(tp) is Union` |
| `is_generic_type(tp)` | `get_origin(tp) is not None` |
| `is_tuple_type(tp)` | `get_origin(tp) is tuple` |
| `is_typevar(tp)` | `isinstance(tp, TypeVar)` |
| `is_literal_type(tp)` | `get_origin(tp) is Literal` |

These are used across `hamilton/htypes.py`, `hamilton/node.py`, `hamilton/function_modifiers/dependencies.py`, `hamilton/function_modifiers/expanders.py`, `hamilton/function_modifiers/adapters.py`, and `hamilton/experimental/h_cache.py`. A small helper module (or additions to `htypes.py`) could centralize these ~5 one-liner functions and eliminate the dependency entirely.

**Impact**: Removes 1 of 4 required dependencies.

---

## 2. Make remaining top-level pandas/numpy imports lazy

`base.py` was already fixed, but 3 files still have top-level imports:

| File | Import | Fix |
|---|---|---|
| `hamilton/driver.py:48` | `import pandas as pd` | Move inside the functions that use it (it's only used in the `__main__` block) |
| `hamilton/log_setup.py:21` | `import numpy as np` | Move inside `setup_logging()` — it's only used on line 47 for `np.seterr()` |
| `hamilton/models.py:21` | `import pandas as pd` | Move to `TYPE_CHECKING` block — only used for type annotations in `predict()` signature |

These are the remaining blockers for `hamilton-core` working cleanly without pandas/numpy installed. Without these fixes, `import hamilton.driver` or `import hamilton.log_setup` will crash in a `sf-hamilton-core` environment.

---

## 3. `typing_extensions` — keep for now

This dependency is minimal (used for `Annotated`, `NotRequired`, `get_origin`, `get_args`, `is_typeddict` — all backports for Python < 3.11). It's lightweight and can't be removed until the minimum Python version is bumped to 3.11+. The current constraint `> 4.0.0` is correct.

**Minor fix needed**: `hamilton/function_modifiers/expanders.py:762` imports `typing_extensions.is_typeddict` unconditionally — should add a version check like the other imports:

```python
if sys.version_info >= (3, 10):
    from typing import is_typeddict
else:
    from typing_extensions import is_typeddict
```

---

## 4. Improve the hamilton-core build with uv workspaces

The current `hamilton-core/setup.py` copies the entire `hamilton/` source tree into `hamilton/_hamilton/` at build time via `shutil.copytree`. This works but is fragile:

- Stale copies if setup.py isn't re-run
- `_hamilton/` is `.gitignore`d, so debugging the installed package is confusing
- The proxy module in `hamilton-core/hamilton/__init__.py` adds runtime overhead and complexity

**Better approach with uv workspaces**:

```toml
# Root pyproject.toml — add workspace config
[tool.uv.workspace]
members = ["hamilton-core"]

# hamilton-core/pyproject.toml — replace setup.py entirely
[project]
name = "sf-hamilton-core"
dynamic = ["version"]
dependencies = [
    "typing_extensions > 4.0.0",
    # no pandas, numpy, or typing_inspect
]

[tool.uv.sources]
sf-hamilton = { workspace = true }
```

This eliminates the copy step, the proxy module, and the dynamic `setup.py` entirely. The `hamilton` package is referenced directly from the workspace root.

---

## 5. Use prek workspace mode for pre-commit hooks

With `prek` now on main, you can use its workspace feature to manage hooks per-package:

```yaml
# hamilton-core/.pre-commit-config.yaml
orphan: true  # isolate from root hooks
repos:
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.15.1
    hooks:
      - id: ruff-check
      - id: ruff-format
```

`prek` auto-discovers this config and scopes the hooks to `hamilton-core/` files only. This keeps linting/formatting consistent while allowing each sub-package to customize.

---

## 6. Drop `networkx` from visualization extra

The current main `pyproject.toml` lists `networkx` as part of the `visualization` extra, but the `hamilton-core` setup.py already drops it:

```python
**{"visualization": ["graphviz"]},  # drop networkx
```

If networkx is genuinely unused in the visualization code path, this change should be applied to the main `pyproject.toml` as well.

---

## Summary — final dependency picture

| | `sf-hamilton` (current) | `sf-hamilton-core` (current) | After proposed changes |
|---|---|---|---|
| `pandas` | required | removed | removed |
| `numpy` | required | removed | removed |
| `typing_inspect` | required | required | **removed** |
| `typing_extensions` | required | required | required (until Python 3.11 min) |
| **Total core deps** | **4** | **2** | **1** |

The single remaining hard dependency would be `typing_extensions`, and even that goes away once Python < 3.11 is dropped. At that point, `sf-hamilton-core` would have **zero** third-party dependencies.

skrawcz:
@Dev-iL those are good points. @zilto what do you think?

PR #1387, opened 2025-09-23
Adds flit to build project

skrawcz:
We switch to flit on the recommendation of apache airflow.

Note: flit does not handle tuples for versions, so we have to duplicate this information for now. [I have submitted a PR](https://github.com/pypa/flit/pull/777) that they will hopefully accept to fix this.

## Changes
 - pyproject.toml

## How I tested this
- `flit build` works

## Checklist

### DIFF
--- a/scripts/apache_release_helper.py
+++ b/scripts/apache_release_helper.py
@@ -139,17 +139,23 @@ def sign_artifacts(archive_name: str) -> Optional[list[str]]:
     return files
 
 
-def create_release_artifacts(version) -> tuple[list[str], list[str]]:
+def create_release_artifacts(version) -> list[str]:
     """Creates the source tarball, GPG signature, and checksums using `python -m build`."""
     print("Creating release artifacts with 'python -m build'...")
-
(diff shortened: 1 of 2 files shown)


skrawcz:
I'm going to ignore that unit test -- test_huggingface_extensions related.

elijahbenizzy (approved):

pyproject.toml
```diff
@@ -272,3 +274,10 @@ exclude = ["*tests*"]
 
 [tool.setuptools.package-data]
 hamilton = ["*.json", "*.md", "*.txt"]
+
+[tool.flit.module]
+name = "hamilton"
+
+[tool.flit.sdist]
+include = ["LICENSE", "NOTICE", "DISCLAIMER"]
```
takluyver, line 282: This list serves the same purpose as `MANIFEST.in` does for setuptools, so if you do go ahead with using Flit, that file would be unused and you could remove it.

skrawcz: nice catch - thanks!

PR #1403, opened 2025-10-05
Updates UI: removes elkjs, updates backend deps

skrawcz:
## Changes

[Removes elkjs and adds dagre](https://github.com/apache/hamilton/commit/5921124f2de41372a7f32a1d55a1932f40f0d40c) 

Need to test this more -- but basic graphs work.

[Updates backend to use UV and updates to use 3.9](https://github.com/apache/hamilton/commit/23b955dfc94b85d34bbeb847144d9316227e2e08) 

So python 3.8 is gone. Updates to use python 3.9
Updates deps and to use uv for reproducible builds.

## How I tested this
 - locally

## Checklist


ui/backend/Dockerfile.backend
zilto, line 1: For reference, there are ways to optimize the Docker build process such that it uses the host's local `uv` cache when building Python dependencies inside the container. This can lead to significant speed improvement in dev

example: https://github.com/astral-sh/uv-docker-example/blob/main/multistage.Dockerfile

ui/backend/Dockerfile.backend-prod
```diff
@@ -1,20 +1,24 @@
...
+    && curl -LsSf https://astral.sh/uv/install.sh | sh
+
+ENV PATH="/root/.local/bin:$PATH"
 
 COPY . /code/
-COPY server/requirements.txt /code/
+COPY server/requirements-locked.txt /code/
 
-RUN pip install -r /code/requirements.txt
+RUN uv pip install --system -r /code/requirements-locked.txt
```
zilto, line 22: I see that `Dockerfile.backend` and `Dockerfile.backend-prod` already existed, but it seems they were identical except the additional line 15 (origin) in `backend-prod`?

I have a few questions:
- why are two files needed?
- why use the frozen `requirements-locked.txt` in `backend-prod`? This just seems like a weaker `uv.lock`. Are there any particular trade-offs?
- In `backend-prod`, the line `COPY . /code/` appears twice (line 15 and line 20 in origin). Is this intended?

skrawcz:
Superseded by #1600 which takes a cleaner approach using dagre's native compound graph support.

PR #1411, opened 2025-10-19
Drops support for 3.9, adds support for 3.13

elijahbenizzy:
3.9 is EOL in october

## Changes
- adds 3.13
- drops 3.8/3.9 provisions, tests, etc...

## How I tested this
- unit tests + CI

## Checklist


cswartzvi:
@elijahbenizzy I am very excited for this! What do you think about including freethreaded python in the actions matrix (`3.13t`)? Maybe we are not ready to go on record as supporting free threading, but it might be nice to keep an eye on it. What do you think?

ShoaibG1:
@elijahbenizzy I am still using AWS Glue Python Shell jobs which are still on Python 3.9. I hope AWS does something soon. I trust the current version of Hamilton will work fine with Python 3.9? I guess I won't get future enhancements and patches until I upgrade.

examples/hello_world/my_dag.png
zilto, line 1: remove file from PR? I believe it's a test artifact

examples/styling_visualization/dag.png
zilto, line 1: remove from PR? It's the same graph but with a different random layout

hamilton/node.py
```diff
@@ -68,7 +68,7 @@ def __init__(
         doc_string: str = "",
         callabl: Callable = None,
         node_source: NodeType = NodeType.STANDARD,
-        input_types: Dict[str, Union[Type, Tuple[Type, DependencyType]]] = None,
+        input_types: Optional[Dict[str, Union[Type, Tuple[Type, DependencyType]]]] = None,
```
skrawcz: Why did this change? Otherwise | No e would be better

elijahbenizzy: Just linting but this doesn't have to be part of it.

skrawcz (approved):
Just one Q.

PR #1425, opened 2025-12-28
Adds more licenses and utilities for license checking

skrawcz:
Adds more license headers

## Changes

1. Updated headers 

  - examples/ - 537 files
    - 343 Python files (.py)
    - 105 Markdown files (.md)
    - 73 Jupyter notebooks (.ipynb)
    - 2 Shell scripts (.sh)
    - 2 SQL files (.sql)
    - 9 Dockerfiles
    - 4 README files (no extension)
  - ui/ - 23 files
    - 22 test files in ui/sdk/tests/
    - 1 file in ui/backend/server/trackingserver_base/
  - docs/ - 3 files
    - conf.py
    - data_adapters_extension.py
    - make_testimonials.py
  - Other directories - 5 files
    - 2 files in scripts/
    - 1 file in hamilton/experimental/
    - 1 file in writeups/
    - 1 file in ui/backend/

  Total: 568 files with Apache 2 license headers added

 2. Created Tooling

  - scripts/check_license_headers.py - Script to find files missing Apache 2 license headers
    - Supports: .py, .md, .ipynb, .sh, .sql, Dockerfile, and README files
    - Can check specific file types or all at once
    - Excludes build artifacts, node_modules, and documentation snippets
  - scripts/add_license_headers.py - Script to automatically add license headers
    - Adds appropriate comment formats based on file type:
        - Python/Shell/Dockerfile: # comments
      - Markdown/README:  HTML comments
      - SQL: -- comments
      - Jupyter notebooks: Markdown cell at the beginning
    - Handles special cases (shebangs, existing headers)
    - Dry-run mode for testing

  3. Updated Notebook Validation

  - examples/validate_examples.py - Updated pre-commit hook validator
    - Now correctly handles notebooks with license headers
    - Checks for license at cell 0, setup at cell 1, title at cell 2 (if license present)
    - Maintains backward compatibility with notebooks without license headers

 ### Exclusions

  - Documentation snippets in docs/code-comparisons/*snippets/ - These are embedded in docs via literalinclude and adding headers would clutter the rendered documentation
  - examples/plotly/interactive.html - Has a different license
  - dev_tools/language_server/tests/ (2 files) - Already have appropriate Apache 2.0 licenses from third-party sources (Open Law Library, Palantir Technologies)

### Testing

  - Verified all files with scripts/check_license_headers.py
  - Pre-commit notebook validation passes for all example notebooks
  - License headers use correct comment format for each file type

###  Notes

  - License header format follows Apache Software Foundation standard
  - Headers are added at the beginning of files (after shebang if present)
  - No functional code changes - only license header additions

## Checklist

### DIFF
--- a/scripts/add_license_headers.py
+++ b/scripts/add_license_headers.py
@@ -1,10 +1,32 @@
 #!/usr/bin/env python3
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
 """Script to add Apache 2 license headers to files in the Hamilton repository."""
 
 import json
 import sys
 from pathlib import Path
 from typing import List
+
+# TODO: Simplify this script if we add more file types. Since the license text
+# (including line breaks) stays the same, we really just need different comment
+# character insertion helpers based on file extension rather than duplicating
+# the full license text for each file type.
 
 # Apache 2 license header for Python files
 PYTHON_LICENSE_HEADER = """# Licensed to the Apache Software Foundation (ASF) under one
@@ -87,7 +109,7 @@ SQL_LICENSE_HEADER = """-- Licensed to the Apache Software Foundation (ASF) unde
 """
 
 
-def add_license_to_python(file_path: Path, content: str) -> str:
+def add_license_to_python(content: str) -> str:
     """Add Apache 2 license header to Python file content."""
     # Handle shebang lines - preserve them at the top
     lines = content.split("\n", 1)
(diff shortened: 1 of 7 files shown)


scripts/add_license_headers.py
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/env python3
```
jernejfrank, line 1: I think this is fine for now, but maybe worth adding a TODO to simplify this in case we get more file types with different comment behaviour. Since the licence text (including line breaks) stays the same, we really just need different char insertion helpers based on file extension.

scripts/add_license_headers.py
```diff
@@ -0,0 +1,297 @@
...
+        with open(file_path, "r", encoding="utf-8") as f:
+            content = f.read()
+    except (UnicodeDecodeError, PermissionError) as e:
+        print(f"  ✗ Error reading {file_path}: {e}")
+        return False
+
+    # Check if file already has a license header
+    if "Licensed to the Apache Software Foundation" in content or "Apache License" in content:
+        print(f"  ↷ Skipping {file_path} (already has license header)")
+        return False
```
jernejfrank: Here we check the whole file, so some comment in the middle of the file saying "Apache License" could trigger this. Would be better to check only first and second (in case of shebang) line of the file.

scripts/add_license_headers.py
```diff
@@ -0,0 +1,297 @@
...
+-- software distributed under the License is distributed on an
+-- "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+-- KIND, either express or implied.  See the License for the
+-- specific language governing permissions and limitations
+-- under the License.
+
+"""
+
+
+def add_license_to_python(file_path: Path, content: str) -> str:
```
jernejfrank: All `add_licence_to_*` helpers have `file_path` but only use `content`

scripts/add_license_headers.py
jernejfrank: Just saw below there's already a helper below to check for licence header. Would make sense to just re-use it here.

scripts/check_license_headers.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
```
jernejfrank, line 1: Here too, file missing licence header ;)

scripts/add_license_headers.py
skrawcz: make sense

scripts/check_license_headers.py
skrawcz: done.

scripts/add_license_headers.py
skrawcz: done.

scripts/add_license_headers.py
skrawcz: fixed

scripts/add_license_headers.py
skrawcz: added.

PR #1535, opened 2026-04-01
Remove license comments from databackend.py

pjfanning:
Removed Apache License comments from the file.

This Apache header should not have been added - see point 3 in https://www.apache.org/legal/src-headers.html#3party


skrawcz (approved):

skrawcz:
I will update pre-commit to skip this one file

Issue #1551, opened 2026-04-19
Support optional `from_` extractors when DAG target does not exist


mgdemers:
**Is your feature request related to a problem? Please describe.**
When using a set of extractors, if the DAG module being built does not reference the extractor targets, an exception is thrown during building: `ValueError: Could not find node with name: input3 in function graph. Available nodes: ['input1', 'input2', Ellipsis]`. 

**Describe the solution you'd like**
When using `from_` to create extractors, support an additional flag which allows builds to skip an extractor if the target node does not exist. 

**Describe alternatives you've considered**
* Additional introspection on the dag to check what targets are used and explicitly avoid registering those unused extractors. This adds additional inspection steps outside of the DAG building paths, and does not work if the target feeds a decorator.

**Additional context**
Originally discussed in Hamilton dev slack [here](https://hamilton-opensource.slack.com/archives/C03M33QB4M8/p1776273350116259).

skrawcz:
Closed — implemented in PR #1552.

PR #1552, opened 2026-04-19
feat: support optional flag in from_ extractors

mgdemers:
Support optional `from_` extractors when DAG modules' targets do not reference an extractor.

## Changes

* Added `optional` flag to `ExtractorFactory`
* When no target node is identified during `generate_nodes`, return an empty list instead of raising `ValueError` when flag is set.
* Add unit test for an optional extractor not referenced by a DAG module

## How I tested this

* Unit tests

## Notes

Discussion in #1551 

## Checklist

### DIFF
--- a/hamilton/io/materialization.py
+++ b/hamilton/io/materialization.py
@@ -136,6 +136,7 @@ class ExtractorFactory:
         self,
         target: str,
         loaders: list[type[DataLoader]],
+        optional: bool = False,
         **data_loader_kwargs: Any | SingleDependency,
     ):
         """Instantiates an ExtractorFactory. Note this is not a public API -- this is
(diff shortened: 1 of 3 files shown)


hamilton/io/materialization.py
```diff
@@ -144,10 +145,12 @@ def __init__(
 
         :param target: Parameter, into which we're loading the data
         :param loaders: A list of data loaders that are viable candidates, given the key after `from_`
+        :param optional: Whether missing targets should raise exceptions or skip being loaded.
```
skrawcz, line 148: I would add a bit more here. 

```suggestion
        :param optional: Whether missing targets should raise exceptions or skip being loaded. Optional=True will mean that if the target does not exist in your graph, we skip try to inject this into the graph.
```

skrawcz (commented):
looks good. can you update the materialization docs too? A comment or example snippet should suffice.

hamilton/io/materialization.py
mgdemers: Added - I also added a note on the materialization docs about specifying optional data loaders.

skrawcz (approved):
Thanks for the contribution! Clean implementation and nice that you added the docs note too. Appreciate it!

PR #1559, opened 2026-04-26
Fix resolve docs parameterization example

### LINKED ISSUE #1384: Documentation for @resolve has broken example
The [documentation for `@resolve` decorator](https://hamilton.apache.org/reference/decorators/resolve/) includes this example code:

```python
@parameterize_values(
    series_sum_1={"s1": "series_1", "s2": "series_2"},
    series_sum_2={"s1": "series_3", "s2": "series_4"},
)
def summation(df: pd.DataFrame, s1: str, s2: str) -> pd.Series:
    return df[s1] + df[s2]
```

However, that is incorrect usage of the `@parameterize_values` decorator. That is the usage for `@parameterize` decorator.

1. Go to [the docs page](https://hamilton.apache.org/reference/decorators/resolve/)
   * Alternatively, view [this snapshot of the source content](https://github.com/apache/hamilton/blob/c838df2f46a4105ee413ad233d469a2b445afbb2/hamilton/function_modifiers/delayed.py#L80).

# Additional context

This example was changed from `@parameterize` to `@parametrize_values` in commit 67f7918 17 months ago. I'm unsure if it was valid then, but I'm fairly certain it is not valid now.

Abdeltoto:
## Summary
- Fix the `@resolve` reference docs example to use `@parameterize_sources` instead of `@parameterize_values`.
- The example maps function parameters to upstream source nodes, which matches `parameterize_sources` semantics and the later dynamic `@resolve` example.

Fixes #1384

## Test plan
- `python -m ruff check hamilton/function_modifiers/delayed.py`
- `git diff --check`

### DIFF
--- a/hamilton/function_modifiers/delayed.py
+++ b/hamilton/function_modifiers/delayed.py
@@ -78,7 +78,7 @@ class resolve(DynamicResolver):
 
     .. code-block:: python
 
-        @parameterize_values(
+        @parameterize_sources(
             series_sum_1={"s1": "series_1", "s2": "series_2"},
             series_sum_2={"s1": "series_3", "s2": "series_4"},
         )


skrawcz (approved):
Thanks! Good catch on the docs example.

PR #1560, opened 2026-04-26
Fix async builder fluent typing

### LINKED ISSUE #1452: Typing issue when defining an async driver
I'm encountering a type-checking error when trying to define an `async_driver` like this:

```python
driver = await async_driver.Builder().with_modules(my_flow).build()
```

### Current behavior

Type checker (e.g. Pyright or mypy) reports:

```
Diagnostics:
1. "Driver" is not awaitable
   "Driver" is incompatible with protocol "Awaitable[_T_co@Awaitable]"
   "__await__" is not present [reportGeneralTypeIssues]
```

It seems the returned `Driver` object is missing an `__await__` implementation or the builder’s `build()` method isn’t correctly annotated as returning an awaitable type.

Abdeltoto:
## Summary
- Use `typing_extensions.Self` for fluent `driver.Builder` methods that return `self`.
- Annotate `async_driver.Builder.build()` as returning `AsyncDriver` when awaited.
- This preserves the async builder type through chains like `async_driver.Builder().with_modules(...).build()` instead of falling back to the synchronous `driver.Builder.build()` type.

Fixes #1452

## Test plan
- `python -m ruff check hamilton/driver.py hamilton/async_driver.py`
- `git diff --check`
- `python -m pytest tests/test_async_driver.py -k Builder` could not run locally because `typing_inspect` is not installed in this environment.
- `pyright` is not installed locally, so I could not run the original reporter's checker here.

### DIFF
--- a/hamilton/async_driver.py
+++ b/hamilton/async_driver.py
@@ -453,7 +453,7 @@ class Builder(driver.Builder):
             allow_module_overrides=self._allow_module_overrides,
         )
 
-    async def build(self):
+    async def build(self) -> AsyncDriver:
         """Builds the async driver. This also initializes it, hence the async definition.
         If you don't want to use async, you can use `build_without_init` and call `ainit` later,
         but we recommend using this in an asynchronous lifespan management function (E.G. in fastAPI),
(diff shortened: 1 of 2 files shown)


skrawcz (approved):
Thanks! LGTM — clean use of `Self` for the fluent builder pattern.

PR #1570, opened 2026-05-05
fix tqdm: Fix progress bar to 100% on successful execution

### LINKED ISSUE #845: TDQM hook with overrides
# Current behavior
When using `hamilton.plugins.h_tdqm.ProgressBar()`, the progress bar uses "the number of nodes to return" as denominator and "the number of nodes executed" as numerator. However, passing values to `.execute(overrides=...)` counts towrds the first, but not the latter. Code runs to completion but the progress bar completes without reaching 100%

## Screenshots
image

## Dataflow
```python
# joke.py

def joke_prompt(topic: str) -> str:
    return f"Knock, knock. Who's there? {topic}"

def reply(joke_prompt: str) -> str:
    _, _, right = joke_prompt.partition("? ")
    return f"{right} who?"

def punchline(reply: str) -> str:
    left, _, _ = reply.partition(" ")
    return f"No, {left} MooOOooo"
```

## Execution
```python
import joke
from hamilton import driver
from hamilton.plugins.h_tqdm import ProgressBar
dr = (
    driver.Builder()
    .with_adapters(ProgressBar())
    .with_modules(joke)
    .build()
)

dr.execute(["joke_prompt", "reply"], overrides=dict(punchline="..."), inputs=topic("hello")).
```

ArnavBalyan:
## Changes
 - Fixes #845. The `h_tqdm.ProgressBar` adapter never reached 100% when `overrides` were passed to `.execute()` because override nodes are counted in the bar's `total` but do not trigger `run_after_node_execution`,
  leaving the bar short of `total` on a successful run.
 - On a successful graph execution we top up `progress_bar.n` to `progress_bar.total` inside `run_after_graph_execution`.. On failure we leave the partial count visible.

## How I tested this
 - UT
## Notes
 - N/A
## Checklist

### DIFF
--- a/hamilton/plugins/h_tqdm.py
+++ b/hamilton/plugins/h_tqdm.py
@@ -111,12 +111,18 @@ class ProgressBar(
     def run_after_node_execution(self, **future_kwargs):
         self.progress_bar.update(1)
 
-    def run_after_graph_execution(self, **future_kwargs):
+    def run_after_graph_execution(self, *, success: bool = True, **future_kwargs):
         name_part = "Execution Complete!"
         if len(name_part) > self.node_name_target_width:
             padding = ""
         else:
             padding = " " * (self.node_name_target_width - len(name_part))
+        # Overrides are counted in `total` but never trigger run_after_node_execution,
+        # so on a successful run we top the bar up to 100% rather than leaving a gap.
+        if success and self.progress_bar.total is not None:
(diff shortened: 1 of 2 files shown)


ArnavBalyan:
cc @skrawcz :)

skrawcz (approved):
thanks!

PR #1575, opened 2026-05-06
docs(reuse): add User Guide page on reusing functions and nodes (#1045)

### LINKED ISSUE #1045: Documentation: add guide on node reuse
Node reuse is a common question.

We should have a section in the user guide on it.

The content for this exists. It just needs to be written up / referenced.

* Slides from meetup on this - https://github.com/skrawcz/talks/files/14657471/Hamilton.March.2024.Meetup.pdf
* [youtube video explaining this ](https://youtu.be/IJByeN41xHs)
* [Notebook that has examples](https://github.com/DAGWorks-Inc/hamilton-tutorials/blob/main/2024-03-19/march-meetup.ipynb)

N/A

**Additional context**
This would be a highly read section.

Hore01:
This adds a User Guide page that pulls together Hamilton's four patterns for reusing the same logic across different data sources, regions, or runtime contexts:

- driving the same function module from multiple Drivers
  (`examples/feature_engineering/feature_engineering_multiple_contexts/`)
- `Builder.allow_module_overrides()` for swapping same-named functions (`examples/module_overrides/`)
- `@subdag` to repeat a transformation graph inside one Driver (`examples/reusing_functions/`)
- `@parameterized_subdag` for many subdags at once, with the readability warning quoted from the source docstring about preferring static DAGs when you don't have many copies

Closes #1045.

The new page is `docs/how-tos/reuse-nodes.rst`, slotted into the User Guide toctree just after `use-for-feature-engineering`. Every code block on the page is a `.. literalinclude::` pointing at the example
files that already exist in the repo (`module_a.py`, `module_b.py`, `run.py`, `unique_users.py`, `reusable_subdags.py`), so the prose can't drift away from the runnable code.

I tested locally on Python 3.12: sphinx (`-b dirhtml`) builds cleanly and no warnings reference the new page or any of the example files it pulls from. The 41 warnings I do see locally are the same autodoc warnings about optional plugin deps that already exist on main and that CI resolves via the full `docs` dependency group.

A couple of notes:

- The page deliberately doesn't introduce a new example folder. Per the issue, the existing `reusing_functions/`, `module_overrides/` and `feature_engineering_multiple_contexts/` examples already cover the patterns; the goal here is to point readers at them in one place with the right framing.
- Commit is DCO-signed.

### DIFF
--- a/docs/how-tos/index.rst
+++ b/docs/how-tos/index.rst
@@ -12,6 +12,7 @@ directory. If there's an example you want but don't see, reach out or open an is
    load-data
    caching-tutorial
    use-for-feature-engineering
+   reuse-nodes
    ml-training
    llm-workflows
    run-data-quality-checks
--- a/docs/how-tos/reuse-nodes.rst
+++ b/docs/how-tos/reuse-nodes.rst
@@ -1,0 +1,262 @@
+..
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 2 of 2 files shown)


skrawcz (approved):
thanks!

PR #1582, opened 2026-05-12
Fix check input type bug for parameterized tuples

### LINKED ISSUE #1550: check_input_type rejects parameterized tuple annotations and PEP 695 type aliases
`hamilton.htypes.check_input_type` rejects parameterized tuple annotations and PEP 695 `type X = ...` aliases, raising "Type requirement mismatch" at `dr.execute(inputs=...)` time even when the input value matches the declared type.

# Current behavior

Two issues:

1. Parameterized tuple annotations are rejected. The sequence-generic branch in `check_input_type` includes `tuple` in its origin check, but `typing_inspect.is_generic_type(tuple[float, float, float, float])` returns False, so that branch never fires. The same call returns True for `list[int]` and `dict[str, int]`, which do pass. `list[int]` accepts a matching list, but `tuple[float, float, float, float]` rejects a matching 4-tuple.

2. PEP 695 `type X = ...` aliases are never unwrapped. A `TypeAliasType` instance is not a class, not a generic per `typing_inspect`, and not a TypeVar, so every branch falls through and the function returns False regardless of what the alias resolves to.

Both failures surface at `dr.execute(inputs=...)` as a `ValueError` whose message reads "Type requirement mismatch. Expected X got Y", which misattributes the failure to the input value when the real issue is the type form.

## Stack Traces

```
ValueError: 1 errors encountered:
  Error: Type requirement mismatch. Expected area_bbox:tuple[float, float, float, float] got (4.883, 52.366, 4.895, 52.374):<class 'tuple'> instead.
```

Raised from `hamilton/driver.py:608` in `Driver.validate_inputs`, after `hamilton/htypes.py:333` `check_input_type` returns False.

1. Save this script as `repro.py`:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = ["sf-hamilton"]
# ///
from hamilton.htypes import check_input_type

assert check_input_type(tuple[float, float, float, float], (1.0, 2.0, 3.0, 4.0))

type Bbox = tuple[float, float, float, float]
assert check_input_type(Bbox, (1.0, 2.0, 3.0, 4.0))
```

2. Run `uv run repro.py`. Both asserts fail on Hamilton 1.89.0.

- Python 3.14.3
- sf-hamilton 1.89.0
- typing_inspect 0.9.0
- macOS 25.4.0 (also reproducible on Linux)

# Expected behavior

Both assertions should return True.

- `check_input_type(tuple[float, float, float, float], (1.0, 2.0, 3.0, 4.0))` should accept the tuple, matching the existing behavior of `list[int]` and `dict[str, int]`.
- `check_input_type(Bbox, value)` for a PEP 695 alias should behave the same as the type the alias resolves to.

# Additional context

Issue 1 root cause: `typing_inspect.is_generic_type` returns False for lowercase PEP 585 parameterized tuples specifically. The stdlib `typing.get_origin` handles this correctly since 3.8. Migrating the tuple/list branch from `typing_inspect.is_generic_type(node_type)` to `typing.get_origin(node_type) is not None` fixes it without changing semantics for list/dict.

Issue 2 root cause: no `TypeAliasType` handling. Unwrapping it at the top of `check_input_type`:

```python
from typing import TypeAliasType  # stdlib since 3.12

def check_input_type(node_type, input_value):
    if isinstance(node_type, TypeAliasType):
        node_type = node_type.__value__
    # ... rest unchanged
```

resolves the second issue and composes with any type the alias ultimately resolves to. For Python versions older than 3.12, `typing_extensions.TypeAliasType` is a drop-in source.

ArnavBalyan:
## Changes
 -  The runtime type checker rejects parameterized tuples and new type alias, leading to type mismatch errors at execute time.
   - The internal type check wrongly returns false for parameterized tuples/dicts.
   - Fix by switching to the standard library origin check and per position matching for fixed length tuples. Also unwrapping the alias and recursing on the underlying type.
   - Closes #1550.

## How I tested this
 - UT
## Checklist

### DIFF
--- a/hamilton/htypes.py
+++ b/hamilton/htypes.py
@@ -21,9 +21,19 @@ import typing
 from collections.abc import Iterable
 from typing import Any, Literal, Protocol, TypeVar, Union
 
+import typing_extensions
 import typing_inspect
 
 from hamilton.registry import COLUMN_TYPE, DF_TYPE_AND_COLUMN_TYPES
+
+_TYPE_ALIAS_TYPES = tuple(
+    t
+    for t in (
+        getattr(typing, "TypeAliasType", None),
+        getattr(typing_extensions, "TypeAliasType", None),
+    )
+    if t is not None
+)
 
 BASE_ARGS_FOR_GENERICS = (typing.T,)
 
@@ -336,6 +346,8 @@ def check_input_type(node_type: type, input_value: Any) -> bool:
     :param input_value: Value to check.
     :return: True if the input value is of the correct type, False otherwise.
     """
+    if _TYPE_ALIAS_TYPES and isinstance(node_type, _TYPE_ALIAS_TYPES):
+        return check_input_type(node_type.__value__, input_value)
     if node_type == Any:
         return True
     # In the case of dict[str, Any] (or equivalent) in python 3.9 +
@@ -343,6 +355,7 @@ def check_input_type(node_type: type, input_value: Any) -> bool:
     elif (
         inspect.isclass(node_type)
         and not typing_inspect.is_generic_type(node_type)
+        and typing.get_origin(node_type) is None
         and isinstance(input_value, node_type)
     ):
         return True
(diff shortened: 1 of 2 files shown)


ArnavBalyan:
cc @skrawcz ty :)

ArnavBalyan:
cc @skrawcz gentle reminder

skrawcz (approved):
Nice fix! Per-position matching for fixed-length tuples is correct per PEP 484, and the TypeAlias unwrapping handles PEP 695 cleanly. All existing tests pass. Thanks!

PR #1594, opened 2026-05-21
Fix main failing due to sklearn bump

ArnavBalyan:
## Changes
 - Main/CI is failing, sklearn got updated to 1.9.0rc1
 - We are not pinned so we picked up the new release which has breaking changes
 - Pin to 1.8.0 last safe version which keeps main intact

## How I tested this
 - CI 
## Checklist


ArnavBalyan:
cc @skrawcz thanks!

skrawcz (commented):
yeah I think I just fixed this on main.

In general we try to either (a) fix forward, (b) handle multiple versions if possible, (c) if we do pin, specify what and why and when to unpin. (d) we could probably lower pin against major version.

ArnavBalyan:
Sg closing this

PR #1595, opened 2026-05-22
Add LLM information extraction example

### LINKED ISSUE #241: [Example] Create a dataflow modeling information extraction using an LLM
Write an example dataflow that uses Hamilton to model an information extraction task using an LLM. 

For example:
1. given an output schema
2. given input text
3. make a prompt that is sent to an LLM API (pick one)
4. then write a function to validate the output

For inspiration you can look at [Langchain's implementation](https://github.com/langchain-ai/langchain/blob/490ad93b3cf7d24b30f8993f860b654ff107e638/libs/langchain/langchain/chains/openai_functions/extraction.py#L46).

The code for this example should end up under the /examples/LLM_Workflows directory.

Sarthak-commits:
Closes #241.\n\nAdds a Hamilton LLM workflow example for structured information extraction from customer feedback. The example defines a Pydantic output schema, builds the extraction prompt, calls the OpenAI chat API, parses the JSON response, and validates it before returning serializable records.\n\nThe runner also supports a mocked response so the validation path can be exercised without an API key.

### DIFF
--- a/examples/LLM_Workflows/information_extraction/README.md
+++ b/examples/LLM_Workflows/information_extraction/README.md
@@ -1,0 +1,65 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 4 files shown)


jernejfrank (commented):
Hey, thankl you for submitting this. Could you also:
1. save the dag representation as a PNG (driver has a method for this already) and embed it in the README?
2. Create a jupyter notebook that does the same as `run.py` and allows users to walk through the example in that way (see other examples)?

Sarthak-commits:
Yeah sure i will submit it, give me some time i am out of town for the moment.

PR #1596, opened 2026-05-22
docs: enable source links in API reference

RaghunandanKumar:
## Summary

Enable `sphinx.ext.viewcode` in the Sphinx docs config so generated API reference pages include links back to the source code.

## Test Plan

- Attempted local Sphinx build
- Blocked locally by host packaging/env mismatch (`pip` on this machine does not support the repo's `--group docs` install flow)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -72,6 +72,7 @@ html_css_files = [
 extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
+    "sphinx.ext.viewcode",
     "myst_nb",
     "sphinx_llms_txt",
     "sphinx_sitemap",


skrawcz:
@RaghunandanKumar this doesn't seem to work as intended? The docs built so I can merge, but the staged docs doesn't seem to have source code links?

skrawcz (approved):

RaghunandanKumar:
I re-verified locally with same  build CI uses, and generated API reference does include  links now.

For example, built  contains links like:


 check also passes on this PR, so I think current staged page is serving stale/pre-change artifact rather than output from this branch.

RaghunandanKumar:
I re-verified locally with same `dirhtml` build CI uses, and generated API reference does include `[source]` links now.

For example, built `reference/drivers/Driver/index.html` contains links like:
`../../../_modules/hamilton/driver/#Builder`

`build-docs` check also passes on this PR, so I think current staged page is serving stale/pre-change artifact rather than output from this branch.

PR #1598, opened 2026-05-24
Allow bypassing data quality checks via a config key

Dev-iL:
## Context

`@check_output` and `@check_output_custom` decorators let users attach data validators to function outputs. At graph-construction time, each decorated node is expanded into a subgraph:

- `{name}_raw` — runs the actual function
- `{name}_{validator}` — one node per validator (tagged `hamilton.data_quality.contains_dq_results`)
- `{name}` — aggregates validation results and returns the raw output

This flag is useful for two main reasons:
1. Validation may be crucial during development, but carries a real runtime cost in production where the pipeline is already trusted. Previously there was no way to turn it off short of removing the decorators from the source code.
2. When generating graph visualizations and we don't want to crowd them with validation nodes, passing  `disable_data_quality_checks=True` at construction time causes the `_raw` and validator nodes not to be added to the graph - so any visualization of that driver's graph will only show the "business logic" nodes.

### `hamilton/function_modifiers/validation.py`

`BaseDataValidationDecorator` (parent of both `check_output` and `check_output_custom`) now:

1. Overrides `optional_config()` to declare `{"disable_data_quality_checks": False}`. This registers the key with Hamilton's config-filtering pipeline so it is threaded through to `transform_node` automatically — no driver plumbing needed.

3. Adds an early-return guard at the top of `transform_node`:

   ```python
   if config.get("disable_data_quality_checks", False):
       return [node_]
   ```

   When the flag is set, the node is returned unchanged — no `_raw` node, no validator nodes, and no expansion happens. The cost is literally zero: the extra nodes are never created.

### `hamilton/driver.py`

`Builder` gains a convenience method:

```python
def with_data_quality_disabled(self) -> "Builder":
    return self.with_config({"disable_data_quality_checks": True})
```

This is a thin wrapper over `with_config` — discoverable by IDE autocomplete and explicit about intent. Because `with_config` does a dict `.update()`, a later call like `.with_config({"disable_data_quality_checks": False})` will re-enable validation, which is the expected last-write-wins behavior.

## Usage

**Via the `Builder` convenience method (recommended):**

```python
dr = (
    hamilton.driver.Builder()
    .with_modules(my_pipeline)
    .with_data_quality_disabled()
    .build()
)
```

**Via `with_config` directly (equivalent, useful when config is assembled dynamically):**

```python
dr = (
    hamilton.driver.Builder()
    .with_modules(my_pipeline)
    .with_config({"disable_data_quality_checks": True})
    .build()
)
```

**Legacy `Driver` constructor:**

```python
dr = hamilton.driver.Driver(
    {"disable_data_quality_checks": True},
    my_pipeline,
    adapter=DefaultAdapter(),
)
```

## Tests added

| File | Test | What it covers |
|------|------|----------------|
| `tests/function_modifiers/test_validation.py` | `test_check_output_disabled_via_config_returns_original_node` | `check_output_custom` returns the original node unchanged when flag is set |
| `tests/function_modifiers/test_validation.py` | `test_check_output_builtin_disabled_via_config_returns_original_node` | `check_output` (built-in validators) also respects the flag |
| `tests/test_end_to_end.py` | `test_builder_with_data_quality_disabled_removes_validator_nodes` | No DQ-tagged nodes appear in `list_available_variables()` when disabled |
| `tests/test_end_to_end.py` | `test_builder_with_data_quality_disabled_still_executes_correctly` | Driver executes correctly and returns the function's real output when disabled |
| `tests/test_end_to_end.py` | `test_disable_data_quality_checks_config_key_works_directly` | Raw `with_config` path (no convenience method) also suppresses validator nodes |

All 13 pre-existing validation tests continue to pass.

## Design notes

- **Graph-construction time, not execution time.** Disabling at construction eliminates the extra nodes entirely. An execution-time approach (e.g., a lifecycle adapter) would still pay graph construction and scheduling overhead, and would be harder to reason about.

- **Config key, not a subclass.** A `NoValidationBuilder` subclass was considered. It was rejected: this is a single boolean flag, not a fundamentally different execution model. The existing `Builder` method pattern (`with_config`, `with_adapters`, …) is the right weight here. The `Builder.with_data_quality_disabled()` convenience method is offered as a discoverable alias — if maintainers prefer, the method alone (without the raw config key) or the raw key alone (without the method) is equally viable.

- **`optional_config()` is the correct hook.** Hamilton's `resolve_config` / `filter_config` machinery passes only declared config keys to each decorator. Registering the key via `optional_config()` means the flag is silently ignored (defaulting to `False`) by all existing drivers that never set it — no breaking change.

## Checklist

### DIFF
--- a/docs/how-tos/run-data-quality-checks.rst
+++ b/docs/how-tos/run-data-quality-checks.rst
@@ -16,4 +16,47 @@ Async validators
 
 For validation logic that requires async operations (e.g., async database queries or API calls), use ``AsyncDataValidator`` or ``AsyncBaseDefaultValidator`` from ``hamilton.data_quality.base``. These define ``async def validate()`` and work with ``AsyncDriver``. You can mix sync and async validators in a single ``@check_output_custom`` call.
 
+Disabling validators at runtime
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Validators are useful during development but may be unnecessary overhead in a trusted production pipeline. You can disable all ``@check_output`` and ``@check_output_custom`` validators at graph-construction time, so no extra nodes are ever created:
+
+.. code-block:: python
+
+    dr = (
+        hamilton.driver.Builder()
+        .with_modules(my_pipeline)
+        .with_data_quality_disabled()
+        .build()
+    )
+
+This is equivalent to passing ``{"hamilton.data_quality.disable_checks": True}`` via ``.with_config()``, which is useful when the flag is controlled dynamically (e.g., from an environment variable):
+
+.. code-block:: python
+
+    import os
+
+    dr = (
+        hamilton.driver.Builder()
+        .with_modules(my_pipeline)
+        .with_config({"hamilton.data_quality.disable_checks": os.getenv("DISABLE_DQ", "false") == "true"})
+        .build()
+    )
+
+Because the flag is resolved at graph-construction time, disabled drivers carry zero runtime overhead from validation — no validator nodes are created at all.
+
+A second use case is graph visualization. Each decorated function normally expands into several nodes (``{name}_raw``, one per validator, and the final ``{name}`` node), which can clutter a visualization when you want to communicate pipeline structure rather than validation wiring. Building a driver with ``with_data_quality_disabled()`` gives a clean visualization with only the business-logic nodes:
+
+.. code-block:: python
+
+    dr_viz = (
+        hamilton.driver.Builder()
+        .with_modules(my_pipeline)
+        .with_data_quality_disabled()
+        .build()
+    )
+    dr_viz.display_all_functions("pipeline.png")
+
+Note that this requires a separate driver instance from the one used for execution if you still want validations to run.
+
 See the :doc:`check_output reference <../reference/decorators/check_output>` and `data quality writeup <https://github.com/apache/hamilton/blob/main/writeups/data_quality.md>`_ for details and examples.
--- a/hamilton/driver.py
+++ b/hamilton/driver.py
@@ -45,6 +45,7 @@ from hamilton.caching.adapter import HamiltonCacheAdapter
 from hamilton.caching.stores.base import MetadataStore, ResultStore
 from hamilton.dev_utils import deprecation
 from hamilton.execution import executors, graph_functions, grouping, state
+from hamilton.function_modifiers.validation import DISABLE_DATA_QUALITY_CHECKS_CONFIG_KEY
 from hamilton.graph_types import HamiltonNode
 from hamilton.io import materialization
 from hamilton.io.materialization import ExtractorFactory, MaterializerFactory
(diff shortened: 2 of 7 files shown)


skrawcz (commented):
Nice feature! Clean implementation — the one-liner in Builder delegating to `with_config` is the right approach, and the 2-line check in `transform_node` is minimal and correct. Tests cover both decorators plus end-to-end via the driver.

One nit: `with_data_quality_disabled()` returns `-> "Builder"` (string literal) — should be `-> Self` to match the convention established in #1560. Otherwise LGTM.

The sklearn test fix is unrelated but correct — fine to include here.

skrawcz (commented):
-

skrawcz (commented):
-

docs/how-tos/run-data-quality-checks.rst
```diff
@@ -16,4 +16,47 @@ Async validators
...
+.. code-block:: python
+
+    dr = (
+        hamilton.driver.Builder()
+        .with_modules(my_pipeline)
+        .with_data_quality_disabled()
+        .build()
+    )
+
+This is equivalent to passing ``{"disable_data_quality_checks": True}`` via ``.with_config()``, which is useful when the flag is controlled dynamically (e.g., from an environment variable):
```
skrawcz: we should probably __ prefix or something signify internal? I don't think we've done that yet -- but somethign to think about to avoid key collisions ...

hamilton/function_modifiers/validation.py
```diff
@@ -42,9 +42,14 @@ def get_validators(self, node_to_validate: node.Node) -> list[dq_base.DataValida
         """
         pass
 
+    def optional_config(self) -> dict[str, Any]:
+        return {"disable_data_quality_checks": False}
```
skrawcz: let's make a constant

skrawcz (changes requested):
let's just make a constant

docs/how-tos/run-data-quality-checks.rst
Dev-iL: Changed to `hamilton.data_quality.disable_checks`

hamilton/function_modifiers/validation.py
Dev-iL: Added `DISABLE_DATA_QUALITY_CHECKS_CONFIG_KEY`

Dev-iL:
> One nit: with_data_quality_disabled() returns -> "Builder" (string literal) — should be -> Self to match the convention established in https://github.com/apache/hamilton/pull/1560.

Done.

hamilton/driver.py
```diff
@@ -1843,6 +1843,20 @@ def with_adapters(self, *adapters: lifecycle_base.LifecycleAdapter) -> Self:
...
+        """Disables all ``@check_output`` / ``@check_output_custom`` validators at graph-construction
+        time. No validator nodes are created, so there is zero runtime cost.
+
+        This is equivalent to ``.with_config({DISABLE_DATA_QUALITY_CHECKS_CONFIG_KEY: True})``.
+        Note that a subsequent ``.with_config({DISABLE_DATA_QUALITY_CHECKS_CONFIG_KEY: False})``
+        will re-enable validation, since ``with_config`` always wins on the last write.
+
+        :return: self
+        """
+        from hamilton.function_modifiers.validation import DISABLE_DATA_QUALITY_CHECKS_CONFIG_KEY
```
skrawcz: can we move this to top level?

Dev-iL: Sure - it seems like this is already imported transitively, so has no impact on loading time from making this import eager. Changed.

skrawcz (approved):

PR #1605, opened 2026-05-24
Add Driver helpers for variable lookup

### LINKED ISSUE #1138: Get nodes/variables from driver by name
This is a minor inconvenience, but currently if I'd want to get a node from the driver by name, I'd have to do:
```python
node = next(n for n in dr.list_available_variables() if node.name == "the_node_i_want")
```

A simple way to get a node by name, for example:
```python
node = dr.get_variable("the_node_i_want")
```
or
```python
node = dr.variables["the_node_i_want"]
```
The bottom option seems slightly nicer as in an interactive session `dr.variables` could be called on its own first to have a look at all the names and then called with the correct name.

**Additional context**
I looked at the code and a potential solution for the above would be to add the following to the Driver class:
```python
    @functools.cached_property
    def variables(self) -> dict[str, Variable]:
        """Returns a dictionary of all the variables (HamiltonNodes) in the graph."""
        return {k: Variable.from_node(v) for k, v in self.graph.nodes.items()}
```
or
```python
    def get_variable(self, name: str) -> Variable:
        """Returns a variable (HamiltonNode) by name."""
        return Variable.from_node(self.graph.nodes[name])
```

officialasishkumar:
Adds public `Driver` helpers for looking up variables by name and retrieving the public `HamiltonGraph` representation without iterating over `list_available_variables()`.

## Changes

- Added a cached `Driver.variables` mapping keyed by variable name.
- Added `Driver.get_variable(name)` for direct variable lookup.
- Added `Driver.get_graph()` to return the external `HamiltonGraph` API object.
- Added regression coverage for variable lookup and graph access.

## How I tested this

- `uv run --with pytest pytest tests/test_hamilton_driver.py::test_driver_variable_lookup tests/test_hamilton_driver.py::test_driver_get_graph_returns_hamilton_graph`
- `uv run ruff format --check hamilton/driver.py tests/test_hamilton_driver.py`
- `uv run ruff check hamilton/driver.py tests/test_hamilton_driver.py`

## Notes

Fixes #1138.

## Checklist


hamilton/driver.py
```diff
@@ -792,6 +793,26 @@ def list_available_variables(
             results = [Variable.from_node(n) for n in all_nodes]
         return results
 
+    @functools.cached_property
```
jernejfrank: Any particular reason for this to be cached? IMO, if the cache is required, would be better to have this as an internal variable explicitly instead of introducing hidden state

hamilton/driver.py
```diff
@@ -792,6 +793,26 @@ def list_available_variables(
...
+    def variables(self) -> dict[str, Variable]:
+        """Returns all variables in the graph keyed by name."""
+        return {
+            node_name: Variable.from_node(node_) for node_name, node_ in self.graph.nodes.items()
+        }
```
jernejfrank: On second thought, if we are exposing `get_graph`, that should already give us the names of the nodes and combining it with `get_variable` makes the API consistent and cleaner. So maybe we remove `variables` for now and can always add it in case it is really desired.

hamilton/driver.py
```diff
@@ -792,6 +793,26 @@ def list_available_variables(
...
+
+    def get_variable(self, name: str) -> Variable:
+        """Returns a variable by name.
+
+        :param name: Name of the variable to return.
+        :return: Matching HamiltonNode.
+        :raises KeyError: If the variable does not exist in this Driver's graph.
+        """
+        return self.variables[name]
```
jernejfrank: We could do build this straight from `self.graph` so it still retains the hashmap lookup speed

jernejfrank (commented):
Thanks for tackling this. I left one thought about exposing the variables directly as a dict that might make sense to do later in another PR if it is needed. Lmk, your thoughts :)

hamilton/driver.py
officialasishkumar: Removed the public variables cache and made get_variable use the graph node map directly in 4206838.

hamilton/driver.py
officialasishkumar: Removed the public variables cache and made get_variable use the graph node map directly in 4206838.

hamilton/driver.py
officialasishkumar: Removed the public variables cache and made get_variable use the graph node map directly in 4206838.

jernejfrank (approved):
Great, looks good. Thanks!

PR #1612, opened 2026-05-28
Add frontend build step to UI release process

skrawcz:
## Problem

The UI wheel uploaded to SVN was missing compiled frontend assets (`hamilton_ui/build/`). This directory is gitignored and only exists locally after running `npm run build`. A fresh clone (which is how we build release artifacts) doesn't have it, so the wheel was built without the frontend — causing HTTP 500 when serving `/`.

## Fix

1. **`scripts/apache_release_helper.py`**: When building the `ui` package, automatically runs `npm install` + `npm run build` in `ui/frontend/` and copies the output to `hamilton_ui/build/` before `flit build`.

2. **`scripts/verify-sub-packages.md`**: Documents the manual frontend build steps for reviewers building from source.

3. **`scripts/verify-sub-packages/verify_ui.sh`**: Clarifies that the 'build from source' step builds the backend only; functional tests use the pre-compiled wheel from SVN which includes the frontend.

## Impact

This requires re-cutting the UI RC (RC1) after merging. The other 3 sub-packages (SDK, LSP, Contrib) are unaffected.

### DIFF
--- a/scripts/apache_release_helper.py
+++ b/scripts/apache_release_helper.py
@@ -320,6 +320,34 @@ def create_release_artifacts(package_config: dict, version) -> list[str]:
         if os.path.exists("dist"):
             shutil.rmtree("dist")
 
+        # For the UI package, build the frontend before packaging.
+        if package_name == "apache-hamilton-ui":
+            print("Building UI frontend (npm install + npm run build)...")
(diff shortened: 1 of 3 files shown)


Dev-iL (approved):
LGTM + included in my own PR

skrawcz:
Closed — this was included in #1613 which has been merged.

PR #1613, opened 2026-05-28
Update/robustify the release scripts for subpackages (apache-hamilton-*)

Dev-iL:
This PR contains QoL improvements and fixes to various issues encountered while verifying the `apache-hamilton-{sdk,ui,lsp,contrib}` packages in the context of https://lists.apache.org/thread/d4fzmqq4jbsx5m3krdp1qbr0wh9j131s and related threads.

Includes #1612

### UI Package (`apache-hamilton-ui`)
- **Added `README.md`** for the `ui/backend` package, fixing `long_description`/`long_description_content_type` warnings during `flit build`
- **Frontend build step**: the release script now runs `npm run build` and copies compiled frontend assets into `hamilton_ui/build/` before building the wheel, ensuring the UI package includes its frontend (from #1612)
- **npm audit fix**: resolved one moderate severity vulnerability in frontend dependencies

### Release Helper (`scripts/apache_release_helper.py`)
- **`--no-sign` / `-ns`**: skips GPG signing; implicitly also skips upload since unsigned artifacts must never be published
- **`--no-upload` / `-nu`**: skips git tagging and SVN upload while still signing; useful for rehearsing the full signing step without publishing
- **`apache_id` is no longer required** when `--no-upload` (or `--no-sign`) is set
- **`uv` script dependencies declared** in the PEP 723 inline metadata block so the script can be run with `uv run`

### Verification Scripts (`scripts/verify-sub-packages/`)
- **`cd /tmp` before functional checks** in all four verify scripts (`verify_contrib.sh`, `verify_sdk.sh`, `verify_lsp.sh`, `verify_ui.sh`): prevents Python's implicit `''` (cwd) `sys.path` entry from resolving the local checkout instead of the venv-installed packages when the script is run from the repo root
- **More informative error in `verify_ui.sh`**: improved diagnostics when the UI health-check endpoint is unreachable
- **`uvx flit`** used in place of a separately installed `flit` across all verify scripts
- **TOML verification pre-commit hook** added to `ui/sdk/.pre-commit-config.yaml`

## How I tested this
After introducing these changes I could successfully run the verification **shell** scripts for all distributions as well as `uv run scripts/apache_release_helper.py --package {pkg} {ver} 0 -ns`

## Notes
I had include the version bump commit so that `apache_release_helper.py` runs correctly (with the new versions). We might want to consider making the version mismatch non-fatal when building packages with no intention to upload them.

## Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -284,6 +284,11 @@ exclude = ["*tests*"]
 [tool.setuptools.package-data]
 hamilton = ["*.json", "*.md", "*.txt"]
 
+[tool.uv.sources]
+apache-hamilton-lsp = { path = "dev_tools/language_server", editable = true }
+apache-hamilton-sdk = { path = "ui/sdk", editable = true }
+apache-hamilton-ui = { path = "ui/backend", editable = true }
+
 [tool.flit.module]
 name = "hamilton"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -67,7 +67,7 @@ experiments = [
   "fastui",
   "uvicorn",
 ]
-lsp = ["sf-hamilton-lsp"]  # TODO: switch to apache-hamilton-lsp once stable version is published
+lsp = ["apache-hamilton-lsp"]
 openlineage = ["openlineage-python"]
 pandera = ["pandera"]
 pydantic = ["pydantic>=2.0"]
@@ -77,12 +77,12 @@ pyspark = [
 ]
 ray = ["ray>=2.0.0; python_version < '3.14'", "pyarrow"]
 rich = ["rich"]
-sdk = ["sf-hamilton-sdk"]  # TODO: switch to apache-hamilton-sdk once stable version is published
+sdk = ["apache-hamilton-sdk"]
 slack = ["slack-sdk"]
 
 mcp = ["fastmcp>=3.0.0,<4"]
 tqdm = ["tqdm"]
-ui = ["sf-hamilton-ui"]  # TODO: switch to apache-hamilton-ui once stable version is published
+ui = ["apache-hamilton-ui"]
 
 # vaex -- on >=py3.11 only core part available https://github.com/vaexio/vaex/pull/2331#issuecomment-2437198176
 vaex = [
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ default_stages: [pre-commit, pre-merge-commit]
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     # Ruff version.
-    rev: v0.15.12
+    rev: v0.15.14
     hooks:
       # Run the linter.
       - id: ruff-check
@@ -43,6 +43,7 @@ repos:
       - id: requirements-txt-fixer
       # valid python file
       - id: check-ast
+      - id: check-toml
   - repo: https://github.com/Lucas-C/pre-commit-hooks
     rev: v1.5.6
     hooks:
--- a/contrib/pyproject.toml
+++ b/contrib/pyproject.toml
@@ -21,7 +21,7 @@ build-backend = "flit_core.buildapi"
 
 [project]
 name = "apache-hamilton-contrib"
-version = "0.0.8"
+version = "0.0.8"  # Keep in sync with contrib/hamilton/contrib/version.py until this becomes dynamic
 description = "Apache Hamilton's user contributed shared dataflow library."
 readme = "README.md"
 requires-python = ">=3.10, <4"
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -136,7 +136,7 @@ To build and sign artifacts without uploading to SVN or creating a git tag:
 
 ```bash
 uv run python scripts/apache_release_helper.py \
-    --package ${PACKAGE_KEY} ${VERSION} ${RC} ${APACHE_ID} --dry-run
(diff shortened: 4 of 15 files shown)


pyproject.toml
```diff
@@ -284,6 +284,11 @@ exclude = ["*tests*"]
 [tool.setuptools.package-data]
 hamilton = ["*.json", "*.md", "*.txt"]
 
+[tool.uv.sources]
+apache-hamilton-lsp = { path = "dev_tools/language_server", editable = true }
```
skrawcz, line 288: do we want this always? is this standard uv practice for a setup like ours? or?

contrib/hamilton/contrib/version.py
```diff
@@ -15,4 +15,4 @@
 # specific language governing permissions and limitations
 # under the License.
 
-VERSION = (0, 0, 8)
+VERSION = (0, 0, 9)
```
skrawcz: we can't commit the version bumps til we vote on them -- so you can drop these.

Dev-iL: To my understanding, we can't _release the versions_ before voting. Note how in [Airflow main](https://github.com/apache/airflow/blob/main/pyproject.toml#L59) the package version is 3.3.0 (the next major version) and not 3.2.2 (the latest released).

Dev-iL: Also these changes serve a real purpose: they ensure we're installing from source and not from pypi

pyproject.toml
jernejfrank: there's this thing they rolled out for having a multi-repo: https://docs.astral.sh/uv/concepts/projects/workspaces/#workspace-sources,

The nice thing is you can develop on all, because they ar editable without needing to manually install them as editable. But this is `uv` specific, so anybody using `pip` will not see this/get this functionality.

Dev-iL: > do we want this?

In general yes, but not necessarily how I did it here. These are uv [path sources](https://docs.astral.sh/uv/concepts/dependencies/#path), they tell uv to resolve `apache-hamilton-lsp/sdk/ui` (and core, transitively) from local source rather than PyPI when it locks/syncs the repo. I verified it with `uv lock`: core and all three sub-packages pin to `editable = "."`/`"ui/sdk"`/etc. so nothing comes from pypi - it's what fixes resolution on CI.

>  always?

"Always" is fine because `[tool.uv.sources]` is dev/resolution-time only. pip doesn't read it, and uv strips it from published metadata, so released wheels and anyone doing `pip install apache-hamilton[ui]` are unaffected. It matters when working in this repo with uv, which is exactly when we want local source.

We don't need to repeat it per sub-package because the root sources cover the resolution, and the per-package CI jobs install core via `pip install -e` directly anyway.

> is this standard uv practice for a setup like ours? or?

The heavier alternative is a real uv [workspace](https://docs.astral.sh/uv/concepts/projects/workspaces/) (`[tool.uv.workspace]` + `{ workspace = true }`), which is what [Airflow](https://github.com/apache/airflow/blob/main/pyproject.toml#L1694) does. It'd give us a single shared lockfile and `uv run --package`, but nothing in our CI uses those, and path sources already get us the local resolution we need. Something to think about in the future...

Dev-iL: @jernejfrank looks like you beat me to it :)

I don't mind rewriting this as a workspace. However, consider several words of caution by the AI:

> A workspace is a legitimate, widely-used pattern (Airflow, most large Python monorepos). Switching to a workspace in this case is mostly unnecessary, with a slight lean toward net-negative. It's unnecessary because the thing a workspace buys you — local cross-package resolution — you already have via path sources, proven by the lock. The other workspace perks (`uv run --package`, one `uv sync` to set up every member at once) aren't used anywhere in your CI or scripts.
> 
> Where it tips from "unnecessary" toward "slightly worse for *this* repo":
> 
> - **Tighter coupling for no reason.** A workspace forces all members into one universal resolution, unconditionally — including each member's dev/test groups. A version conflict anywhere (say, something in `ui/backend`'s Django stack vs. core's constraints) would break the whole-workspace lock, even for someone only touching core. The current path sources put the sub-packages behind *optional extras*, so a dev syncing plain core never drags in Django. The workspace would remove that gating.
> - **More config to keep correct, with a footgun.** You'd add `[tool.uv.workspace]` plus a members list, and you'd have to remember to exclude `examples/*/pyproject.toml` — those would otherwise get silently pulled in as members and could break resolution.
> - **It fights the release model slightly.** Each package builds with its own flit backend, versions independently, and goes through ASF voting separately. That's "independent packages sharing a repo," which the path-source approach models more honestly than a single-lock workspace.
> 
> The threshold where it would flip to worth it: many tightly-coupled members developed and released together, shared dev tooling you want bootstrapped in one command, or genuine need for `uv run --package`. This repo is four loosely-coupled packages on separate release cadences. It isn't there.
> 
> So: keep the path sources. A workspace here would be effort and coupling spent to re-acquire something you already have.

contrib/hamilton/contrib/version.py
Dev-iL: Fixed.

pyproject.toml
jernejfrank: no that makes sense, I played with the workspace stuff before and it did introduce a lot of overhead. Doing the paths should be more than enough 👍

jernejfrank (approved):
Nice, lgtm!

PR #1614, opened 2026-05-29
Fix async client swallowing backend errors during flush

ArnavBalyan:
## Changes
 - Async client flush catches HTTPError but the call uses aiohttp, so the except never matched. Backend errors propagated up unhandled and killed the worker task.
 - Add a raise so failures actually surface.
 - Switch to except aiohttp.ClientResponseError, matching every other async method in this file.

## How I tested this
 - UT
## Checklist


ArnavBalyan:
cc @skrawcz ty :)

skrawcz (approved):
LGTM. Clean fix for a real bug — `requests.HTTPError` was dead code in the async path. The switch to `aiohttp.ClientResponseError` matches every other async method in the file, and the re-raise ensures errors are logged before propagating.

Issue #918, opened 2024-05-27
chore: Remove numpy dependency


zilto:
You can use Hamilton without numpy so there's no reason to have it as a dependency.

By searching `numpy` in `hamilton/**.py`, there are 3 results:
- `hamilton/base.py` where `NumpyMatrixResult` is defined. This could be moved to the `hamilton/plugins/h_numpy.py`. We would need to solve the import pattern.
- `hamilton/telemetry.py` there's a check for the `base.NumpyMatrixBuilder`. We can probably just remove or stitch with the above change?
- `hamilton/log_setup.py` where `np.seterr` is used to ignore some numpy error. This should be safe to delete?

By searchin `numpy` in `tests/**.py` there are more results. Numpy is used to test the 3 previous elements, but also as an input for certain tests and for `@check_output` stuff.
- The easy fix is to keep numpy in the test dependencies.
- We could move certain tests to `tests/plugins/test_h_numpy.py`

Also, we would have to make sure that the examples have an explicit dependency on numpy if they are using it.

radujica:
looks like the same could be said about pandas? it does pop up in more places, but none look core

elijahbenizzy:
> looks like the same could be said about pandas? it does pop up in more places, but none look core

Yes! It's only left there for legacy reasons. One thing I've been meaning to do is add a `hamilton[raw]` that doesn't install it or numpy. The library should have zero dependencies. That's dependent on this.

Dev-iL:
## Tentative work plan

~~~md
# Definition: Hamilton 1.9x transitional rung — pandas/numpy decoupling preparation (apache/hamilton#918)

## 1. Intent & Context

- **Goal:** Land every preparatory, non-breaking change on the 1.x line so that Hamilton 2.0 can remove pandas/numpy from core dependencies as a near-pure deletion changeset. Pandas and numpy REMAIN hard dependencies in this release; nothing observable breaks for any existing user.
[...]

PR #1615, opened 2026-05-29
fix: reduce driver/base import overhead

### LINKED ISSUE #1246: [ergonomics] profile driver and base module imports
# Current behavior

It takes ~2 seconds to do:

> from hamilton import driver, base

We should profile this and see where the time is being spent.

Latest hamilton on macOS, python 3.10.

# Expected behavior
It should load in < 1 second.

# Additional context
This is a quality of life task.

This task requires: 

1. to use a profiler to figure out where time is being spent
2. report the results
3. provide recommendations

akashmalbari:
Closes #1246.

This PR reduces import-time overhead for `from hamilton import driver, base` by deferring import-time work that is not needed during the initial `driver`/`base` import path.

## Changes

- Profiled `from hamilton import driver, base` using `python -X importtime` and `cProfile`.
- Identified unnecessary import-time overhead in the `driver`/`base` import path.
- Deferred the relevant import work so it only happens when needed.
- Preserved existing public import behavior for:
  - `from hamilton import driver, base`
  - `from hamilton.driver import Driver`
  - `from hamilton import base`

## How I tested this

Import smoke test:

```bash
python -c "from hamilton import driver, base; print(driver.Driver); print(base.DictResult)"

## Profiling

Command used:

```bash
python -X importtime -c "from hamilton import driver, base"
python -c "import cProfile; cProfile.run('from hamilton import driver, base', '/tmp/hamilton_import_before.prof')"
python -c "import cProfile; cProfile.run('from hamilton import driver, base', '/tmp/hamilton_import_after.prof')"

Before: 1.068s importtime, 0.894s cProfile.
After: 0.111s importtime, 0.188s cProfile.

### CODE REVIEW

File: tests/test_imports.py [hunk 1]
```diff
@@ -0,0 +1,27 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+
+def test_driver_base_public_imports():
```
ArnavBalyan: The test does not cover future regressions, should we add an assert to ensure new code cannot eagerly load the dependency.

```
  def test_driver_import_does_not_load_heavy_modules():
      import sys
      for mod in ("pandas", "numpy"):
          sys.modules.pop(mod, None)
      import hamilton.driver 
      for mod in ("pandas", "numpy"):
          assert mod not in sys.modules, f"{mod} was loaded eagerly"
```
Along the above lines

hamilton/base.py
```diff
@@ -16,29 +16,54 @@
...
-    from . import htypes, node
+    from . import htypes
 except ImportError:
-    import node
+    import htypes
 
 logger = logging.getLogger(__name__)
 
 
+def _get_pandas():
```
ArnavBalyan: There are many _get_dependency functions hard to track/standardize in the future, can we put them behind a shared helper.

Along the following lines:
```
  @cache
  def _lazy_import(module: str, attr: str | None = None):
      """Lazy import a module. Cached after 1st call."""
      mod = importlib.import_module(module)
      return mod if attr is None else getattr(mod, attr)


  pd  = _lazy_import("pandas")
  np  = _lazy_import("numpy")
  pde = _lazy_import("pandas.core.indexes.extension")
```

ArnavBalyan (commented):
Lgtm thanks for the change, some minor comments. The perf improvement looks great!

akashmalbari:
Thanks for the review! I addressed both comments and resolved the latest conflict with `main`.

- Added a subprocess-based regression test to ensure importing `hamilton.driver` does not eagerly load `pandas` or `numpy`.
- Replaced the ad-hoc lazy dependency helpers in `hamilton/base.py` with a shared cached `_lazy_import(...)` helper.
- Resolved the `hamilton/driver.py` conflict by keeping both `with_data_quality_disabled(...)` from `main` and the lazy/string annotation for `with_materializers(...)`.

Validation run:

- `.venv/bin/ruff check hamilton/driver.py hamilton/base.py tests/test_imports.py`
- `.venv/bin/python -m pytest tests/test_imports.py -q`
- Import smoke check for `from hamilton import base, driver`

I also tried the relevant upstream data-quality tests locally, but this environment could not collect them cleanly because extension loading hit the local `xgboost` / missing `libomp.dylib` issue.

ArnavBalyan (approved):
Thanks, cc @skrawcz

Dev-iL (changes requested):
There's an old issue about making numpy and pandas optional dependencies (https://github.com/apache/hamilton/issues/918), so importing them lazily could be considered a step in that direction. Users (like myself) who use polars will not have to pay the penalty of loading a backend they don't use - which is great.
 
Other than that:
- Please don't push `uv.lock` - I think the PPMC needs to actively decide that we're going for this.
- The lazy loading mechanism introduced in this PR moves the runtime cost of importing certain deps from import time to usage time. My question is - why is this the concern of hamilton itself, as opposed to the users of hamilton? Before this change, if a user wanted to lazy-load hamilton, they would import and instantiate the driver wherever it's used. With this change, we're "forcing" users to lazy-load pandas & numpy - and the runtime increase during usage might be surprising. I am not convinced that we should decide for users when to load the deps.

I personally think that the right fix for the import is making numpy and pandas truly optional (per the issue above), and as a result we won't have to add mechanisms for making their imports lazy.

akashmalbari:
Thanks for the detailed review.

I removed `uv.lock` from the PR.

On the lazy-loading concern: understood. My intent here was to keep the change narrow for #1246 by avoiding the import-time cost for users who import `hamilton.driver` / `hamilton.base` but do not use pandas or numpy-backed paths. That said, I see your point that this shifts the cost from import time to first use, and that making pandas/numpy truly optional as part of #918 may be the cleaner long-term direction.

Before changing the implementation further, would you prefer that I:

1. keep this PR as a narrow import-time improvement,
2. reduce this PR to profiling/test coverage only and defer the dependency behavior to #918, or
3. close this PR in favor of a broader #918-style optional dependency change?

Happy to adjust based on the preferred direction.

Dev-iL (approved):
> I removed `uv.lock` from the PR.

Thanks - this was my only immediate concern.

> Happy to adjust based on the preferred direction.

We're all ultimately volunteers here, and you are free to choose what you work on. I believe it is in the benefit of the project to split out numpy & pandas, but this is by no means a "requirement" from you.

akashmalbari:
Thanks, that makes sense. I’ll keep this PR focused on the import-time improvement for #1246 and leave the broader pandas/numpy optional dependency work to #918 or a future dedicated PR.

PR #1616, opened 2026-05-29
Include metadata in numpy/polars cache fingerprints to prevent collisions

skrawcz:
## Summary

Fix hash collisions in the caching subsystem's fingerprinting for numpy arrays and polars DataFrames.

### Problem

- `hash_numpy_array` used only `obj.tobytes()`, which discards shape and dtype. Arrays with identical raw bytes but different shapes (e.g., `shape=(6,)` vs `shape=(2,3)`) or different dtypes (e.g., `float32(1.0)` vs `int32(1065353216)`) produced identical cache keys.

- `hash_polars_dataframe` used only `obj.hash_rows()`, which discards column names. DataFrames with identical cell values but different schemas produced identical cache keys.

Both could cause the cache to silently return incorrect results from a previous execution.

### Fix

- `hash_numpy_array`: prepend `f"{obj.shape}:{obj.dtype}"` to the bytes before hashing
- `hash_polars_dataframe`: include column names and dtypes (schema) alongside row hashes

### Backwards compatibility

This changes hash output for numpy arrays and polars DataFrames. Existing caches will **miss** (different hash = recomputation), not produce incorrect results. Users will see a one-time recomputation after upgrading but no manual cache clearing is needed.

### Tests

Added tests verifying:
- Different shapes produce different hashes
- Different dtypes with same bit pattern produce different hashes
- Different column names produce different hashes
- Identical data still produces identical hashes

Reported-by: Dem0

### DIFF
--- a/hamilton/caching/fingerprinting.py
+++ b/hamilton/caching/fingerprinting.py
@@ -261,11 +261,15 @@ def hash_pandas_obj(obj, *args, depth: int = 0, **kwargs) -> str:
 
 @hash_value.register(h_databackends.AbstractPolarsDataFrame)
 def hash_polars_dataframe(obj, *args, depth: int = 0, **kwargs) -> str:
-    """Convert a polars dataframe, series, or index to
-    a list of hashes then hash it.
+    """Convert a polars dataframe to a hash that includes column names
+    and dtypes (schema) alongside row hashes. This prevents collisions
+    between DataFrames with identical cell values but different schemas.
     """
-    hash_per_row = obj.hash_rows()
-    return hash_sequence(hash_per_row.to_list(), depth=depth + 1)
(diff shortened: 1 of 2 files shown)


elijahbenizzy (approved):
And closing out my duplicate

Dev-iL (approved):
If we're redoing hashes - perhaps it might be worth switching from sha224/md5 to [xxh3](https://github.com/Cyan4973/xxHash)?

ArnavBalyan (approved):
Lgtm ty for the change

skrawcz:
> If we're redoing hashes - perhaps it might be worth switching from sha224/md5 to [xxh3](https://github.com/Cyan4973/xxHash)?

separate PR? would have to validate license, etc.

PR #1617, opened 2026-05-30
fix(caching): include shape/dtype/schema in fingerprint hashing

elijahbenizzy:
## Summary

The numpy, pandas, and polars fingerprint hashes in `hamilton/caching/fingerprinting.py` previously hashed only the underlying values, dropping shape, dtype, column names, and Series names. This caused cache key collisions between semantically distinct inputs:

- `np.array([1,2,3,4,5,6])` vs `np.array([[1,2,3],[4,5,6]])` (same `.tobytes()`, different shapes)
- `np.array([1,2,3], int32)` vs `np.array([1,0,2,0,3,0], int16)` (same bytes, different dtypes)
- `pd.DataFrame({"customer_revenue": [100,200]})` vs `pd.DataFrame({"product_cost": [100,200]})` (same row values, different column names)
- Same pattern for polars

With caching enabled, the cache returned the prior result for the second input silently — producing incorrect outputs with no warning.

## What this PR does

- `hash_numpy_array`: prepends `shape` and `dtype` to the byte buffer.
- `hash_polars_dataframe`: includes `obj.schema` alongside row hashes.
- `hash_pandas_obj`: includes `(column, dtype)` pairs (DataFrame) or `(name, dtype)` (Series/Index) alongside row hashes.

## Test plan

- 8 new regression tests in `tests/caching/test_fingerprinting.py` covering each collision case + identical-input determinism.
- Updated `test_hash_pandas` and `test_hash_numpy` golden hashes for the new schema-aware computation.
- All 117 `tests/caching/` tests pass locally.

### DIFF
--- a/hamilton/caching/fingerprinting.py
+++ b/hamilton/caching/fingerprinting.py
@@ -243,11 +243,33 @@ def hash_set(obj, *args, depth: int = 0, **kwargs) -> str:
     return _compact_hash(hash_object.digest())
 
 
+def _pandas_schema_repr(obj) -> str:
+    """Stringify the schema-defining metadata of a pandas DataFrame/Series/Index.
+
+    Used to ensure objects with the same row values but different column
+    names or dtypes hash differently.
+    """
+    import pandas as pd
+
+    if isinstance(obj, pd.DataFrame):
+        return "DataFrame" + repr(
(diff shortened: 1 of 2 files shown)


elijahbenizzy:
Fixed in #1616

PR #1618, opened 2026-06-01
Include tests and examples in sub-package source distributions

skrawcz:
## Problem

Vote feedback on RC0: tests and examples were excluded from the source tarballs. Per Apache release policy, voters need to be able to build from source and run tests to validate.

## Changes

**Tests included in sdist (all 4 packages):**
- `ui/sdk`: removed `tests/**` from exclude, added to include
- `dev_tools/language_server`: same
- `contrib`: removed `tests/**` from exclude
- `ui/backend`: already included (no change)

**Examples included via symlinks (SDK + Contrib):**
- `ui/sdk/examples` → `../../examples/hamilton_ui/`
- `contrib/examples` → `../examples/contrib/`

**Binary files excluded from sdist:**
- `*.parquet`, `*.pkl`, `*.ipynb`, `*.png` excluded to keep tarballs source-only

## Verification

Built all 4 sdists and confirmed:
- Tests present in all 4 source tarballs
- Examples present in SDK and Contrib tarballs (source files only, no binaries)
- All included files have Apache license headers

## Context

Addresses feedback from Jernej Frank's -1 votes on SDK, LSP, and Contrib RC0:
> In the main package we add tests to the source tarball and sdist, but here they are omitted.


Dev-iL (approved):
The PR change is verified:
- tests are present in all 4 sdists.
- The `contrib` package has 4 test files co-located inside the package (not a top-level `tests/` dir), which is correct given how that package is structured.

contrib/pyproject.toml
jernejfrank, line 1: in th emain pacakge we all include `scripts` folder. `contrib` package also has script, do we wan to include them?

jernejfrank (commented):
Great, there's this one script that seems like we need to include in the source in the `contrib` package. Otherwise, rest looks all good!

contrib/pyproject.toml
skrawcz: Good call — added `"scripts/**"` to the contrib sdist include list. It contains `version_check.py` which is useful for release validation.

elijahbenizzy (approved):

PR #1624, opened 2026-06-07
Remove `flit` from prereq check since handled by uvx

Dev-iL:
Since `flit` is executed via `uvx`, checking for `uv` is sufficient.

## Changes
Remove `flit` from prereq check in the 4 subpackage verification scripts.

## How I tested this
Successfully ran `scripts/verify-sub-packages/verify_sdk.sh 0.9.0 1` in a venv where flit is not installed.

## Notes
N/A

## Checklist


jernejfrank (approved):
Makes sense.

PR #1629, opened 2026-06-08
Vectorize pandas and polars DataFrame hashing

Dev-iL:
Split 2 of 3 of #1619 (stacked on #1628)

1. centralize hashing through a single chokepoint + close type collisions (#1628).
2. **this PR** — vectorize the pandas/polars DataFrame paths.
3. swap the algorithm to `xxhash.xxh3_128`.

## What this does

Replaces the Python-level row iteration in the pandas and polars fingerprint paths with a single vectorized buffer hash, keeping everything else (the `_hash_bytes` chokepoint, the algorithm, the collision-prevention type tags) unchanged from PR1.

**pandas** — `hash_pandas_object(obj).to_dict()` fed through an ordered `hash_mapping` (one Python `hash_value` call per row) → `hash_pandas_object(obj).values.tobytes()` hashed in one `_hash_bytes` call. Column names and dtypes are folded into the hash so frames with identical cell values but different schemas do not collide. The old docstring claiming row order "doesn't matter" was incorrect; the new path is explicitly order-sensitive.

**polars** — `obj.hash_rows().to_list()` fed through `hash_sequence` (one Python `hash_value` call per element) → `obj.hash_rows().to_numpy().tobytes()` hashed in one `_hash_bytes` call. The `schema_hash + row_hash` combine from #1616/#1628 is preserved.

The change is purely mechanical: replace the Python loop with a buffer read. No algorithm change, no new dependency.

## Benchmark

<details><summary>Benchmark code</summary>
<p>

~~~python
# benchmark_fingerprinting.py
"""Corroborating benchmark for the DataFrame fingerprinting speedup.

Compares the new vectorized hashing in
:mod:`hamilton.caching.fingerprinting` against the old per-row approach for
both supported DataFrame backends:

- pandas: vectorized hash of the ``hash_pandas_object(...).values`` buffer vs
  ``.to_dict()`` fed through an ordered ``hash_mapping`` (Python row loop).
- polars: vectorized hash of the ``hash_rows().to_numpy()`` buffer vs
  ``hash_rows().to_list()`` fed through a per-element ``hash_sequence``.

The structural "no per-row loop" assertions in the test suite are the hard
gate; this script is corroborating evidence with a generous floor to avoid the
flakiness of an absolute-time threshold. Run directly:

    python benchmarks/benchmark_fingerprinting.py
"""

import os
import platform
import re
import time

import pandas as pd

from hamilton.caching import fingerprinting as fp

[...]

PR #1630, opened 2026-06-08
Switch the fingerprint algo to xxh3_128

Dev-iL:
Split 3 of 3 of #1619 (stacked on #1629)

1. centralize hashing through a single chokepoint + close type collisions (#1628).
2. vectorize the pandas/polars DataFrame paths (#1629).
3. **this PR** — swap the algorithm to `xxhash.xxh3_128`.

## What this does

Change in `_hash_bytes`: when xxhash is installed, the runtime path changes `hashlib.md5(data).digest()` → `xxhash.xxh3_128(data).digest()`.  `xxh3_128` is a non-cryptographic hash designed for speed. It produces a 16-byte (128-bit) digest, the same width as md5, so the base64url-encoded fingerprints stay 24 characters and all downstream interfaces (cache keys, `data_version` strings) are unaffected.

Adds `xxhash>=0.8.0` as an optional dependency in `pyproject.toml` (under the `performance` extra).

## Benchmark

<details><summary>Benchmark code</summary>
<p>

~~~python
# benchmark_hash_algorithm.py
"""Benchmark: xxh3_128 vs md5 on the vectorized DataFrame fingerprint path.

Holds the (vectorized) implementation constant and varies only the hashing
algorithm applied by ``_hash_bytes``, to quantify what switching md5 -> xxh3_128
buys:

  * "hash-step": md5 vs xxh3_128 over the row-hash buffer alone. This is the
    raw algorithm advantage and the ceiling on any end-to-end gain.
  * "end-to-end": the full vectorized fingerprint, comparing the current
    xxh3_128 functions against an md5 reconstruction of the same construction.
    Buffer construction (``hash_pandas_object`` / ``hash_rows``) is identical
    under either algorithm, so this is the realistic improvement on the hot
    path once that fixed cost is included.

Informational only (no pass/fail gate). Run directly:

    python benchmarks/benchmark_hash_algorithm.py
"""

import base64
import hashlib
import os
import platform
import re
import time

import numpy as np
import pandas as pd
import xxhash

from hamilton.caching import fingerprinting as fp

# Swept so the realized speedup can be read as a function of buffer size.
SIZES = (500, 5_000, 50_000, 500_000, 5_000_000)


def _cpu_model() -> str:
    try:
        with open("/proc/cpuinfo") as f:
            for line in f:
                if line.startswith("model name"):
                    return line.split(":", 1)[1].strip()
    except OSError:
        pass
    return platform.processor() or "unknown"


def _ram_info() -> str:
    import subprocess

    total = "unknown"
    try:
        with open("/proc/meminfo") as f:
            for line in f:
                if line.startswith("MemTotal"):
                    kb = int(re.search(r"\d+", line).group())
                    total = f"{kb / 1024 / 1024:.0f} GiB"
                    break
    except OSError:
        pass
    detail = ""
    try:
        out = subprocess.check_output(
            ["dmidecode", "-t", "memory"], text=True, stderr=subprocess.DEVNULL,
        )
        typ = re.search(r"^\s*Type:\s*(DDR\S*)", out, re.MULTILINE)
        spd = re.search(r"^\s*Configured Memory Speed:\s*(\d+\s*\S+)", out, re.MULTILINE)
        if not spd:
            spd = re.search(r"^\s*Speed:\s*(\d+\s*\S+)", out, re.MULTILINE)
        parts = [m.group(1) for m in (typ, spd) if m]
        if parts:
            detail = f" ({' '.join(parts)})"
    except (OSError, subprocess.CalledProcessError):
        pass
    return f"{total}{detail}"


def _print_env() -> None:
    import sys

    import polars as pl

    print("== Environment ==")
    print(f"  CPU      : {_cpu_model()} ({os.cpu_count()} logical cores)")
    print(f"  RAM      : {_ram_info()}")
    print(f"  Platform : {platform.system()} {platform.release()} ({platform.machine()})")
    print(f"  Python   : {sys.version.split()[0]}")
    print(f"  numpy    : {np.__version__}")
    print(f"  pandas   : {pd.__version__}")
    print(f"  polars   : {pl.__version__}")
    print(f"  xxhash   : {xxhash.VERSION}")
    print()


def _md5_compact(data: bytes) -> str:
    return base64.urlsafe_b64encode(hashlib.md5(data).digest()).decode()


def _numpy_md5(obj) -> str:
    """The numpy path, but hashing the buffer with md5.

    Unlike the DataFrame paths there is no per-row baseline to remove: numpy has
    always hashed ``shape:dtype`` + raw ``tobytes()`` in one shot. Even so the
    end-to-end speedup is bounded well below the raw algorithm speedup, because
    materializing the buffer (``tobytes()`` + concatenation) is a memcpy that
    costs about as much as md5 hashing it.
    """
    metadata = f"{obj.shape}:{obj.dtype}".encode()
    return _md5_compact(b"bytes:" + metadata + obj.tobytes())


def _vectorized_pandas_md5(obj) -> str:
    """The vectorized pandas path, but hashing the buffer with md5."""
    from pandas.util import hash_pandas_object

    row_hashes = hash_pandas_object(obj).values.tobytes()
    if hasattr(obj, "columns"):
        schema = f"{list(obj.columns)}:{[str(dtype) for dtype in obj.dtypes]}"
    else:
        schema = f"{getattr(obj, 'name', None)}:{obj.dtype}"
    return _md5_compact(schema.encode() + row_hashes)


def _vectorized_polars_md5(obj) -> str:
    """The vectorized polars path, but hashing the buffers with md5."""
    schema_str = ",".join(f"{name}:{dtype}" for name, dtype in obj.schema.items())
    schema_hash = _md5_compact(b"bytes:" + schema_str.encode())
    row_hash = _md5_compact(b"bytes:" + obj.hash_rows().to_numpy().tobytes())
    return _md5_compact(schema_hash.encode() + row_hash.encode())


def _time(fn, obj, repeats: int = 3) -> float:
    fn(obj)  # warmup
    best = float("inf")
    for _ in range(repeats):
        start = time.perf_counter()
        fn(obj)
        best = min(best, time.perf_counter() - start)
    return best


def _report_hash_step(label: str, n_rows: int, buffer: bytes) -> None:
    mib = len(buffer) / 1024 / 1024
    md5_t = _time(lambda b: hashlib.md5(b).digest(), buffer)
    xxh_t = _time(lambda b: xxhash.xxh3_128(b).digest(), buffer)
    print(
        f"[{label} n={n_rows:>9,}] hash-step  md5 {md5_t * 1e3:8.3f} ms ({mib / md5_t:>7,.0f} MiB/s)"
        f"  xxh3 {xxh_t * 1e3:7.3f} ms ({mib / xxh_t:>7,.0f} MiB/s)  speedup {md5_t / xxh_t:5.1f}x"
    )


def _report_end_to_end(label: str, n_rows: int, md5_fn, xxh_fn, obj) -> None:
    md5_t = _time(md5_fn, obj)
    xxh_t = _time(xxh_fn, obj)
    print(
        f"[{label} n={n_rows:>9,}] end-to-end md5 {md5_t * 1e3:8.1f} ms"
        f"  xxh3 {xxh_t * 1e3:8.1f} ms  speedup {md5_t / xxh_t:5.2f}x"
    )


def _columns(n_rows: int) -> dict:
    return {
        "a": range(n_rows),
        "b": [float(i) for i in range(n_rows)],
        "c": [f"row-{i}" for i in range(n_rows)],
    }


def main() -> None:
    _print_env()

    from pandas.util import hash_pandas_object

    try:
        import polars as pl
    except ImportError:
        pl = None
        print("[polars] not installed; skipping polars")

    for n_rows in SIZES:
        columns = _columns(n_rows)

        arr = np.arange(n_rows, dtype=np.int64)
        _report_hash_step("numpy ", n_rows, arr.tobytes())
        _report_end_to_end("numpy ", n_rows, _numpy_md5, fp.hash_value, arr)

        df = pd.DataFrame(columns)
        _report_hash_step("pandas", n_rows, hash_pandas_object(df).values.tobytes())
        _report_end_to_end("pandas", n_rows, _vectorized_pandas_md5, fp.hash_pandas_obj, df)

        if pl is not None:
            pdf = pl.DataFrame(columns)
            _report_hash_step("polars", n_rows, pdf.hash_rows().to_numpy().tobytes())
            _report_end_to_end(
                "polars", n_rows, _vectorized_polars_md5, fp.hash_polars_dataframe, pdf
            )


if __name__ == "__main__":
    main()

~~~

</p>
</details> 

<details><summary>Plotting code</summary>
<p>

~~~python
# plot_benchmarks.py
"""Generate Plotly charts for PR2 and PR3 benchmark results.

Reads hardcoded benchmark data (not recomputed) and writes two PNG files
into the same directory as this script.

    python benchmarks/plot_benchmarks.py
"""

from pathlib import Path

import plotly.graph_objects as go

OUT_DIR = Path(__file__).parent

SIZE_LABELS = ["500", "5 K", "50 K", "500 K", "5 M"]

COLORS = {
    "pandas": "#1f77b4",
    "polars": "#ff7f0e",
    "numpy": "#2ca02c",
}

# ── PR2 data (baseline → vectorized) ────────────────────────────────────
PR2 = {
    "pandas": [4.2, 11.0, 12.5, 9.8, 6.0],
    "polars": [13.1, 78.1, 382.0, 431.0, 209.0],
}

# ── PR3 data (md5 → xxh3_128, end-to-end) ───────────────────────────────
PR3 = {
    "numpy": [1.36, 3.21, 6.45, 1.44, 1.55],
    "pandas": [1.03, 0.88, 0.98, 1.00, 1.00],
    "polars": [1.25, 1.84, 2.26, 2.26, 1.67],
}


def _build_chart(data: dict, title: str, log_y: bool = False) -> go.Figure:
    fig = go.Figure()

    for backend, speedups in data.items():
        fig.add_trace(
            go.Scatter(
                name=backend,
                x=SIZE_LABELS,
                y=speedups,
                mode="lines+markers+text",
                text=[f"{s:.1f}×" for s in speedups],
                textposition="top center",
                textfont=dict(size=12),
                line=dict(color=COLORS[backend], width=3),
                marker=dict(size=10),
            )
        )

    fig.update_layout(
        title=dict(text=title, x=0.5, y=0.95, yanchor="top"),
        legend=dict(
            yanchor="top",
            y=0.98,
            xanchor="left",
            x=0.02,
            bgcolor="rgba(255,255,255,0.8)",
        ),
        template="plotly_white",
        height=500,
        width=800,
        margin=dict(t=60, b=60),
    )
    fig.update_xaxes(title_text="Input size (rows)")
    fig.update_yaxes(title_text="Speedup (×)", type="log" if log_y else "linear")

    return fig


def main() -> None:
    pr2_fig = _build_chart(
        PR2,
        title="PR2: Vectorized hashing — speedup over per-row baseline",
        log_y=True,
    )
    pr2_path = OUT_DIR / "PR2_chart.png"
    pr2_fig.write_image(str(pr2_path), scale=2)
    print(f"Wrote {pr2_path}")

    pr3_fig = _build_chart(
        PR3,
        title="PR3: xxh3_128 vs md5 — end-to-end speedup",
    )
    pr3_path = OUT_DIR / "PR3_chart.png"
    pr3_fig.write_image(str(pr3_path), scale=2)
    print(f"Wrote {pr3_path}")


if __name__ == "__main__":
    main()

~~~

</p>
</details> 

`benchmark_hash_algorithm.py` isolates the algorithm swap from the vectorization (#1629), holding the implementation constant and varying only the hash. The raw algorithm is ~8–27× faster, but **the end-to-end gain depends on how much of total time the hash step occupies**:

| rows | backend | md5 (ms) | xxh3 (ms) | speedup |
|-----:|---------|--------:|---------:|--------:|
| 500       | numpy  | 0.03    | 0.02    | **1.4×** |
| 500       | pandas | 1.0     | 1.0     | **1.0×** |
| 500       | polars | 0.1     | 0.1     | **1.3×** |
| 5,000     | numpy  | 0.1     | 0.03    | **3.2×** |
| 5,000     | pandas | 2.9     | 3.3     | **0.9×** |
| 5,000     | polars | 0.2     | 0.1     | **1.8×** |
| 50,000    | numpy  | 0.6     | 0.1     | **6.5×** |
| 50,000    | pandas | 32.3    | 33.0    | **1.0×** |
| 50,000    | polars | 0.9     | 0.4     | **2.3×** |
| 500,000   | numpy  | 9.4     | 6.5     | **1.4×** |
| 500,000   | pandas | 382     | 382     | **1.0×** |
| 500,000   | polars | 9.5     | 4.2     | **2.3×** |
| 5,000,000 | numpy  | 96.3    | 62.0    | **1.6×** |
| 5,000,000 | pandas | 6,434   | 6,405   | **1.0×** |
| 5,000,000 | polars | 121     | 72.6    | **1.7×** |

> CPU: Intel i7-3770 @ 3.40 GHz · 32 GiB DDR3-1600 · pandas 3.0.3 · polars 1.41.2 · xxhash 3.7.0 · Python 3.14.2

PR3_chart

**pandas DataFrames: ~1× (negligible) at all sizes.** `hash_pandas_object` takes 1–6,400 ms; the hash step takes 0.006–52 ms with md5. Hashing is ≤1.5% of end-to-end time. The algorithm swap is invisible here.

**polars DataFrames: 1.3–2.3×.** `hash_rows` is fast enough that the hash step is a meaningful fraction of end-to-end time. The gain grows with size as the hash step's share increases.

**numpy arrays: 1.4–6.5×.** Peaks at mid sizes (~50k) where `tobytes()` is negligible and hashing dominates; at very small sizes per-call overhead limits the gain, at large sizes the memcpy dominates.

**Scalars, sequences, mappings (not benchmarked in isolation):** estimated ~1.4–2×, dominated by per-call overhead and base64 encoding. These are the most frequent fingerprint calls in a typical DAG (node inputs, configs, primitives).

## Why this is still worth landing

1. **Every `_hash_bytes` call gets faster** — the benefit is broadest on the many small-buffer paths (scalars, sequences, configs) even though it's invisible on pandas DataFrames.
1. **Semantic fit.** md5 is a (broken) cryptographic hash misused for speed; xxh3_128 is a non-cryptographic hash designed for exactly this use case.

## Testing

Pinned-digest literals in `test_fingerprinting.py` are updated to the new xxh3_128 values. Relational tests (must-differ, must-match, cross-type collision) and the full caching suite pass unchanged.

## Checklist

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -226,33 +226,3 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
-
--------------------------------
-
-This product depends on xxhash (the python-xxhash package, https://github.com/ifduyue/python-xxhash),
-which is licensed under the BSD 2-Clause License.
-
-BSD 2-Clause License
-
-Copyright (c) 2014-2024, Yue Du
-All rights reserved.
-
-Redistribution and use in source and binary forms, with or without modification,
-are permitted provided that the following conditions are met:
-
-    * Redistributions of source code must retain the above copyright notice,
-      this list of conditions and the following disclaimer.
-    * Redistributions in binary form must reproduce the above copyright notice,
-      this list of conditions and the following disclaimer in the documentation
-      and/or other materials provided with the distribution.
-
-THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
-ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
-WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
-FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/LICENSE
+++ b/LICENSE
@@ -226,3 +226,33 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+-------------------------------
+
+This product depends on xxhash (the python-xxhash package, https://github.com/ifduyue/python-xxhash),
+which is licensed under the BSD 2-Clause License.
+
+BSD 2-Clause License
+
+Copyright (c) 2014-2024, Yue Du
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice,
+      this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice,
+      this list of conditions and the following disclaimer in the documentation
+      and/or other materials provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
--- a/hamilton/caching/fingerprinting.py
+++ b/hamilton/caching/fingerprinting.py
@@ -36,10 +36,11 @@ implementation should pass the `depth` parameter to prevent `RecursionError`.
 import base64
 import datetime
 import functools
-import hashlib
 import logging
 import sys
 from collections.abc import Mapping, Sequence, Set
+
+import xxhash
 
 from hamilton.experimental import h_databackends
 
@@ -77,12 +78,16 @@ def _compact_hash(digest: bytes) -> str:
 
 
 def _hash_bytes(data: bytes) -> str:
-    """Hash raw bytes and compact-encode the digest.
+    """Hash raw bytes with the non-cryptographic xxh3_128 algorithm and
+    compact-encode the digest.
 
     All hashing in this module routes through this single helper so the
     underlying hashing algorithm can be changed in exactly one place.
+    xxh3_128 produces a 16-byte digest (24 base64url chars, the same width
+    as the md5 it replaces) while running substantially faster on
+    buffer-bound paths.
     """
-    return _compact_hash(hashlib.md5(data).digest())
+    return _compact_hash(xxhash.xxh3_128(data).digest())
 
 
 @functools.singledispatch
--- a/hamilton/caching/fingerprinting.py
+++ b/hamilton/caching/fingerprinting.py
@@ -33,14 +33,15 @@ IMPORTANT all container types that make a recursive call to `hash_value` or a sp
 implementation should pass the `depth` parameter to prevent `RecursionError`.
 """
 
+from __future__ import annotations
+
 import base64
 import datetime
 import functools
 import logging
 import sys
 from collections.abc import Mapping, Sequence, Set
-
-import xxhash
+from typing import TYPE_CHECKING
 
 from hamilton.experimental import h_databackends
 
@@ -50,6 +51,31 @@ try:
 except ImportError:
     NoneType = type(None)
 
+if TYPE_CHECKING:
+    from collections.abc import Buffer, Callable
+    from typing import Protocol
+
+    class Hash(Protocol):
+        def digest(self) -> bytes: ...
+
+
+hash_func: Callable[[str | Buffer], Hash]
+try:
+    # xxh3_128 produces a 16-byte digest (24 base64url chars, the same width as the
+    # md5 it replaces) while running substantially faster on buffer-bound paths.
+    # TODO(2.0): revisit once/if a dependency-free `apache-hamilton-core` package
+    # exists, so this optional performance dependency lands in the right tier.
+    import xxhash
+
+    hash_func = xxhash.xxh3_128
+except (ModuleNotFoundError, AttributeError):
+    # ModuleNotFoundError covers xxhash not being installed; AttributeError
+    # covers an xxhash older than 0.8.0 (which added xxh3_128).
+    # usedforsecurity=False avoids a ValueError on FIPS-mode Python; it
+    # doesn't change the digest.
+    import hashlib
+
+    hash_func = functools.partial(hashlib.md5, usedforsecurity=False)
 
 logger = logging.getLogger("hamilton.caching")
 
@@ -78,16 +104,20 @@ def _compact_hash(digest: bytes) -> str:
 
 
 def _hash_bytes(data: bytes) -> str:
-    """Hash raw bytes with the non-cryptographic xxh3_128 algorithm and
-    compact-encode the digest.
+    """Hash raw bytes and compact-encode the digest.
 
     All hashing in this module routes through this single helper so the
     underlying hashing algorithm can be changed in exactly one place.
-    xxh3_128 produces a 16-byte digest (24 base64url chars, the same width
-    as the md5 it replaces) while running substantially faster on
-    buffer-bound paths.
+
+    Uses the non-cryptographic xxh3_128 algorithm when the optional
+    ``xxhash`` package is installed (see the ``performance`` extra),
+    falling back to hashlib's md5 otherwise. Both produce a 16-byte digest
+    (24 base64url chars), so digest width is unaffected either way, but the
+    two algorithms don't produce the same bytes for the same input:
+    fingerprints are stable within an environment, not across installs with
+    a different backend.
     """
-    return _compact_hash(xxhash.xxh3_128(data).digest())
+    return _compact_hash(hash_func(data).digest())
 
 
 @functools.singledispatch
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,6 +51,7 @@ dependencies = [
     "pandas",
     "typing_extensions > 4.0.0",
     "typing_inspect",
+    "xxhash>=0.8.0",
 ]
 
 [project.optional-dependencies]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -51,7 +51,6 @@ dependencies = [
     "pandas",
     "typing_extensions > 4.0.0",
     "typing_inspect",
-    "xxhash>=0.8.0",
 ]
 
 [project.optional-dependencies]
@@ -71,6 +70,10 @@ experiments = [
 lsp = ["apache-hamilton-lsp"]
 openlineage = ["openlineage-python"]
 pandera = ["pandera"]
+performance = [
+    # Internal performance boosters go here
+    "xxhash>=0.8.0",
+]
 pydantic = ["pydantic>=2.0"]
 pyspark = [
   # we have to run these dependencies because Spark does not check to ensure the right target was called
@@ -135,6 +138,7 @@ test = [
   "xgboost; python_version < '3.14'",
   "xlsx2csv", # for excel data loader
   "xlsxwriter",  # Excel export requires 'xlsxwriter'
+  "xxhash>=0.8.0", # exercises the high-performance fingerprinting path in tests
 ]
 docs = [
   {include-group = "dev"},
--- a/tests/caching/test_fingerprinting.py
+++ b/tests/caching/test_fingerprinting.py
@@ -146,11 +146,11 @@ def test_max_recursion_depth():
 @pytest.mark.parametrize(
     ("obj", "expected_hash"),
     [
-        ("hello-world", "L1Q1Kh6_t1atHO_H8RbBeA=="),
-        (17.31231, "mJPTpPyXDSZgU-u8NuztIQ=="),
-        (16474, "6MgAp1NbMW0ZZpe_8iKVsg=="),
-        (True, "J2eGynSuIpd5bwVQzO9VVg=="),
-        (b"\x951!\x89u=\xe6\xadG\xdf", "d1DufDgRQmqi9Kt4Z2PeUQ=="),
+        ("hello-world", "EXXR8_e47ElS18aP2lThJA=="),
+        (17.31231, "tVUSIslYiBcW52c-7w4gvA=="),
+        (16474, "FAJ-iXM_Hwg9TCRreY8AyA=="),
+        (True, "qkJEg3-XQKmGWk5sWqmonw=="),
+        (b"\x951!\x89u=\xe6\xadG\xdf", "pPTyYkSU_x7NLB1Fp_YTyA=="),
     ],
 )
 def test_hash_primitive(obj, expected_hash):
@@ -161,8 +161,8 @@ def test_hash_primitive(obj, expected_hash):
 @pytest.mark.parametrize(
     ("obj", "expected_hash"),
     [
-        ([0, True, "hello-world"], "mlOjj4yeCrSDFSn5zgdEIg=="),
-        ((17.0, False, "world"), "BcRSGfyKeIOdym9B6TmAyQ=="),
+        ([0, True, "hello-world"], "I98OkNhfxtScJrYNTs4ZfQ=="),
+        ((17.0, False, "world"), "catgOMSnsbQj1_KELNQscw=="),
     ],
 )
 def test_hash_sequence(obj, expected_hash):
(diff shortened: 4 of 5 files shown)


jernejfrank:
> `benchmark_hash_algorithm.py` isolates the algorithm swap from the vectorization (#1629), holding the implementation constant and varying only the hash. The raw algorithm is ~8–27× faster, but **the end-to-end gain depends on how much of total time the hash step occupies**:
> rows 	backend 	md5 (ms) 	xxh3 (ms) 	speedup
> 500 	numpy 	0.03 	0.02 	**1.4×**
> 500 	pandas 	1.0 	1.0 	**1.0×**
> 500 	polars 	0.1 	0.1 	**1.3×**
> 5,000 	numpy 	0.1 	0.03 	**3.2×**
> 5,000 	pandas 	2.9 	3.3 	**0.9×**
> 5,000 	polars 	0.2 	0.1 	**1.8×**
> 50,000 	numpy 	0.6 	0.1 	**6.5×**
> 50,000 	pandas 	32.3 	33.0 	**1.0×**
> 50,000 	polars 	0.9 	0.4 	**2.3×**
> 500,000 	numpy 	9.4 	6.5 	**1.4×**
> 500,000 	pandas 	382 	382 	**1.0×**
> 500,000 	polars 	9.5 	4.2 	**2.3×**
> 5,000,000 	numpy 	96.3 	62.0 	**1.6×**
> 5,000,000 	pandas 	6,434 	6,405 	**1.0×**
> 5,000,000 	polars 	121 	72.6 	**1.7×**
> 
> > CPU: Intel i7-3770 @ 3.40 GHz · 32 GiB DDR3-1600 · pandas 3.0.3 · polars 1.41.2 · xxhash 3.7.0 · Python 3.14.2
> 

Maybe just me, but this seems to give no benefit for `pandas`, some for `polars`, and where it would be most useful is `numpy` (which I am tempted to assume is a marginal use-case for using hamilton.

Compared to #1619, it seems like #1628 actually does the lions share and not the switch to `xxhash`. I agree with @skrawcz and @elijahbenizzy here that this makes most sense to add as `hamilton[cache]` to keep the dependency surface small. Just my two pennies

Dev-iL:
> Maybe just me, but this seems to give no benefit for `pandas`, some for `polars`, and where it would be most useful is `numpy` (which I am tempted to assume is a marginal use-case for using hamilton.

Yup. I would look at this differently though: 
- Indeed, the effect of the hashing algo on pandas is insignificant. OTOH, people looking for performance don't stay on pandas.
- The ~2x speedup on polars is amazing for such a trivially-looking change.
- I too don't know how common numpy is, but if it's supported by hamilton, I guess there was interest in it at some point. Perhaps this is what people used for performance before polars became popular?

> Compared to #1619, it seems like #1628 actually does the lions share and not the switch to `xxhash`. I agree with @skrawcz and @elijahbenizzy here that this makes most sense to add as `hamilton[cache]` to keep the dependency surface small. Just my two pennies

- One of the reasons for splitting the PRs is so we can attribute the performance gains, honestly, to the right code change.
- It is still not clear to me what all of you mean when you say `hamilton[cache]` - do we: (a) want the entire caching functionality to be locked behind an extra, or (b) do we ship "slow" caching by default and the extra means "fast caching"? Claude's analysis of this possibility:

  > Two distinct approaches are considered: (A) make `xxhash` optional with a `hashlib` fallback, and (B) move caching behind `hamilton[caching]`. These are not the same and (A) is the dangerous one. An optional fallback means the persisted `data_version` depends on whether `xxhash` happens to be installed — so the same value fingerprints differently across machines, and a cache populated where `xxhash` exists is dead weight where it doesn't. For a feature whose entire pitch (per the docs) is _centralizing/sharing cache across pipelines and environments_, an environment-dependent fingerprint is a portability footgun. The determinism argument is not a preference, it's a correctness regression.

  Then the question becomes - do we create the new extra now, or do we split it out during a larger refactor when making pandas and numpy optional?

jernejfrank:
> > Maybe just me, but this seems to give no benefit for `pandas`, some for `polars`, and where it would be most useful is `numpy` (which I am tempted to assume is a marginal use-case for using hamilton.
> 
> Yup. I would look at this differently though:
> 
> * Indeed, the effect of the hashing algo on pandas is insignificant. OTOH, people looking for performance don't stay on pandas.
> * The ~2x speedup on polars is amazing for such a trivially-looking change.
> * I too don't know how common numpy is, but if it's supported by hamilton, I guess there was interest in it at some point. Perhaps this is what people used for performance before polars became popular?
> 
Yeah, the weird thing is also that to my knowledge `pandas` uses numpy arrays under the hood, so the discrepency between the two is very surprising. 

> * It is still not clear to me what all of you mean when you say `hamilton[cache]` - do we: (a) want the entire caching functionality to be locked behind an extra, or (b) do we ship "slow" caching by default and the extra means "fast caching"?

So I was thinking about (b). Enforcing (a) would be difficult because we would completely break backwards compatibility.

Having it as optional dependency with the `hashlib` as the standard implementation seems to me the safer choice because we do not commit to the dependency. We had before marked things experimental when we were unsure about the exact shape to signal to the user that they should use this with caution. This might be such a case where the best we can do is docs, until we figure out how to handle dependencies that are internal performance boosters without expanding the actual dependency surface.

Dev-iL:
> Yeah, the weird thing is also that to my knowledge `pandas` uses numpy arrays under the hood, so the discrepency between the two is very surprising.

It makes a lot of sense if you think of pandas as a very inefficient wrapper around numpy. Even if you optimize the numpy part of it, since the wrapper's overhead dominates by 1-2 orders of magnitude (see table above), the speedup is practically invisible.

Dev-iL:
@jernejfrank @skrawcz @elijahbenizzy 
I updated the PR such that xxhash is no longer a required dependency, and instead gated behind the optional `[performance]` extra. I hope this addresses your concerns.

skrawcz (commented):
Nice work — this is a clean, well-scoped final step in the fingerprinting series. Routing everything through the single `_hash_bytes` chokepoint means the algorithm swap touches exactly one line, the `xxhash` dependency is properly optional with a transparent md5 fallback, and the LICENSE attribution follows the existing pattern. Test coverage is strong: pinned digests for both backends, the `force_md5_backend` fixture, and the reload-based import-time tests genuinely exercise the `try/except`. Locally all 42 fingerprinting tests and the full 130-test caching suite pass; ruff check and format are clean.

A few items noted inline: two small cleanliness fixes (the bespoke `TYPE_CHECKING` re-declaration and a redundant `import functools`), plus a couple of optional nice-to-haves (a digest-width assertion and a docs note for the new `performance` extra + the cache-invalidation-on-backend-change behavior).

hamilton/caching/fingerprinting.py
```diff
@@ -49,6 +50,31 @@
 except ImportError:
     NoneType = type(None)
 
+TYPE_CHECKING: bool = False
```
skrawcz: This re-declares `TYPE_CHECKING` as a module-level bool, but the rest of the codebase imports it (`from typing import TYPE_CHECKING` — see `lifecycle/api.py`, `lifecycle/base.py`, `plugins/h_mcp/_helpers.py`, `graph_types.py`). Since `from __future__ import annotations` is already active, the standard import is free at runtime and keeps things consistent + tool-friendly. Suggest:

```python
from typing import TYPE_CHECKING
```

hamilton/caching/fingerprinting.py
```diff
@@ -49,6 +50,31 @@
...
+    # xxh3_128 produces a 16-byte digest (24 base64url chars, the same width as the
+    # md5 it replaces) while running substantially faster on buffer-bound paths.
+    import xxhash
+
+    hash_func = xxhash.xxh3_128
+except (ModuleNotFoundError, AttributeError):
+    # ModuleNotFoundError covers xxhash not being installed; AttributeError
+    # covers an xxhash older than 0.8.0 (which added xxh3_128).
+    # usedforsecurity=False avoids a ValueError on FIPS-mode Python; it
+    # doesn't change the digest.
```
skrawcz, line 75: `functools` is already imported at the top of the module (line 40), so this nested re-import is redundant and can be dropped. (`import hashlib` here is fine — that one is legitimately deferred to the fallback path.)

tests/caching/test_fingerprinting.py
```diff
@@ -212,20 +227,125 @@ def test_hash_mapping_where_order_doesnt_matter():
...
+    expected_hash = "mswHhNBBYN5mv6i-LcEeVw=="
     fingerprint = fingerprinting.hash_set(obj)
     assert fingerprint == expected_hash
 
 
 def test_hash_numpy():
     # dtype is pinned explicitly so the literal digest is reproducible across
     # platforms (the default integer dtype is platform-dependent).
+    array = np.array([[0, 1], [2, 3]], dtype=np.int64)
+    expected_hash = "Y1uek_eQTHejo2YtRvdWPQ=="
```
skrawcz, line 239: The 24-char digest width is the invariant that keeps all downstream interfaces (cache keys, `data_version`) unaffected by the swap, but it's only asserted implicitly via the pinned literals. A tiny explicit check under both backends would make the contract obvious, e.g. `assert len(fingerprinting._hash_bytes(b"x")) == 24`.

pyproject.toml
```diff
@@ -70,6 +70,10 @@ experiments = [
 lsp = ["apache-hamilton-lsp"]
 openlineage = ["openlineage-python"]
 pandera = ["pandera"]
+performance = [
```
skrawcz, line 73: The new `performance` extra isn't documented. A one-liner in the caching docs (`pip install "sf-hamilton[performance]"` for faster fingerprinting) plus a note that switching the backend invalidates previously persisted `data_version` caches would help users with large caches understand the first-run recompute.

hamilton/caching/fingerprinting.py
```diff
@@ -49,6 +50,31 @@
...
+    import functools
+    import hashlib
+
+    hash_func = functools.partial(hashlib.md5, usedforsecurity=False)
```
skrawcz, line 78: for Hamilton 2.0 I think we'll want to create a apache-hamilton-core package that really strives for no dependencies -- and then have apache-hamilton be batteries included where we can have some dependencies that give a good OOTB experience.

skrawcz: so put a TODO here for 2.0.0 consideration? or at least in the commit?

hamilton/caching/fingerprinting.py
```diff
@@ -49,6 +50,31 @@
...
+    import xxhash
+
+    hash_func = xxhash.xxh3_128
+except (ModuleNotFoundError, AttributeError):
+    # ModuleNotFoundError covers xxhash not being installed; AttributeError
+    # covers an xxhash older than 0.8.0 (which added xxh3_128).
+    # usedforsecurity=False avoids a ValueError on FIPS-mode Python; it
+    # doesn't change the digest.
+    import functools
+    import hashlib
```
skrawcz, line 76: Do we need to an env variable to enable someone to opt-out of xxhash? Just thinking of the very small case where someone has it, but was using the old way... ?

LICENSE
```diff
@@ -226,3 +226,33 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+-------------------------------
+
+This product depends on xxhash (the python-xxhash package, https://github.com/ifduyue/python-xxhash),
+which is licensed under the BSD 2-Clause License.
```
skrawcz: @Dev-iL I don't think this is required because it is not bundled as part of the source. 

Let's ask one of our mentors to @potiuk just to be sure?

hamilton/caching/fingerprinting.py
Dev-iL: I believe my approach is better since it avoids importing the `typing` module. Regardless, done for the sake of consistency.

hamilton/caching/fingerprinting.py
Dev-iL: Done

tests/caching/test_fingerprinting.py
Dev-iL: Done, but I don't see why the hash length matters.

pyproject.toml
Dev-iL: Done: added a "Hashing backend" section to docs/concepts/caching.rst covering the install command and the cache-invalidation behavior when the backend changes.

hamilton/caching/fingerprinting.py
Dev-iL: Done: added a TODO(2.0) comment above the xxhash import noting this should be revisited if a dependency-free apache-hamilton-core package materializes.

hamilton/caching/fingerprinting.py
Dev-iL: Valid case, I suppose. However: 1) I think there's nothing wrong with nudging users towards using what we consider better practices; 2) Cache was already invalidated by other recent changes, so impact on users should be minimal.

LICENSE
Dev-iL: Based on the example of Airflow (an Apache project with hundreds of dependencies), whose top-level LICENSE is pristine, looks like this is indeed not required. We can add it in a followup if this is the wrong move.

Dev-iL: The submodule LICENSE files [do have notices occasionally](https://github.com/apache/airflow/blob/main/airflow-core/LICENSE#L244). So at minimum - _this shouldn't be at the top level LICENSE_.

potiuk: Not required - yes.  https://infra.apache.org/licensing-howto.html#mod-notice

skrawcz (approved):

PR #1634, opened 2026-06-16
Fix sub-package LICENSE files + add build-from-source docs

skrawcz:
Addresses the ASF mentor -1 on apache-hamilton-lsp 0.2.0-RC1 (and the same latent issues in the other sub-packages).

## 1. LICENSE: stale `databackend.py` notice (all sub-packages)
Each sub-package's LICENSE was copied verbatim from the main `apache-hamilton` package, carrying a bundled-code notice for `hamilton/experimental/databackend.py` — a file that ships **only** in the main package, not in any sub-package. Removed from sdk, contrib, lsp, and ui.

## 2. LICENSE: lsp pygls path
The pygls/`ls_setup.py` notice used the monorepo path `dev_tools/language_server/tests/ls_setup.py`, which doesn't resolve inside the re-rooted archive. Fixed to the archive-relative path `tests/ls_setup.py`.

## 3. LICENSE: UI bundled frontend attribution
The `apache-hamilton-ui` **wheel** bundles a compiled frontend (`hamilton_ui/build/`) containing third-party JavaScript dependencies (React, Redux, Chart.js, React Flow, etc.) with no attribution.
- `ui/frontend/package.json`: new `licenses` script generating `THIRD-PARTY-LICENSES.txt` for production deps (license-checker-rseidelsohn)
- `scripts/apache_release_helper.py`: runs `npm run licenses` after the frontend build so the file is bundled into the wheel
- `ui/backend/LICENSE`: documents the bundled frontend and points to `hamilton_ui/build/THIRD-PARTY-LICENSES.txt`

All bundled frontend deps are permissive (MIT, Apache-2.0, BSD, ISC, etc.). Verified the rebuilt wheel contains `hamilton_ui/build/THIRD-PARTY-LICENSES.txt` and the corrected LICENSE.

## 4. Build-from-source docs
The mentor couldn't tell how to build the wheel from the source distribution. Added a "Building from source" section to all four READMEs:
- sdk, lsp, contrib: `flit build --no-use-vcs`
- ui: backend-only sdist builds directly; the wheel requires building the frontend (npm install/build/licenses), copying it into the package, then building the wheel — matching what `apache_release_helper.py` does.

## Follow-up
Requires re-cutting **RC2** for all four sub-packages after merge.


dev_tools/language_server/README.md
```diff
@@ -30,3 +30,24 @@ under the License.
...
+# artifacts are written to dist/
+```
+
+`flit build` produces both the `.tar.gz` source distribution and the
+`.whl` wheel. `--no-use-vcs` selects files from the working tree rather than
+git, which is what you want when building from an unpacked source release.
+
+# License
+
+The code here is licensed under the Apache 2.0 license. See the main repository [LICENSE](../../LICENSE) for details.
```
jernejfrank: Quick Q: when we make a release, we only have artefacts of this folder in the tarball, will the License link not break?

ui/backend/README.md
```diff
@@ -80,6 +80,44 @@ dr = (
...
+flit build --no-use-vcs --format wheel
+# artifact is written to dist/
+```
+
+The release script `scripts/apache_release_helper.py --package ui` performs all
+of these steps automatically.
+
 ## License
 
 Apache 2.0. See the main repository [LICENSE](../../LICENSE) for details.
```
jernejfrank: same here with the relative link to repo root

ui/sdk/README.md
```diff
@@ -88,5 +88,22 @@ Then run Apache Hamilton as normal! Each DAG run will be tracked, and you'll hav
...
+flit build --no-use-vcs
+# artifacts are written to dist/
+```
+
+`flit build` produces both the `.tar.gz` source distribution and the
+`.whl` wheel. `--no-use-vcs` selects files from the working tree rather than
+git, which is what you want when building from an unpacked source release.
+
 # License
 The code here is licensed under the Apache 2.0 license. See the main repository [LICENSE](../../LICENSE) for details.
```
jernejfrank: here too

jernejfrank (commented):
I just have that one thing with the license file link, happy to resolve if it is not an issue

ui/backend/LICENSE
```diff
@@ -202,27 +202,16 @@
...
-copies or substantial portions of the Software.
-
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
-OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
-SOFTWARE.
+This package is distributed in two forms:
```
pjfanning: I don't think this should be in the license - can it go in the readme?

dev_tools/language_server/README.md
skrawcz: Good catch — fixed in 37083ab3. Pointed it at the package's own bundled `LICENSE` (`./LICENSE`) instead of `../../LICENSE`, since each package tarball is re-rooted at the package dir and ships its own LICENSE.

ui/backend/README.md
skrawcz: Fixed in 37083ab3 — now points to the package-local `LICENSE` that's included in the wheel/sdist.

ui/sdk/README.md
skrawcz: Fixed in 37083ab3 — now points to the package-local `LICENSE`.

ui/backend/LICENSE
skrawcz: Agreed — removed the prose from `LICENSE` (now clean Apache-2.0 only) and moved the bundled-frontend explanation into the README's License section (37083ab3). The actual third-party license texts still ship in the wheel at `hamilton_ui/build/THIRD-PARTY-LICENSES.txt`.

jernejfrank (approved):
Cool!

PR #1648, opened 2026-06-22
Add optional module allowlist to PickleLoader deserialization

elijahbenizzy:
## Summary

Adds an optional allowlist to `PickleLoader.load_data` in `hamilton/io/default_data_loaders.py`. Without an allowlist: backward-compatible behavior, but emits a one-time `UserWarning` per load to surface that unrestricted pickle deserialization is in use. With an allowlist: a `_RestrictedUnpickler(pickle.Unpickler)` subclass overrides `find_class(module, name)` to validate against the allowlist and raises `UnpicklingError` before any `__reduce__` payload can resolve.

## API

Two ways to set the allowlist, in resolution order (highest priority first):

1. Per-instance: `PickleLoader(allowlist=[("myapp.models", "MyClass"), ...])`
2. Process-wide: `set_pickle_loader_allowlist([...])`
3. Otherwise: legacy unrestricted path + a one-time `UserWarning`

Allowlist entries are `(module, qualname)` tuples — per-class granularity rather than prefix-string. The motivation: pickle attacks routinely reach for specific symbols within *otherwise-trusted modules* (`builtins.eval`, `os.system`, `subprocess.Popen`), so denying by full `(module, qualname)` pair is more useful than module-prefix matching.

## Trust model (now in the docstring)

`PickleLoader` reads pickle bytes from a caller-supplied path. Pickle deserialization is unsafe when those bytes originate from a source the application doesn't fully control — files passed between teams, intermediate caches written by upstream pipelines, shared persistence backends. The allowlist closes the `__reduce__` arbitrary-code-execution primitive while keeping the historical contract of `PickleLoader` available for callers who explicitly opt out of restriction.

## Tests

5 new tests in `tests/io/test_default_adapters.py`:

- Legacy unrestricted-with-warning path roundtrip
- Allowlist permits a legitimate roundtrip when the type is listed
- Per-instance allowlist blocks a malicious `__reduce__` payload (sentinel side effect never fires)
- Module-level `set_pickle_loader_allowlist` applies to fresh `PickleLoader` instances
- Per-instance allowlist overrides module-level default

Full `tests/io/` runs 36/36 green.

### DIFF
--- a/hamilton/io/default_data_loaders.py
+++ b/hamilton/io/default_data_loaders.py
@@ -23,7 +23,7 @@ import pathlib
 import pickle
 import warnings
 from collections.abc import Collection
-from typing import Any, Optional
+from typing import Any
 
 from hamilton.io.data_adapters import DataLoader, DataSaver
 from hamilton.io.utils import get_file_metadata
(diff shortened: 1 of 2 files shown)


jernejfrank (approved):
Looks good

PR #1663, opened 2026-07-19
Add tag_release.sh for creating final (non-RC) release tags

skrawcz:
## Problem

During an RC build, `apache_release_helper.py` creates a tag like
`<package>-v<version>-incubating-RC<rc>` pointing at the commit the artifacts
were built from — typically a version-bump commit on the release branch, which
is **not** reachable from `main`. After the vote passes and the release branch is
squash-merged, there was no documented step to create a clean, RC-less release
tag on the mainline commit.

This left two gaps (both flagged by ASF mentors on prior releases):
- The only tag was the `-RC` one, which points at an off-main commit.
- `-RC`-suffixed tags make poor GitHub release names.

## Changes

- **`scripts/tag_release.sh`** — creates an annotated `<package>-v<version>-incubating`
  tag on a target commit (default `HEAD`; typically the logical mainline commit the
  release was cut from). It:
  - validates the package name,
  - warns if the target commit isn't reachable from `main`,
  - shows the existing `-RC` tag(s) as the record of what was voted on,
  - refuses to clobber an existing final tag,
  - supports `--commit <sha>`, `--remote <name>`, and `--push`.

- **`scripts/README.md`** — adds **step 7 (create the final release tag)** to the
  post-vote finalize sequence, with a multi-package example, and lists the script
  in the scripts table.

## Notes

The original `-RC` tag is intentionally left intact as the immutable record of
exactly what was voted on. This was used for the recent
apache-hamilton-{sdk,lsp,ui,contrib} sub-package releases — the final
`-incubating` tags now point at the logical mainline commit, consistent with how
`apache-hamilton-v1.90.0-incubating` is tagged.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -59,6 +59,7 @@ The core `apache-hamilton` package must be released first. The other four packag
 |---|---|
 | `scripts/apache_release_helper.py` | Build artifacts, sign, tag, upload RC to SVN, generate vote email |
 | `scripts/promote_rc.sh` | Move a voted RC from SVN dev to SVN release |
+| `scripts/tag_release.sh` | Create the final (non-RC) `-incubating` release tag on the merge commit |
 | `scripts/verify_apache_artifacts.py` | Verify GPG signatures, checksums, and license headers |
 | `scripts/verification-script.sh` | End-to-end RC validation (download, verify, build, test, examples) |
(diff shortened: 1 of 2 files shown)


jernejfrank (approved):
👍

PR #1675, opened 2026-08-01
Fix Polars temporal column stats JSON serialization for tracker

### LINKED ISSUE #1127: Hamilton tracker gives an error with datetime columns in polars dataframes
## Stack Traces

```
  File "/Users/yaser/Documents/shitcoins/.venv/lib/python3.12/site-packages/hamilton/node.py", line 249, in __call__
    return self.callable(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/elyase/test/.venv/lib/python3.12/site-packages/hamilton_sdk/tracking/polars_col_stats.py", line 52, in std
    return col.std()
           ^^^^^^^^^
  File "/Users/elyase/test/.venv/lib/python3.12/site-packages/polars/series/series.py", line 2049, in std
    return self._s.std(ddof)
           ^^^^^^^^^^^^^^^^^
polars.exceptions.InvalidOperationError: `std` operation not supported for dtype `datetime[μs]`
```

dgvj-work:
Make Polars Date/Datetime column stats JSON-safe for the tracker, and keep Time/Duration off the datetime UI path.

## Changes

- Keep Date/Datetime on `datetime_column_stats` with ISO timestamps (`base_data_type="datetime"`)
- Route Time/Duration to unhandled stats so the UI does not try to parse `"01:00:00"` / `"1 day, 0:00:00"` as Dates
- Add a Datetime Series regression test matching #1127, plus coverage that Time/Duration stay unhandled

## How I tested this

- `pytest ui/sdk/tests/tracking/test_polars_col_stats.py ui/sdk/tests/tracking/test_polars_stats.py` (22 passed)
- `ruff format` on the touched files (passed)

## Notes

- On main, temporal columns already avoid the original `std()` crash by routing through `datetime_column_stats` with `std=0.0`, but #1127 was still open: there was no regression test pinning that Datetime path. This PR closes that loop without sending Time/Duration through the datetime renderer.

Closes #1127

## Checklist

### DIFF
--- a/ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
+++ b/ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
@@ -19,6 +19,7 @@ import datetime
 
 import polars as pl
 from polars.exceptions import InvalidOperationError
+
 from hamilton_sdk.tracking import dataframe_stats as dfs
 
 
@@ -130,6 +131,16 @@ def numeric_column_stats(
     )
 
 
+def _temporal_to_jsonable(value: object) -> object:
+    # datetime.datetime is a subclass of datetime.date. Time/Duration also arrive via
+    # selectors.temporal() and must be JSON-safe for the UI (same base_data_type="datetime").
+    if isinstance(value, (datetime.date, datetime.time)):
+        return value.isoformat()
+    if isinstance(value, datetime.timedelta):
+        return str(value)
+    return value
+
+
 def datetime_column_stats(
     name: str,
     position: int,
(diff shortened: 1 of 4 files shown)


ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
```diff
@@ -130,6 +131,16 @@ def numeric_column_stats(
...
+def _temporal_to_jsonable(value: object) -> object:
+    # datetime.datetime is a subclass of datetime.date. Time/Duration also arrive via
+    # selectors.temporal() and must be JSON-safe for the UI (same base_data_type="datetime").
+    if isinstance(value, (datetime.date, datetime.time)):
+        return value.isoformat()
+    if isinstance(value, datetime.timedelta):
+        return str(value)
+    return value
+
+
```
jernejfrank: This breaks a couple of daterendering on the UI:

```
"2021-01-01T00:00:00" → valid Date
  "01:00:00"             → Invalid Date / NaN
  "1 day, 0:00:00"       → Invalid Date / NaN
  "2 days, 0:00:00"      → Invalid Date / NaN
```

Can we route Time/Duration to unhandled stats for now?

jernejfrank (commented):
Hey, just a small thing since not everything renders on UI

ui/sdk/src/hamilton_sdk/tracking/polars_col_stats.py
dgvj-work: @jernejfrank  Good catch, those Time/Duration strings are not valid JS Dates, so they were breaking the datetime range renderer.

I routed Time and Duration to unhandled stats instead of datetime stats. Date/Datetime still go through the datetime path with ISO timestamps (`2021-01-01T00:00:00`), which `new Date(...)` can parse.

The #1127 Datetime regression coverage is still there, plus a test that Time/Duration stay on the unhandled path (no min/max date strings).

dgvj-work:
@jernejfrank can you please review the changes.

PR #1687, opened 2026-08-17
Fix failing plotly CI

jernejfrank:
Plotly dropped the engine in v7 that got deprecated in v6 and we had a workaround pinning Kaleido/Plotly for py<3.14

With the new changes we can drop these workaround.

## Checklist


elijahbenizzy (approved):
engine=None by default so we stop passing a kwarg plotly 7 removed, explicit engine still honored, both tested. Also un-gates plotly/kaleido on 3.14. This unbreaks unit tests for every open PR. LGTM.

PR #1626, opened 2026-06-07
fix(ui/sdk): fix failing CI on master due to formatting issues

ArnavBalyan:
## Changes
 - CI linter was failing
 - Fix the same.

## How I tested this
 - CI
## Notes
 - N/A
## Checklist

- [X] PR has an informative and human-readable title (this will be pulled into the release notes)
- [X] Changes are limited to a single goal (no scope creep)
- [X] Code passed the pre-commit check & code is left cleaner/nicer than when first encountered.
- [X] Any _change_ in functionality is tested
- [X] New functions are documented (with a description, list of inputs, and expected output)
- [X] Placeholder code is flagged / future TODOs are captured in comments
- [X] Project documentation has been updated if adding/changing functionality.

PR #1692, opened 2026-08-22
Flush buffered events before the async tracking client's worker exits

### LINKED ISSUE #1214: Stop python process from exiting too early with async tracker still having data to send
# Current behavior
If you're in aws lamdba, the lambda can kill the event loop before we manage to send all events.
This means that things complete, but we're missing data for a few nodes/functions.

This is hard to recreate locally unless you kill the process immediately after the driver finished executing.
1. run async driver logging to tracker
2. immediate kill process

latest

# Expected behavior
The tracker should be able to register something with python to say "hey run this before the process closes" to ensure the events get sent --  we do this with registering `stop()` with the sync client. 

# Additional context
Asyncio and atexit don't seem to play well together. There seems to be a library that could help. But yeah could be easy, or could be digging into the internals of python to figure this out.

AmirF194:
Fixes #1214.

## Changes

`BasicAsynchronousHamiltonClient` (`ui/sdk/src/hamilton_sdk/api/clients.py`) had no `stop()`
and registered no atexit hook, unlike its synchronous sibling. `worker()`'s only exit path
waits for a `None` sentinel on `data_queue`, and nothing ever put one there for the async
client, so a buffered batch sat until the periodic flush timer happened to catch it, or was
lost outright if the process (or the event loop that owned the worker task) exited first.

- Added `async def stop()` on `BasicAsynchronousHamiltonClient`: sends the sentinel and awaits
  the worker task so it flushes and exits. This is not wired to `atexit`, deliberately: the
  worker is a task on the caller's event loop, and by the time atexit callbacks run at
  interpreter shutdown that loop may already be closed, so callers must await `stop()`
  themselves before the loop stops running (mirroring the sync client isn't possible here for
  that reason; happy to add an `atexit`-scheduled best-effort variant too if maintainers want
  one, but that has its own shutdown-ordering edge cases worth discussing first).
- Added the matching `async def stop()` on `AsyncHamiltonTracker` (`adapters.py`), delegating
  to the client.
- Fixed a bug in `worker()`'s sentinel handling that this exercises for the first time: it
  appended the received item to `batch` before checking whether it was the `None` sentinel, so
  the first real `stop()` call flushed a batch containing `None` and raised. Reordered to check
  first, matching the synchronous client's worker.

## How I tested this

Ran in a clean `python:3.13-slim` and `python:3.10-slim` Docker container (the SDK's CI matrix
runs 3.10-3.14; I could not run the 3.11/3.12/3.14 legs locally but the change has no
version-specific code paths):

- New regression tests in `tests/test_clients.py` and `tests/test_adapters.py`: fail with
  `AttributeError` on unmodified `main` (confirmed via `git stash`), pass on this branch.
  `test_async_client_stop_flushes_buffered_item_and_exits_worker` enqueues an item, calls
  `stop()` before the flush interval elapses, and asserts the mocked backend request actually
  fired and the queue drained; it caught the sentinel-ordering bug above on first run.
- Full `pytest tests/` (minus the `ray`/`pyspark`-gated tests, which are `pytest.importorskip`
  guarded and unrelated to this change): 140 passed, 5 skipped, no failures.
- `ruff check` and `ruff format --diff` on the four changed files: clean. This repo's CI does
  not run a lint/static-checks job for `ui/sdk`-only changes (checked `gh pr checks` on a
  recently merged `ui/sdk`-only PR, #1626), so this was a courtesy check, not a required gate.

Not verified: behavior under a real ASGI/FastAPI shutdown sequence (e.g. the app in
`examples/async/fastapi_example.py`, which doesn't call the new `stop()` either); the fix is
scoped to the tracker itself, not to updating that example.

### DIFF
--- a/ui/sdk/src/hamilton_sdk/adapters.py
+++ b/ui/sdk/src/hamilton_sdk/adapters.py
@@ -469,6 +469,14 @@ class AsyncHamiltonTracker(
         self.initialized = True
         return self
 
+    async def stop(self):
+        """Flush any buffered tracking data and stop the background worker.
+
+        Call this before your event loop shuts down; nothing else flushes
+        the tracker's queue for you.
+        """
+        await self.client.stop()
+
     async def post_graph_construct(
         self, graph: h_graph.FunctionGraph, modules: list[ModuleType], config: dict[str, Any]
     ):
--- a/ui/sdk/src/hamilton_sdk/api/clients.py
+++ b/ui/sdk/src/hamilton_sdk/api/clients.py
@@ -585,9 +585,24 @@ class BasicAsynchronousHamiltonClient(HamiltonClient):
         self.data_queue = asyncio.Queue()
         self.running = True
         self.max_batch_size = 100
+        self.worker_task: asyncio.Task | None = None
 
     async def ainit(self):
-        asyncio.create_task(self.worker())
+        self.worker_task = asyncio.create_task(self.worker())
+
+    async def stop(self):
+        """Signal the worker to flush whatever is buffered and exit.
+
+        Unlike BasicSynchronousHamiltonClient, this is not wired to atexit:
+        the worker is a task on this event loop, and by the time atexit
+        callbacks run at interpreter shutdown the loop may already be
+        closed, so callers must await this explicitly before the loop
+        that owns it stops running.
+        """
+        if self.worker_task is None or self.worker_task.done():
+            return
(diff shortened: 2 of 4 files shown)


AmirF194:
Checking in, no rush. Happy to rebase or adjust scope if that would help it move.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

